PR #16, opened 2022-05-16
Fix availability label stories and eslint settings

Adamik10:
There was an error in indentation of argument types for the Availability label story, which prevented the default values to be read.

### DIFF
--- a/package.json
+++ b/package.json
@@ -111,7 +111,7 @@
     "prop-types": "Since we use former ddb-react components that depend on prop-types we keep this. Should be removed when usage of prop-types is deprecated."
   },
   "dependencies": {
-    "@danskernesdigitalebibliotek/dpl-design-system": "^0.0.0-3bd678a68e0854369f27016877773f6878a1cb61",
+    "@danskernesdigitalebibliotek/dpl-design-system": "0.0.0-372e102a9ded64f39067eb2a13f7485342418a23",
     "@reach/alert": "^0.17.0",
     "@reach/dialog": "^0.17.0",
     "@reduxjs/toolkit": "^1.8.1",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -111,7 +111,7 @@
     "prop-types": "Since we use former ddb-react components that depend on prop-types we keep this. Should be removed when usage of prop-types is deprecated."
   },
   "dependencies": {
-    "@danskernesdigitalebibliotek/dpl-design-system": "^0.0.0-3bd678a68e0854369f27016877773f6878a1cb61",
+    "@danskernesdigitalebibliotek/dpl-design-system": "0.0.0-372e102a9ded64f39067eb2a13f7485342418a23",
```
cableman, line 114: Would it be possible to always use latest release? Or do the design-system not change that offen?

It was this we talked about moving into npmjs.org

cableman (approved):

package.json
spaceo: The thing is that with npm/yarn you **can** run eg.: `yarn add @danskernesdigitalebibliotek/dpl-design-system@latest` but you'll end up with `@danskernesdigitalebibliotek/dpl-design-system@0.0.0-xxxxxxx`anyways. Apparently that is how it works opposed to composer.json where you literally have `@latest` specified in composer.json.
The result of that is that we have to remember run the `yarn add` command on a regular basis in order to fetch the latest HEAD.

cableman: We may need to add that information to a readme to help others to not asking the same question :-)

Adamik10: Diddly done :)

cableman (approved):

package.json
Adamik10: It can be found in this commit: [Mention dpl-design-system npm package in README.md](https://github.com/danskernesdigitalebibliotek/dpl-react/pull/16/commits/40d5e31047122e1bf037ef8ca5a0335336e8127e)

PR #17, opened 2022-05-17
Add clock component

sinejespersen:


src/components/clock/clock.scss
```diff
@@ -1,58 +1,82 @@
...
 	}
 
-	.mark-9 {
-		width: 6px;
-		height: 2px;
-		left: 0%;
-		top: calc(50% - 2px);
-		position: absolute;
-	}
+	$size: 4px;
```
Adamik10, line 17: @sinejespersen I'm just wondering - would this variable naming maybe be too general? When this styling becomes a part of the dpl-design-system, and it's imported by base.scss from its own file, it could pose some problems.

sinejespersen: Yes valid input thanks, I've changed it and changed the scss to be bem in structure (dusting off my bem-skills as well...) and prefixing the class with "dpl" as per your guidelines...

Adamik10: @sinejespersen oh yeah, about the dpl prefix - I'm not quite sure how that works. These dpl- prefix rules come from before my age, and I haven't dwelled in the topic yet.
I made all my design in the design system, and then just imported it. We really have to help to make sure you can do that as well 🙏🏼.

PR #18, opened 2022-05-17
Graphql codegen setup + demo components

spaceo:
#### Link to issue

https://reload.atlassian.net/browse/DDFSOEG-103

#### Description

This PR is introducing following:
- A way to set and persist a library token through a component only meant to be used in Storybook context.
- A demo story that includes following components:
   - A demo search header
   - A demo search bar
   - A demo suggest drop down
 
You need to set a library token in order to see the suggest drop down in action.

#### How to review

Read commits one by one.
Try to run storybook (`sudo yarn dev`) , insert a library token, and try out the search header app.

#### Screenshot of the result
image

image

#### Checklist


package.json
```diff
@@ -33,7 +33,9 @@
     "start:storybook:dev": "NODE_ENV=development  start-storybook --port 80",
     "dev": "concurrently --raw \"yarn start:storybook:dev\" \"yarn watch:lint:scss\"",
     "test": "cypress run",
-    "cypress:open": "cypress open"
+    "cypress:open": "cypress open",
+    "codegen:dbc:gateway": "CODEGEN_SCHEMA_AGENCY=775100 CODEGEN_SCHEMA_PROFILE=opac graphql-codegen --config dbc-gateway.codegen.yml",
```
cableman, line 37: Hardcode agency id?

package.json
```diff
@@ -101,7 +112,9 @@
     "stylelint-scss": "^4.2.0",
     "stylelint-webpack-plugin": "^3.2.0",
     "svg-url-loader": "^7.1.1",
+    "ts-node": "^10.7.0",
```
cableman, line 115: Adding ts-node should it be used for other TS compile tasks?

cableman (changes requested):
Minor comments and I not sure way the `src/core/dbc-gateway/generated/` generated code is added to the codebase and not generated on the fly.

package.json
spaceo: It is used for constructing the schema url form where we generate the code and types from.
This is the info that we got so far, from where to retrieve the schema.
So why not just use a hardcoded url, one might ask?
This is just going one step further and parameterising the agency id and profile.
And I put the the parameters directly into package.json in for helping the developers.
Otherwise you would need to export the parameters every time in your console session or implement usage of .env.
But since there is nothing secret here I choose to put it as a part of the command.

spaceo:
> Minor comments and I not sure way the `src/core/dbc-gateway/generated/` generated code is added to the codebase and not generated on the fly.

We need the types and code generated while developing.
That's why it is committed as a part of the codebase.
An extra advantage of that is that we can also track changes in the API

package.json
spaceo: We use it to run typescript in node. Which is done in this script, that has been introduced in this PR:
https://github.com/danskernesdigitalebibliotek/dpl-react/pull/18/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R38

package.json
cableman: But should we here than use 100200 (bibliotek.dk's agency id)?

package.json
cableman: So it's to speed up local development and have "continuous compilation" of TS?

package.json
spaceo: Yes. Makes sense. I'll change that

cableman (approved):

PR #22, opened 2022-05-31
Feature/search header app

spaceo:
#### Link to issue
https://reload.atlassian.net/browse/DDFSOEG-95
https://reload.atlassian.net/browse/DDFSOEG-97

#### Description
The search header app is one of three apps that will be created for this project. It serves as a vessel for search bar, search string suggestion, material suggestion, and topic suggestion components, and can be mounted into other projects. 

Upon searching (writing a term in the input, and either pressing enter or clicking on the search button) it redirects to a url containing the search query parameters. Base of this URL can also be adjusted based on the domain it is set into, as the app accepts a prop.  

#### Screenshot of the result
image

As seen on the screenshot, there is a bit that is highlighted by arrows. This is the search bar component. But the whole header design (the part around the input field) is included in the story as well - although it is not a part of the search header app per-say. This is because the design of the website header shapes the design of the input field - and it doesn't hold its shape when taken out of the website header. 
However, the surrounding header is only added in the storybook story to provide context. It is not a part of the react app itself. This context will be provided by the project this app is mounted into (e.g. dpl-cms).


src/components/search-bar/search-bar.tsx
```diff
@@ -0,0 +1,41 @@
...
+  altText?: string;
+  inputPlaceholder?: string;
+}
+
+const SearchBar: React.FC<SearchBarProps> = ({
+  q,
+  setQuery,
+  searchHeaderUrl = "/search",
+  altText = "search icon",
+  inputPlaceholder = "Search here"
```
spaceo, line 17: Just a note: All texts are depending on how we want to implement strings in the future. But let's keep it like this for now.

src/apps/search-header/search-header.dev.tsx
```diff
@@ -0,0 +1,37 @@
...
+  }
+} as ComponentMeta<typeof SearchHeader>;
+
+export const Default: ComponentStory<typeof SearchHeader> = (
+  args: SearchHeaderProps
+) => (
+  // We use the Header component as context to the search bar.
+  // It is the Header that creates the Search bar's design -
+  // - without it, the Search bar loses its shape.
+  <Header>
```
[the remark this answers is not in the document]
Adamik10: gotcha :)

src/components/search-bar/search-bar.tsx
Adamik10: True, let us see how we solve the implementation for system texts in the future.

PR #27, opened 2022-06-05
Add functionality for handling app string props

spaceo:
With this initiative we can access strings in nested components inside 
the apps without passing props.

#### Link to issue

https://reload.atlassian.net/browse/DDFSOEG-122

#### Description

A high order component wraps the entry component and puts all props that ends with `*Text` in the redux store.

When the strings are needed you implement the `useText()` hook that returns a function that you can use to fetch them

#### Checklist

### DIFF
--- a/docs/ui_text_handling.md
+++ b/docs/ui_text_handling.md
@@ -1,0 +1,63 @@
+# UI Text Handling
+
+This document describes how to use the text functionality
+that is partly defined in src/core/utils/text.tsx and in src/core/text.slice.ts.
+
+## Main Purpose
+
+The main purpose of the functionality is to be able to access strings defined
+at app level inside of sub components
+without passing them all the way down via props.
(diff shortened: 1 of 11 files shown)


Adamik10 (approved):
Very nice :) Can't wait to start using this

docs/ui_text_handling.md
```diff
@@ -0,0 +1,63 @@
...
+
+The main purpose of the functionality is to be able to access strings defined
+at app level inside of sub components
+without passing them all the way down via props.
+You can read more about the decision
+and considerations [here](../architecture/adr-002-ui-text-handling.md).
+
+## How to use it
+
+In order to use the system the component that has the text props
```
sinejespersen, line 16: + the props should end on "text" yes?

sinejespersen (approved):

PR #31, opened 2022-06-14
poc-modal

sinejespersen:
#### Link to issue

https://jira.itkdev.dk/browse/CMN-215

#### Description

Created a modal wrapper, that accepts "children", which are supposed to be the content of the modal.
The demo has two modals, where the button to open the second is located within the first (as this is a scenario we have in our spec).

The modal has autofocus set on the close button, to move screen reader focus from main content. I have a suspicion this can be solved in a more elegant way, so feedback is welcome.

Sister pr: 
https://github.com/danskernesdigitalebibliotek/dpl-design-system/pull/68

#### Screenshot of the result

image

#### Checklist

### DIFF
--- a/src/apps/demo-modal/demo-modal.tsx
+++ b/src/apps/demo-modal/demo-modal.tsx
@@ -6,11 +6,13 @@ import { openModal } from "../../core/modal.slice";
 export interface DemoModalProps {
   ariaLabelModalOne: string;
   ariaLabelModalTwo: string;
+  screenReaderModalDescriptionText: string;
 }
 
 const DemoModal: React.FC<DemoModalProps> = ({
   ariaLabelModalOne,
-  ariaLabelModalTwo
+  ariaLabelModalTwo,
+  screenReaderModalDescriptionText
 }) => {
   const dispatch = useDispatch();
   const modalIdOne = "demo-modal-one";
@@ -19,13 +21,15 @@ const DemoModal: React.FC<DemoModalProps> = ({
   return (
     <>
       <ModalWrapper
-        closeModalAriaLabel={ariaLabelModalTwo}
+        closeModalAriaLabelText={ariaLabelModalTwo}
+        screenReaderModalDescriptionText={screenReaderModalDescriptionText}
         modalId={modalIdTwo}
       >
         <h1>{modalIdTwo}</h1>
       </ModalWrapper>
       <ModalWrapper
-        closeModalAriaLabel={ariaLabelModalOne}
+        closeModalAriaLabelText={ariaLabelModalOne}
+        screenReaderModalDescriptionText={screenReaderModalDescriptionText}
         modalId={modalIdOne}
       >
         <h1>{modalIdOne}</h1>
--- a/src/apps/demo-modal/demo-modal.tsx
+++ b/src/apps/demo-modal/demo-modal.tsx
@@ -1,0 +1,53 @@
+import * as React from "react";
+import { useDispatch } from "react-redux";
+import ModalWrapper from "../../components/modal/modal-wrapper";
+import { openModal } from "../../core/modal.slice";
+
+export interface DemoModalProps {
+  ariaLabelModalOne: string;
+  ariaLabelModalTwo: string;
+}
+
+const DemoModal: React.FC<DemoModalProps> = ({
+  ariaLabelModalOne,
+  ariaLabelModalTwo
+}) => {
+  const dispatch = useDispatch();
+  const modalIdOne = "demo-modal-one";
+  const modalIdTwo = "demo-modal-two";
+
+  return (
+    <>
+      <ModalWrapper
+        closeModalAriaLabel={ariaLabelModalTwo}
+        modalId={modalIdTwo}
+      >
+        <h1>{modalIdTwo}</h1>
+      </ModalWrapper>
+      <ModalWrapper
+        closeModalAriaLabel={ariaLabelModalOne}
+        modalId={modalIdOne}
+      >
+        <h1>{modalIdOne}</h1>
+        <button
+          type="button"
+          onClick={() => {
+            dispatch(openModal({ modalId: modalIdTwo }));
+          }}
+        >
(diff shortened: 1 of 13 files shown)


src/core/modal.slice.ts
```diff
@@ -0,0 +1,28 @@
...
+  payload: {
+    modalId: string;
+  };
+}
+interface StateProps {
+  modalIds: string[];
+}
+const modalSlice = createSlice({
+  name: "modal",
+  initialState: { modalIds: [] },
```
spaceo, line 13:
```suggestion
  initialState: { modalIds: [] }: StateProps,
```

src/core/modal.slice.ts
```diff
@@ -0,0 +1,28 @@
+import { createSlice } from "@reduxjs/toolkit";
+
```
spaceo, line 2: Really a detail but since you use the `modalId`several times you could consider defining it as a type:
```ts
type ModalId = string;

interface PayloadProps {
  payload: {
    modalId: ModalId;
  };
}
interface StateProps {
  modalIds: ModalId[];
}

File: src/apps/demo-modal/demo-modal.tsx [hunk 3]
```diff
@@ -0,0 +1,59 @@
...
+export interface DemoModalProps {
+  ariaLabelModalOne: string;
+  ariaLabelModalTwo: string;
+  screenReaderModalDescriptionText: string;
+}
+
+const DemoModal: React.FC<DemoModalProps> = ({
+  ariaLabelModalOne,
+  ariaLabelModalTwo,
+  screenReaderModalDescriptionText
```
spaceo, line 15: I think all the texts should be handled in the way of the newly introduced text handling feature:
https://github.com/danskernesdigitalebibliotek/dpl-react/blob/main/docs/ui_text_handling.md#ui-text-handling

src/components/modal/modal-wrapper.tsx
```diff
@@ -0,0 +1,79 @@
...
+  modalId,
+  children,
+  closeModalAriaLabelText,
+  screenReaderModalDescriptionText
+}: ModalWrapperProps) {
+  const dispatch = useDispatch();
+  const { modalIds } = useSelector((s: ModalIdsProps) => s.modal);
+
+  useEffect(() => {
+    const searchParams = new URLSearchParams(window.location.search);
```
spaceo, line 29: I think we should do this with a helper funktion that is on its way to your review.
Although I can see now that I missed using the url parameter available as an parameter for the function:
https://github.com/danskernesdigitalebibliotek/dpl-react/pull/33/files#diff-b388ec997dbc272cfdf0283ac04a0762401ff68fab0d3c9eba31a132b581e18bR11

spaceo (commented):
Looks great. :)
Nice with redux added to the mix. 
I have a few comments

src/components/modal/modal-wrapper.tsx
```diff
@@ -0,0 +1,79 @@
...
+    >
+      <div className="modal-screen-reader-description" id={`modal-${modalId}`}>
+        {screenReaderModalDescriptionText}
+      </div>
+      <button
+        type="button"
+        /* I am not sure this is the best way, a focusable element in a modal must have focus when opened, 
+        or else the screen reader will remain on the main page */
+        // eslint-disable-next-line jsx-a11y/no-autofocus
+        autoFocus
```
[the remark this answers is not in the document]
sinejespersen: Yeah, I think I made a comment requesting feedback on that..? first child rendered should probably be first focusable child rendered, and what if the modal doesn't have a focusable child?

src/apps/demo-modal/demo-modal.tsx
```diff
@@ -0,0 +1,59 @@
...
+}) => {
+  const dispatch = useDispatch();
+  const modalIdOne = "demo-modal-one";
+  const modalIdTwo = "demo-modal-two";
+
+  return (
+    <>
+      <ModalWrapper
```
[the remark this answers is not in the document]
sinejespersen: Yesssss

src/components/modal/modal-wrapper.tsx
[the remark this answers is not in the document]
spaceo: Yes you did. Maybe the close button could be a fallback in that situation?

src/core/modal.slice.ts
sinejespersen: And then export it and use it in other files..?

src/core/utils/modal.tsx
```diff
@@ -0,0 +1,75 @@
...
+          zIndex: modalIds.indexOf(modalId)
+        }}
+        aria-label={closeModalAriaLabelText}
+        onClick={() => {
+          dispatch(closeModal({ modalId }));
+        }}
+      >
+        <img src={CloseIcon} alt="" />
+        {/* alt="": Hidden from screen readers, because the aria-label is sufficient */}
+      </button>
```
kasperg, line 69: :+1: for adding this explanation.

src/core/utils/modal.tsx
```diff
@@ -0,0 +1,75 @@
...
+        {t("screenReaderModalDescriptionText")}
+      </div>
+      <button
+        type="button"
+        /* A focusable element in a modal must have focus when opened, 
+        or else the screen reader will remain on the main page */
+        // eslint-disable-next-line jsx-a11y/no-autofocus
+        autoFocus
+        className="btn-ui modal-btn-close"
+        aria-describedby={`modal-${modalId}`}
```
kasperg, line 58: Please remember to include `aria-describedby` in the design system.

kasperg (changes requested):
:+1: I think the POC look good!

Note that 9ecb970d8a6fa04dc1fbe9ef84807bab08b32250 contains a lot of code style changes. I do not know why they have been introduced but please remove them from this PR.

Also note that Cypress tests fail. Please look into this.

kasperg (approved):
:+1: LGTM

PR #40, opened 2022-06-16
Feature/button favorite ddfsoeg 141 V1

kasperbirch1:
#### This is the first version

It only checks if id is on used list, the next version will also be able to add / remove using orval auto generated code

This show that we can using orval auto generate code to help retrieve data to see if a user has a materials on his list. and introduce the button favorite

#### Link to issue
https://reload.atlassian.net/browse/DDFSOEG-142

#### Description
The button favorite accepts an id and if that id is on the used favorite list the fill prop will change. 

#### Checklist


orval.config.ts
```diff
@@ -0,0 +1,31 @@
+// eslint-disable-next-line import/no-extraneous-dependencies
```
cableman, line 1: This config file should be moved into `src/core/material-list-api` as every generated client needs to have there own config file. Then run orval with `--config` alternative rename the file to `material-list-api.orval.config.ts`?

package.json
```diff
@@ -35,6 +35,7 @@
     "test": "cypress run",
     "cypress:open": "cypress open",
     "codegen:dbc:gateway": "CODEGEN_SCHEMA_AGENCY=100200 CODEGEN_SCHEMA_PROFILE=opac graphql-codegen --config dbc-gateway.codegen.yml",
+    "codegen:rest-services": "orval",
```
cableman, line 38: Add `--config <FILE>` to use the one that matches this client.

src/apps/hello-world/hello-world.entry.tsx
```diff
@@ -8,8 +8,6 @@ export interface HelloWorldEntryProps {
   whatText: string;
 }
 
-const HelloWorldEntry: React.FC<HelloWorldEntryProps> = (
-  props: HelloWorldEntryProps
-) => <HelloWorld />;
+const HelloWorldEntry: React.FC<HelloWorldEntryProps> = () => <HelloWorld />;
```
cableman, line 11: Why do `HelloWord` change by adding an API client?

cableman (changes requested):

orval.config.ts
cableman: Forget about this one @kasperg has just made me adware that it is possible to have more than one client defined in the same config file :-)

package.json
kasperg: I will resolve this based on the comment above.

src/apps/hello-world/hello-world.entry.tsx
kasperg: I think it is leftovers from #30 where the generated client was used in the Hello World app.

PR #42, opened 2022-06-21
Feature/cover service

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFSOEG-145

#### Description
New Material component is introduced
that uses orval(react-query) to fetch data from this API: https://cover.dandigbib.org with a materialId. 
- it only retrieves the image size specified in the size prop
- it changes from <span> to <a> if there is materialDescription
- it changes from <span> to <a> if there is materialUrl

#### Screenshot of the result
Skærmbillede 2022-06-21 kl  15 26 36

#### Checklist


src/components/button-favourite/button-favourite.tsx
```diff
@@ -0,0 +1,35 @@
...
+const ButtonFavourite: React.FC<ButtonFavouriteProps> = ({ id }) => {
+  const [fillState, setFillState] = useState<boolean>(false);
+  // There is a variety of properties to be used here.
+  // Eg. if you need a loader you can use `isLoading`.
+  // Read more here: https://react-query.tanstack.com/reference/useQuery
+  const { data, isSuccess } = useGetList("default");
+
+  useEffect(() => {
+    if (isSuccess && data) {
+      // This is because there does not come a property it is called materials but instead is called collections
```
sinejespersen, line 19: Not sure I understand this comment

src/components/material/material.tsx
```diff
@@ -0,0 +1,66 @@
...
+      `material material--${size}`,
+      tintClasses[tint || "default"],
+      {
+        material__animate: animate
+      }
+    )
+  };
+
+  const coverUrl = data?.[0]?.imageUrls?.[`${size}`]?.url;
+  const materialCover = coverUrl && (
```
sinejespersen, line 51: This is a tricky one, as if the img has alt="" and is inside a link that doesnt have any text inside, it will cause an error with the accessibility-tool digitaliseringsstyrelsen are using (sorry for the danglish). Do we have a material title as an alternative instead?

src/components/button-favourite/button-favourite.tsx
kasperbirch1: You can ignore that, it's changed in this PR
https://github.com/danskernesdigitalebibliotek/dpl-react/pull/43

src/components/material/material.tsx
kasperbirch1: Unfortunately, we do not have a title unless we send one from a component above. 
Do you think we should do it?

sinejespersen: This element will probably show up a lot, which could cause a lot of accessibility errors, so I think something should be done. The title would be good if it is possible

kasperbirch1: We have changed it so that there must be a description before an alt text before a tag will be rendered

kasperbirch1: It has been changed so that there must be a description for it to be rendered a tag. https://github.com/danskernesdigitalebibliotek/dpl-react/pull/42/commits/5c47757ab7fb6783408db96a4160ac9ab1a8f647

orval.config.ts
```diff
@@ -0,0 +1,55 @@
...
+        },
+        query: {
+          useQuery: true
+        }
+      },
+      prettier: true
+    },
+    input: {
+      target:
+        "https://raw.githubusercontent.com/danskernesdigitalebibliotek/ddb-cover-service/master/public/spec.yaml",
```
cableman, line 49: Should use the online version as that matches the currently deployed: https://coverservice.itkdev.dk/api/v2/spec.yml

cableman (changes requested):

sinejespersen (approved):

cableman (approved):

PR #43, opened 2022-06-21
button-favorite component  V2 (use has item)

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFSOEG-141

#### Description
This is a version 2 of button-favorite. This is PR for V1 https://github.com/danskernesdigitalebibliotek/dpl-react/pull/40

- This is a switch from using useGetList to useHasItem to not have to loop over all the user's favorites on their list.  
- Tests have also been made for button-favorite

#### Screenshot of the result

#### Checklist


src/components/button-favourite/button-favourite.tsx
```diff
@@ -1,32 +1,59 @@
...
+} from "../../core/material-list-api/material-list";
 
 export interface ButtonFavouriteProps {
-  id: string;
+  materialId: string;
 }
 
-const ButtonFavourite: React.FC<ButtonFavouriteProps> = ({ id }) => {
+// TODO
+// We have to check if user is login and redirect if not
```
kasperg, line 14:
```suggestion
// TODO We have to check if the user is logged and redirect if not
```

src/core/material-list-api/mutator/fetcher.ts
```diff
@@ -61,6 +65,6 @@ export const fetcher = async <ResponseType>({
 
 export default fetcher;
 
-export type ErrorType<ErrorData> = ErrorData;
+export type ErrorType<ErrorData> = ErrorData & { status: number };
```
kasperg, line 68: Please check if we actually need to add anything here to our ErrorType or this change should be removed.

src/components/button-favourite/button-favourite.test.ts
```diff
@@ -0,0 +1,46 @@
+describe("Favorite button", () => {
+  it("Material id is not on user's favorites list", () => {
```
kasperg, line 2: Consider following the textual convention á la:

```suggestion
  it("Adds a new material to the favorites list when clicked", () => {
```

src/components/button-favourite/button-favourite.tsx
```diff
@@ -0,0 +1,61 @@
...
+    } else {
+      addItem("default", materialId);
+      setFillState(true);
+    }
+  };
+
+  return (
+    <button
+      type="button"
+      aria-label={fillState ? "Fjern fra favoritter" : "Tilføj til favoritter"}
```
sinejespersen, line 52: Texts as props?

kasperbirch1: I'm not sure what you mean? fillState is a boolean that determines which of the texts is displayed?

sinejespersen: I mean, the aria-label should be a prop?

kasperbirch1: Okay, we have switched to using useText() instead, please see this commit:
https://github.com/danskernesdigitalebibliotek/dpl-react/pull/43/commits/50f5237532b03cd701ae1b183a039072eb363839#diff-8eccef2c16d12483231c47763961a184babc8f6788f6e3883890726714ccb3ca

cableman: But all tekst should come from the CMS.

PR #44, opened 2022-06-22
5463: FBS adapter client

cableman:
#### Link to issue

[5463](https://platform.dandigbib.org/issues/5463)

#### Description

* FBS adapter client with swagger 2.0 yaml definition based on FBS json swagger documentation.
* UserToken add to story-book to enable test without adgangsplatform redirects.

#### Screenshot of the result

No screenshot.

#### Checklist

Tool used to convert the swagger files: https://github.com/itk-dev/dpl-fbs-adapter-tool/tree/develop


docker-compose.yml
```diff
@@ -23,3 +24,8 @@ volumes:
       type: none
       device: $PWD
       o: bind
+  cache:
+    driver_opts:
+      type: none
+      device: $PWD/.cache
+      o: bind
```
kasperg, line 31: Consider whether this could be simplified and optimized.

As I understand it this will effectively mount the `.cache` directory twice: Once in the root because it is included in `$PWD` and once in the node cache.

I get why we want to use a directory within the project as the cache dir but [we could configure Yarn to use `./.cache` as the cache folder using an environment variable for the Node image](https://classic.yarnpkg.com/en/docs/cli/cache/#toc-change-the-cache-path-for-yarn).

This way we would not have to add an extra volume.

src/core/fbs/fbs-adapter.yaml
```diff
@@ -0,0 +1,1712 @@
+swagger: '2.0'
+info:
+    title: 'FBS Adapter'
+    version: '1.0'
+basePath: /
+paths:
```
kasperg, line 6: Please add some from of documentation regarding how we ended up with this local file and how it should be maintained.

I image this could either be in the form of a markdown file in the `docs` dir or a code comment at the top of the file.

src/core/fbs/mutator/fetcher.ts
```diff
@@ -0,0 +1,64 @@
...
+
+const baseURL = "https://fbs-openplatform.dbc.dk"; // use your own URL here or environment variable
+
+export const fetcher = async <ResponseType>({
+  url,
+  method,
+  headers,
+  params,
+  data
+}: {
```
kasperg, line 11: With this PR, #42 and #43 we are now at three fetchers which (should) look about the same apart from which base url and token is used.

At some point we should look into creating something that is more easily shareable.

cableman: Agreed :-) Think that there will be a small rewrite to re-use fetchers in the next sprint when the current once are merged. I'll add it to the next meeting schedule.

docker-compose.yml
cableman: Fixed

src/core/fbs/fbs-adapter.yaml
cableman: Added document with information

spaceo (commented):
Why is this PR against main?

src/apps/adgangsplatformen/user-token.dev.tsx
```diff
@@ -0,0 +1,12 @@
+import { ComponentMeta, ComponentStory } from "@storybook/react";
+import React from "react";
+import UserToken from "./user-token";
+
+export default {
+  title: "Set User Token",
```
spaceo, line 6: I think this should be organized under: "SB Utilities / ..."
Like [here](https://github.com/danskernesdigitalebibliotek/dpl-react/blob/release/2/src/apps/adgangsplatformen/library-token.dev.tsx#L6)

spaceo (changes requested):

src/apps/adgangsplatformen/user-token.dev.tsx
cableman: It did not exists when I made this... but will move it

spaceo: Great be aware that it only implies a change in the title. No files need to be moved ;)

cableman:
Force pushed just to update the ticket no. in the commits.

PR #45, opened 2022-06-28
Feature/advanced autosuggest

Adamik10:
#### Link to issue
image

#### Description
This PR will adds the rest of the functionality to our autosuggest that was not included in the basic version:
- updates the DOM structure and classes based on the latest version of the design system npm package
- uses useText() from [#27](https://github.com/danskernesdigitalebibliotek/dpl-react/pull/27) instead of prop drilling. 
- adds material covers 
- if there are no materials, don't show the material suggestion part of the dropdown at all
- fixes the problem with wrong order of items when using down key even if work suggestions came before position 7, 8, 9
- doesn't call the API service on change of the highlighted item, but still updates the input field  
- fixes typescript TODO errors
- simulates form submission when the user selects an item using enter

#### Screenshot of the result
image

#### Checklist


Adamik10:
Moved to PR #53

PR #46, opened 2022-06-29
Availability Labels

kasperbirch1:
#### Description
This component is used in the search-results-item to display the availability of a material.
It Takes an array of IDs and if one of them is available it will show that the material is available

There is a big change in fetcher, because we need to build url parameter in a certain way.

#### Screenshot of the result
Skærmbillede 2022-06-29 kl  10 00 05

#### Checklist


src/components/availability-label/availability-label.dev.tsx
```diff
@@ -42,19 +47,29 @@ const Template: ComponentStory<typeof AvailabilityLabel> = (
...
-  availabilityText: "online",
-  state: "unavailable",
-  link: undefined
+  availabilityText: "online"
+};
+
+export const harryPotter = Template.bind({});
+harryPotter.args = {
+  materialId: ["62523611"]
 };
```
kasperg, line 75: Please remove this story.

I do not see a reason for it to be added. It does not seem to add anything to the stories already available.

src/components/availability-label/availability-label.test.js
```diff
@@ -0,0 +1,60 @@
+describe("Availability Label", () => {
+  it("shows that at least one of the materials is available", () => {
+    cy.intercept("GET", "**/external/agencyid/catalog/availability/v3?**", {
```
kasperg, line 3:
```suggestion
    cy.intercept("GET", "**/external/agencyid/catalog/availability/v3?recordid=62523611&recordid=62150041&recordid=61435867", {
```

Consider including the query parameters in the mocked request.

Since we have unusual requirements regarding the formatting of these arguments in the query string it may be worthwhile to also require that specific formatting in the test.

src/components/availability-label/availability-label.tsx
```diff
@@ -2,46 +2,60 @@ import React from "react";
...
+  materialId
 }) => {
-  const triangleState = {
-    available: "success",
-    unavailable: "alert",
-    selected: "alert"
-  };
+  const { data } = useGetAvailabilityV3({
+    recordid: materialId
+    // exclude: [""]
```
kasperg, line 25: Please avoid uncommented code in VCS.

src/components/availability-label/availability-label.tsx
```diff
@@ -2,46 +2,60 @@ import React from "react";
...
 import clsx from "clsx";
 import { Link } from "../utils/link";
+import { useGetAvailabilityV3 } from "../../core/fbs/fbs";
 
 export interface AvailabilityLabelProps {
   manifestText: string;
   availabilityText: string;
-  state: "available" | "unavailable" | "selected";
+  selected: boolean;
+  state: string;
```
kasperg, line 11: Please remove the `state` prop. It does not seem as if we use it at all.

src/components/availability-label/availability-label.tsx
```diff
@@ -2,46 +2,60 @@ import React from "react";
...
+    recordid: materialId
+    // exclude: [""]
+  });
+
+  // Checking for Error,
+  // If user is not authorized, data is an object with an error message.
+  // Therefore check if data is an array
+  if (!Array.isArray(data)) {
+    return null;
+  }
```
kasperg, line 33: Please use `isError` from React Query to determine if we have an error or not.

src/components/availability-label/availability-label.tsx
```diff
@@ -2,46 +2,60 @@ import React from "react";
...
-  state: "available" | "unavailable" | "selected";
+  selected: boolean;
+  state: string;
   link: string | undefined;
+  materialId: string[];
 }
 
 export const AvailabilityLabel: React.FC<AvailabilityLabelProps> = ({
   manifestText,
   availabilityText,
```
kasperg, line 18: Please remove the `availabilityText` prop.

We now determine availability status from a service and have `useText()` for static texts. We can use these to set the availability text to `t('Available')` or `t('Unavailable')` depending on the response from FBS.

spaceo (approved):

cableman (approved):

PR #52, opened 2022-07-05
FBI graphql changes  added

spaceo:
#### Link to issue

https://reload.atlassian.net/browse/DDFSOEG-165

#### Description

After the graphql API changed to the new FBI API we needed to change queries, types and functionalities in the autosuggest and search result part of the code

#### Checklist

#### Note
Old demo components/apps were deleted as well since we do not need them anymore.


src/components/search-result-list/search-result-list-item/search-result-list-item-series.tsx
```diff
@@ -2,17 +2,19 @@ import React from "react";
...
 interface SearchResultListItemSeriesProps {
-  series: SeriesSimpleFragment;
+  series: SeriesSimpleFragment[];
 }
 
 const SearchResultListItemSeries: React.FC<SearchResultListItemSeriesProps> = ({
-  series: { part, title }
+  series
 }) => {
+  const display = series?.[0]?.numberInSeries?.display;
```
Adamik10, line 11: I maybe would call this const something that is meaningful. In the data, it is called display, that's right, but I would maybe suggest calling it something like numberInSeries or stng ike that.

src/core/fbs/mutator/fetcher.ts
```diff
@@ -1,4 +1,4 @@
-import { getToken, TOKEN_USER_KEY } from "../../token";
+import { getToken, TOKEN_LIBRARY_KEY, TOKEN_USER_KEY } from "../../token";
```
Adamik10, line 1: @spaceo what about these GitHub warnings?
'TOKEN_USER_KEY' is defined but never used

Adamik10 (changes requested):
Very good job 🥇 You've done well with this mildly annoying API change task. I have a couple of comments + one that i couldn't comment in connection to the file -> src/core/dbc-gateway/CreatorSimple.fragment.graphql -> this file is empty. What is it?

src/components/search-result-list/search-result-list-item/search-result-list-item-series.tsx
spaceo: 👍

src/core/fbs/mutator/fetcher.ts
spaceo: 👍

cableman (approved):

PR #54, opened 2022-07-06
Add function that can extract params

spaceo:
Either from props or from url parameters

#### Description

We have several places where we need to check if data attributes are set -> if not check url params.

#### Checklist


kasperbirch1:
LGTM 👍

cableman (approved):

PR #55, opened 2022-07-11
Feature/5387 loan page

sinejespersen:
#### Link to issue

#### Description

#### Screenshot of the result

#### Checklist


src/apps/loan-list/list/loan-list.tsx
```diff
@@ -0,0 +1,159 @@
+import React, { useEffect, useState } from "react";
+import dayjs from "dayjs";
+import localizedFormat from "dayjs/plugin/localizedFormat";
```
spaceo, line 3: Great dayjs looks like a nifty tool. I think it would be a good idea to mention when we introduce a global tool like that in Zulip. I probably forgot it myself so this is also a reminder for myself.

src/apps/loan-list/list/loan-list.tsx
```diff
@@ -0,0 +1,159 @@
...
+  useEffect(() => {
+    dayjs.extend(localizedFormat);
+  }, []);
+
+  useEffect(() => {
+    const searchParams = new URLSearchParams(window.location.search);
+    // regex for finding date string from modal query param
+    const regex = /^\d{4}-\d{2}-\d{2}$/;
+    // modal query param
+    const modalString = searchParams.get("modal");
```
spaceo, line 31: We have a helper function for that I think we should use:
https://github.com/reload/dpl-react/blob/feature/dpl-react-disclosure-DDFSOEG-168/src/core/utils/helpers.ts#L12

src/apps/loan-list/list/loan-list.tsx
```diff
@@ -0,0 +1,159 @@
...
+
+    if (modalString && loans) {
+      const found = modalString.toString().match(regex);
```
spaceo, line 34: Could we move the date string matching in a global helper function as well? We could probably use that multiple places.

src/apps/loan-list/list/loan-list.tsx
```diff
@@ -0,0 +1,159 @@
...
+      // If there is a date string in the modal query param
+      if (found) {
+        setDueDateModal(found[0]);
+        // THe loans are filtered with said date string
+        const loansForModal = loans.filter(
```
spaceo, line 39: You could consider move logic like this into functions in a file like: src/apps/loan-list/helpers.ts.
The function could be called filterLoansByDate.
You have quite a lot of logic in the component and the action of moving it into functions could serve multiple purposes:
* It is easier to get an overview of the component
* The functions could be used in combination with other functions
* The functionality is ready to be unit tested if we should want that

src/apps/loan-list/list/loan-list.tsx
```diff
@@ -0,0 +1,159 @@
...
+
+        // Loans for modal (the modal shows loans stacked by due date)
+        setLoansModal(loansForModal);
+        setRenewable(amountOfRenewableLoans);
+      }
+    }
+  }, [loans, dueDateModal]);
+
+  useEffect(() => {
+    getLoansV2().then((value) => {
```
spaceo, line 55: I think we should use the [useGetLoansV2](https://github.com/danskernesdigitalebibliotek/dpl-react/blob/c36e34c7dc23455b18459b23a68a1104364bf535/src/core/fbs/fbs.ts#L1061) hooks here. That is how we are doing in the other places where use one of the API's.
In that way we get a unified way of requesting no matter if we are an Orval based client or graphql-codegen generated client.
The useXxx hooks uses react-query under the hood, which gives as a lot of nice tools we can use.
Here is an example using another service from fbs:
https://github.com/danskernesdigitalebibliotek/dpl-react/blob/release/3/src/components/availability-label/availability-label.tsx#L22

src/apps/loan-list/list/loan-list.tsx
```diff
@@ -0,0 +1,159 @@
...
+        setLoans(sortedByDueDate);
+        setAmountOfLoans(sortedByDueDate.length);
+      }
+    });
+  }, []);
+
+  function openModalDueDate(dueDateModalInput: string) {
+    if (loans) {
+      setDueDateModal(dueDateModalInput);
+      setLoansModal(
```
spaceo, line 80: Helper function maybe?

src/apps/loan-list/list/loan-list.tsx
```diff
@@ -0,0 +1,159 @@
...
+        loans.filter(
+          ({ loanDetails }) => loanDetails.dueDate === dueDateModalInput
+        )
+      );
+    }
+  }
+
+  return (
+    <div aria-hidden={loansModal !== null ?? true}>
```
spaceo, line 89: Shouldn't this be a string?
Maybe:
```tsx
<div aria-hidden={loansModal ?? "true"}>
```
But according to [this](https://github.com/danskernesdigitalebibliotek/dpl-react/pull/55/files#diff-2cdd62020bea6d73b0ed5c5b7ead29e571bdb06dedcfeb649598dcc202fe42c3R17) loansModal is: `LoanV2[] | null`.
I am confused :)

src/apps/loan-list/list/loan-list.tsx
```diff
@@ -0,0 +1,159 @@
...
+        type="button"
+        aria-describedby={t("loanListRenewMultipleButtonExplanationText")}
+        className="btn-primary btn-medium  btn-filled arrow__hover--right-small"
+      >
+        {t("loanListRenewMultipleButtonText")}
+      </button>
+      {loans && (
+        <div className="list-reservation-container">
+          {view === "stacked" &&
+            dueDates?.map((uniqueDueDate) => {
```
spaceo, line 111: Why the `dueDates ` check here?
According to [this](https://github.com/danskernesdigitalebibliotek/dpl-react/pull/55/files#diff-2cdd62020bea6d73b0ed5c5b7ead29e571bdb06dedcfeb649598dcc202fe42c3R15) it is always an array of stirngs

src/apps/loan-list/list/loan-list.tsx
```diff
@@ -0,0 +1,159 @@
...
+              // Stack items:
+              // if multiple items have the same due date, they are "stacked"
+              // which means styling making it look like there are multiple materials,
+              // but only _one_ with said due date is visible.
+              const loan = loans.filter(
```
spaceo, line 116: Helper function

src/apps/loan-list/list/loan-list.tsx
```diff
@@ -0,0 +1,159 @@
...
+      <button
+        type="button"
+        aria-describedby={t("loanListRenewMultipleButtonExplanationText")}
+        className="btn-primary btn-medium  btn-filled arrow__hover--right-small"
+      >
+        {t("loanListRenewMultipleButtonText")}
+      </button>
+      {loans && (
+        <div className="list-reservation-container">
+          {view === "stacked" &&
```
spaceo, line 110: You could consider creating two sub components called something like: 
LoanListItems and LoanListItemsStacked that is shown depending on the `view`.
That would reduce the size of this component/file.

src/apps/loan-list/materials/manifestation.graphql
```diff
@@ -0,0 +1,20 @@
+query getMaterialManifestation($faust: String!) {
```
spaceo, line 1: This query name is a bit to generic in the entire scope of queries. I think you need to be more specific. We are also using material manifestations in other places and maybe we do not need the same set of properties.

src/apps/loan-list/materials/material-decorator.tsx
```diff
@@ -0,0 +1,100 @@
...
+  GetMaterialManifestationQuery,
+  useGetMaterialManifestationQuery
+} from "../../../core/dbc-gateway/generated/graphql";
+import SelectableMaterial from "./selectable-material";
+import StackableMaterial from "./stackable-material";
+import { useText } from "../../../core/utils/text";
+
+interface MaterialDecoratorProps {
+  materialType: string;
+  faust: string;
```
spaceo, line 12: We have a faust type here: https://github.com/reload/dpl-react/blob/feature/dpl-react-disclosure-DDFSOEG-168/src/core/utils/types/ids.ts#L1

src/apps/loan-list/materials/material-decorator.tsx
```diff
@@ -0,0 +1,100 @@
...
+  renewableStatus,
+  amountOfMaterialsWithDueDate,
+  selectDueDate,
+  loanDate
+}) => {
+  const t = useText();
+  const [material, setMaterial] = useState<GetMaterialManifestationQuery>();
+
+  // Create a string of authors with commas and a conjunction
+  const getAuthorName = (
```
spaceo, line 45: Is this function doing the same as [this](https://github.com/reload/dpl-react/blob/feature/dpl-react-disclosure-DDFSOEG-168/src/components/search-result-list/helpers.ts#L47).
If yes we should consolidate them.

src/apps/loan-list/materials/stackable-material.tsx
```diff
@@ -0,0 +1,131 @@
...
+      </div>
+      <div>
+        <div className="list-reservation__status">
+          <StatusCircle loanDate={loanDate} dueDate={dueDate} />
+          <div>
+            <div className="list-reservation__deadline">
+              <StatusBadge dueDate={dueDate} />
+              <p className="text-small-caption">
+                {t("LoanListToBeDeliveredText")}{" "}
+                {dayjs(dueDate).format("DD-MM-YYYY")}
```
spaceo, line 103: It think this definitely should be moved into a global function

src/apps/loan-list/materials/utils/checkbox.tsx
```diff
@@ -0,0 +1,44 @@
...
+  label,
+  hideLabel,
+  additionalClasses
+}) => {
+  return (
+    // todo add space between label adn checkbox
+    <div className={`checkbox ${additionalClasses}`}>
+      <label className="checkbox__label" htmlFor={id}>
+        <span className="checkbox__icon">
+          <svg width="20px" height="20px">
```
spaceo, line 21: Could we move this svg into a separate svg file?

src/core/configuration/status-thresholds.json
```diff
@@ -0,0 +1,4 @@
+{
```
spaceo, line 1: Great :) 💯

spaceo (changes requested):

src/apps/loan-list/list/loan-list.tsx
sinejespersen: Noted.

src/apps/loan-list/list/loan-list.tsx
sinejespersen: Yes, where? And also, I am not sure all the dates will match this regex? the reason I put it here is because I know it will match duedate from fbs as the string is "YYYY-MM-DD" - but I guess the data will be different in different places?

src/apps/loan-list/list/loan-list.tsx
sinejespersen: I honestly have no idea why i didnt do that to begin with

src/apps/loan-list/materials/utils/checkbox.tsx
sinejespersen: Yes. In the design system?

src/apps/loan-list/materials/material-decorator.tsx
sinejespersen: It doesn't "et al" the string.

src/apps/loan-list/list/loan-list.tsx
sinejespersen: image

I'll have a look at it

src/apps/loan-list/list/loan-list.tsx
spaceo: Either in src/core/utils/helpers.ts or we could create a folder called: src/core/utils/helpers with a file called src/core/utils/helpers/timeNdates.ts or something similar.
And the function could be called `dateMatchesUsFormat(date: string): boolean`

src/apps/loan-list/materials/manifestation.graphql
spaceo: All the querys are used by graphql-codegen and end in the same scope in src/core/dbc-gateway/generated/graphql.tsx.
If we are specific we will not get name collisions although we are using the same type of queries eg: `manifestation(faust: $faust) {`
Regarding properties we use fragments in order to define sets of reusable properties which the codegen tool also creates types from. But this is a subject that is more easy to explain p2p :)

src/apps/loan-list/materials/material-decorator.tsx
spaceo: Then maybe we could expand the function like:
```ts
export const creatorsToString = (creators: string[], t: UseTextFunction, useEtAlSuffix: boolean = false) => {

File: src/apps/loan-list/materials/utils/checkbox.tsx (hunk 15 above)
Comment (spaceo) [reply src/apps/loan-list/materials/utils/checkbox.tsx:15]: Either that or like this:
[src/components/icon-favourite/icon-favourite.tsx](https://github.com/reload/dpl-react/blob/dpl-react-materiale-udgave-kort_description-list_DDFSOEG-169/src/components/icon-favourite/icon-favourite.tsx)

**Review by spaceo — CHANGES_REQUESTED** [review]
Made some replies to some of the conversations...

File: src/apps/loan-list/helpers.ts [hunk 17]
```diff
@@ -0,0 +1,27 @@
...
+
+export const getAmountOfRenewableLoans = (list: LoanV2[]) => {
+  return list.filter(({ isRenewable }) => isRenewable).length;
+};
+
+export const formatDate = (date: string) => {
+  return dayjs(date).format("DD-MM-YYYY");
+};
+
+export const materialOverdue = (date: string) => {
```
spaceo, line 23: Really a minor, but I think if a function returns a boolean it should be called something like:
`materialIsOverdue` to emphasize the purpose of it

src/apps/loan-list/materials/material-decorator.tsx
```diff
@@ -0,0 +1,77 @@
...
+  description: string;
+  fullTitle: string;
+  datePublished: unknown;
+  materialType: string;
+  creators: { __typename?: "Creator" | undefined; name: string }[];
+}
+
+const MaterialDecorator: React.FC<MaterialDecoratorProps> = ({
+  materialType,
+  recordId,
```
spaceo, line 33: Why are you not calling it `faustId: FaustId` when that is what it is?

spaceo (approved):
👍

PR #56, opened 2022-07-12
Add Material header

kasperbirch1:
#### Link to issue

#### Description

#### Screenshot of the result

#### Checklist


src/components/material/MaterialHeader.tsx
```diff
@@ -0,0 +1,60 @@
...
+      <div className="material-header__content">
+        <ButtonFavourite materialId={pid} />
+        <MaterialHeaderText title={title} author={allAuthorsString} />
+        <div className="material-header__availability-label">
+          <AvailabiltityLabels manifestations={manifestations} />
+        </div>
+        {showPeriodikumSelect && <MaterialPeriodikumSelect />}
+        <div className="material-header__button">
+          <Button
+            label="RESERVER BOG"
```
spaceo, line 39: Should be translatable

src/components/material/MaterialHeader.tsx
```diff
@@ -0,0 +1,60 @@
...
+            buttonType="none"
+            variant="filled"
+            disabled={false}
+            collapsible={false}
+            size="large"
+          />
+          <Button
+            label="FIND PÅ HYLDEN"
```
spaceo, line 47: Should be translatable

src/components/material/MaterialHeaderText.tsx
```diff
@@ -0,0 +1,20 @@
+import React from "react";
+
+interface MaterialHeaderTextProps {
+  title: string;
+  author: string;
+}
+
+const MaterialHeaderText = ({ title, author }: MaterialHeaderTextProps) => {
```
spaceo, line 8: Remember FC<> type here

src/components/material/MaterialHeaderText.tsx
```diff
@@ -0,0 +1,20 @@
...
+  return (
+    <>
+      <h1 className="text-header-h1 mb-16">{title}</h1>
+      <p className="text-body-large">
+        <span>Af </span>
```
spaceo, line 13: Should be translatable

src/components/material/MaterialPeriodikumSelect.tsx
```diff
@@ -0,0 +1,43 @@
+import React from "react";
+
+const MaterialPeriodikumSelect: React.FC = () => {
+  const placeholderData = {
+    year: ["2017", "2018", "2019", "2020", "2021", "2022"],
+    weeks: ["30", "31", "32", "33", "34", "35"]
```
spaceo, line 6: Should be replaced by real data before it can be merged

src/components/material/MaterialPeriodikumSelect.tsx
```diff
@@ -0,0 +1,43 @@
...
+  };
+
+  return (
+    <div className="text-small-caption material-periodikum ">
+      <div className="material-periodikum-select">
+        {/* This is because the design requires label and select input to be separated */}
+        {/* eslint-disable-next-line jsx-a11y/label-has-associated-control */}
+        <label htmlFor="year">Årgang</label>
```
spaceo, line 14: Translatable...

src/components/material/MaterialPeriodikumSelect.tsx
```diff
@@ -0,0 +1,43 @@
...
+                {year}
+              </option>
+            ))}
+          </select>
+        </div>
+      </div>
+      <div className="material-periodikum-select">
+        {/* This is because the design requires label and select input to be separated */}
+        {/* eslint-disable-next-line jsx-a11y/label-has-associated-control */}
+        <label htmlFor="weeks">Uge</label>
```
spaceo, line 28: Translatable

spaceo (changes requested):
👍 Needs changes

src/components/material/MaterialHeaderText.tsx
```diff
@@ -9,11 +10,13 @@ const MaterialHeaderText: React.FC<MaterialHeaderTextProps> = ({
...
   author
 }) => {
+  const t = useText();
+
   return (
     <>
       <h1 className="text-header-h1 mb-16">{title}</h1>
       <p className="text-body-large">
-        <span>Af </span>
+        <span>{t("Af")} </span>
```
spaceo, line 19: The text key needs to:
* Be in english
* Be camel case
* End with "Text" suffix
* Have a name that ensures that it does not duplicate other text strings (unless it is intentional).
So in this case it could be: "MaterialHeaderAuthorByText"

src/apps/material/material.tsx
```diff
@@ -17,8 +17,14 @@ const Material: React.FC<MaterialProps> = ({ pid, workId }) => {
     return <div>Loading...</div>;
   }
 
+  if (!data?.work) {
+    return <div>No work data</div>;
```
spaceo, line 21: DDF wont accept that we write an english string like this.
If we choose this solution I think the text should be translatable 🎉

spaceo (changes requested):

spaceo:
One last thing (to be even more irritating):
On the app entry level you need to specify all the texts being used as a part of the props.
See the [search-result](https://github.com/reload/dpl-react/blob/feature/dpl-react-disclosure-DDFSOEG-168/src/apps/search-result/search-result.entry.tsx#L7-L20) as an example.
In the same example you can also see how the entry component is wrapped with: `withtext()`
And then you put all the texts as controllable args as well, as seen here:
https://github.com/reload/dpl-react/blob/feature/dpl-react-disclosure-DDFSOEG-168/src/apps/search-result/search-result.dev.tsx#L20:L20

src/components/Buttons/Button.tsx
```diff
@@ -0,0 +1,74 @@
...
+  const getVariant = () => {
+    if (variant === "filled") return "btn-filled";
+    if (variant === "outline") return "btn-outline";
+    return "";
+  };
+
+  return (
+    <button
+      type="button"
+      className={`btn-primary ${getVariant()} ${getSize()} arrow__hover--right-small`}
```
spaceo, line 66:
```suggestion
      className={`btn-primary ${getVariant()} btn-${size}} arrow__hover--right-small`}
```

sinejespersen (approved):

PR #58, opened 2022-07-13
Add disclosure component

spaceo:
To be used in the bottom of the material page

#### Link to issue

https://reload.atlassian.net/browse/DDFSOEG-168

#### Description

It is a very simple component with the proper classes.

#### Screenshot of the result

image

#### Checklist

### DIFF
--- a/src/components/material/disclosures/disclosure.tsx
+++ b/src/components/material/disclosures/disclosure.tsx
@@ -1,0 +1,35 @@
+import ExpandMoreIcon from "@reload/dpl-design-system/build/icons/collection/ExpandMore.svg";
+import React, { FC, ReactNode } from "react";
+import { useText } from "../../../core/utils/text";
+
+export interface DisclosureProps {
+  svgIcon: { path: string; altText: string };
+  title: string;
+  children?: ReactNode;
+}
+
+export const Disclosure: FC<DisclosureProps> = ({
+  title,
+  children,
+  svgIcon: { path, altText }
+}) => {
+  const t = useText();
+  return (
+    <details className="disclosure text-body-large">
+      <summary className="disclosure__headline text-body-large">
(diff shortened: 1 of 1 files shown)


kasperbirch1 (approved):

src/components/material/disclosures/disclosure.tsx
```diff
@@ -0,0 +1,35 @@
...
+    <details className="disclosure text-body-large">
+      <summary className="disclosure__headline text-body-large">
+        <div className="disclosure__icon bg-identity-tint-120 m-24">
+          <img className="disclosure__icon" src={path} alt={altText} />
+        </div>
+        {title}
+        <img
+          className="disclosure__expand mr-24 noselect"
+          src={ExpandMoreIcon}
+          alt={t("disclosureIconExpandAltText")}
```
sinejespersen, line 27: I would definitely alt="" this

src/components/material/disclosures/disclosure.tsx
```diff
@@ -0,0 +1,35 @@
...
+  title,
+  children,
+  svgIcon: { path, altText }
+}) => {
+  const t = useText();
+  return (
+    <details className="disclosure text-body-large">
+      <summary className="disclosure__headline text-body-large">
+        <div className="disclosure__icon bg-identity-tint-120 m-24">
+          <img className="disclosure__icon" src={path} alt={altText} />
```
sinejespersen, line 21: I'd definitely hide both img from screen readers -> alt=""

spaceo: But alt, in this line, comes from `svgIcon.altText`.
In the other image the alt text comes from t.
But on second thought I think both texts should come as parameters, since this component is a simple rendering component

spaceo: So. Now it is changed

PR #60, opened 2022-07-14
Feature/material description

kasperbirch1:
#### Description

This PR 
- adds MaterialDescription to Material
- rewrites fragments.graphql 
- change name and parameter for for useGetWork to  useGetMaterialQuery

MaterialDescription render series, seriesMembers, subjects in the SeriesLine component

#### Screenshot of the result
Skærmbillede 2022-07-14 kl  13 30 11

#### Checklist


src/components/material/MaterialDescription.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+  return (
+    <section className="material-description">
+      <h2 className="text-header-h4 pb-24">Beskrivelse</h2>
+      <p className="text-body-large material-description__content">
+        {description}
+      </p>
+      <div className="material-description__links mt-32">
+        <HorizontalTermLine
+          title={`Nr. ${work?.series[0]?.numberInSeries?.number?.[0]}`}
```
sinejespersen, line 30: "nr." as props?

src/components/material/MaterialDescription.tsx
```diff
@@ -0,0 +1,45 @@
...
+  const t = useText();
+
+  const seriesMembersList = work.seriesMembers.map(
+    (item) => item.titles.main[0]
+  );
+  const subjectsList = work.subjects.all.map((item) => item.display);
+
+  return (
+    <section className="material-description">
+      <h2 className="text-header-h4 pb-24">Beskrivelse</h2>
```
sinejespersen, line 24: Beskrivelse af props

src/components/horizontal-term-line/HorizontalTermLine.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+const HorizontalTermLine: React.FC<HorizontalTermLineProps> = ({
+  title,
+  subTitle,
+  linkList
+}) => {
+  return (
+    <div className="text-small-caption horizontal-term-line">
+      <p className="text-label-semibold">
+        {`${title}`}{" "}
```
sinejespersen, line 17:
```suggestion
        {title}{" "}
```

src/apps/material/material.tsx
```diff
@@ -24,6 +25,10 @@ const Material: React.FC<MaterialProps> = ({ pid }) => {
   return (
     <main>
       <MaterialHeader pid={pid} work={data.work} />
+      <MaterialDescription
+        work={data.work}
+        description="Stormandssønnen Arn får hos cisterciensermunkene i Sverige og Danmark den bedste uddannelse, der findes i middelalderens Europa. Men hans lærere aner, at han ikke er bestemt til at være klosterbroder og vil gøre bedre fyldest som Kristi strids... "
```
sinejespersen, line 30: Is this some sort of example?

src/components/search-result-list/search-result-list-item/search-result-list-item.tsx
```diff
@@ -72,7 +72,13 @@ const SearchResultListItem: React.FC<SearchResultListItemProps> = ({
       <div className="search-result-item__text">
         <div className="search-result-item__meta">
           <ButtonFavourite materialId={workId} />
-          {series && <SearchResultListItemSeries series={series} />}
+          {series && (
+            <HorizontalTermLine
+              title={`Nr. ${series[0].numberInSeries?.number?.[0]}`}
```
sinejespersen, line 77: same as above.

src/apps/material/material.tsx
spaceo, line 30: Try look into the `PhysicalDescriptions` on a `Manifestation`

src/components/horizontal-term-line/HorizontalTermLine.tsx
```diff
@@ -0,0 +1,33 @@
...
+        {subTitle && <span className="text-small-caption">{subTitle} </span>}
+      </p>
+      <ul className="horizontal-term-line__list">
+        {linkList?.map((item) => (
```
spaceo, line 21: linkList has the type: `string[]` and is required. So no optional chaining check is needed here.

src/components/horizontal-term-line/HorizontalTermLine.tsx
```diff
@@ -0,0 +1,33 @@
...
+          <li>
+            <a href="/" className="link-tag">
```
spaceo, line 23: We should use the link component

src/components/material/MaterialDescription.tsx
```diff
@@ -0,0 +1,45 @@
...
+  description?: string;
+}
+
+const MaterialDescription: React.FC<MaterialDescriptionProps> = ({
+  work,
+  description
+}) => {
+  const t = useText();
+
+  const seriesMembersList = work.seriesMembers.map(
```
spaceo, line 17: I think you should add a note that we are not sure but we assume that titles.main is the right place to find the title. We should do this globally becasuse we have looked it up in search result items as well

src/components/material/MaterialDescription.tsx
spaceo: That should be translatable

src/components/material/MaterialDescription.tsx
spaceo, line 30: You have to look into the schema to see what you can expect form data. There is no need for checking every single property.
Also I think it would be prettier to:
* Create a component that is called something like: SeriesLine.
* SeriesLine accepts one prop called: series
* The SeriesLine component destruct the needed parameters like:
```tsx
const SeriesLine: React.FC<SeriesLineProps> = ({
  numberInSeries: { number, display: linkList }
}) => {
``

src/components/search-result-list/search-result-list-item/search-result-list-item.tsx
spaceo, line 77: Same comment as above.

spaceo (changes requested):
Looks great! ;)
But I think there could be some improvements.

PR #62, opened 2022-07-18
Feature/5387 viewport hook

sinejespersen:
#### Link to issue

#### Description

So, a button is supposed to travel from the top of the page to the bottom of the page once it is not visible any longer due to scroll. So this is a hook to determine whether a element is in view, and the hook is used to determine whether the bottom button is visible.

#### Screenshot of the result

scroll

#### Checklist


src/components/utils/in-viewport.tsx
```diff
@@ -0,0 +1,32 @@
+import { RefObject, useEffect, useRef, useState, useCallback } from "react";
+
+function useIsInViewport(offset = 0): [boolean, RefObject<HTMLInputElement>] {
```
kasperg, line 3: Please consider using one of the many community implementations of this.

Examples:
- https://www.npmjs.com/package/react-cool-inview
- https://www.npmjs.com/package/use-is-in-viewport
- https://www.npmjs.com/package/react-hook-inview

Apart from leveraging other peoples work all these options also use [IntersectionObserverAPI which the recommended approach to handling such things](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API). [It is supported by all the browsers that we target](https://caniuse.com/intersectionobserver).

src/apps/loan-list/list/loan-list.test.tsx
```diff
@@ -612,6 +612,225 @@ describe("Loan list", () => {
...
+    cy.visit("/iframe.html?path=/story/apps-loan-list--loan-list-entry");
+    cy.wait(["@loans", "@work", "@cover"]);
+
+    cy.get("#test-stack").click();
+    cy.get("#test-more-materials").click();
+    cy.get(".modal").find(".modal-loan__buttons").should("exist");
+    cy.get(".modal").find(".modal-loan__buttons--bottom").should("not.exist");
+    // Add duration to scroll, if I don't it scrolls to fast for the eventlistener
+    // to work...
+    cy.get(".modal-loan__container").scrollTo("bottom", { duration: 50 });
```
kasperg, line 827: :+1: for adding an explanation here.

src/apps/loan-list/modal/renew-loans-modal-content.tsx
```diff
@@ -9,21 +9,29 @@ import {
...
 const RenewLoansModalContent: FC<RenewLoansModalContentProps> = ({
   renewable,
   loansModal,
   checkboxLabel,
-  buttonLabel
+  buttonLabel,
+  buttonBottomLabel,
+  checkboxBottomLabel
 }) => {
+  const t = useText();
```
kasperg, line 32: Please remember to remove `t` if it is not used.

src/apps/loan-list/modal/renew-loans-modal-content.tsx
```diff
@@ -131,6 +139,24 @@ const RenewLoansModalContent: FC<RenewLoansModalContentProps> = ({
             );
           })}
         </ul>
+        {!topButtonsVisible && (
+          <div
+            className="modal-loan__buttons modal-loan__buttons--bottom"
+            style={{ position: "fixed", left: 0, bottom: 0, right: 0 }}
```
kasperg, line 145: Please move styling to the design system or add a comment describing why it is necessary in this case.

Normally we do not use inline styling and based on the current implementation and commit comments I do not see any apparent reason to do so here either.

As far as I can tell it could be implemented by adding a new modifier to the `modal-loan__buttons` element.

src/apps/loan-list/modal/renew-loans-modal-content.tsx
```diff
@@ -131,6 +139,24 @@ const RenewLoansModalContent: FC<RenewLoansModalContentProps> = ({
...
+              onChecked={selectAll}
+              id="checkbox-select-all"
+              label={checkboxBottomLabel}
+            />
+            <button
+              type="button"
+              className="btn-primary btn-filled btn-small arrow__hover--right-small"
+            >
+              {buttonBottomLabel}({renewable})
+            </button>
```
kasperg, line 157: Consider if this could use our `Button` component.

kasperg (changes requested):
:+1: I have some comments that I would like you to look into.

src/apps/loan-list/modal/renew-loans-modal-content.tsx
sinejespersen: oh I forgot this, I have actually moved it to the design system already 

https://github.com/danskernesdigitalebibliotek/dpl-design-system/blob/3b1e783a8809165988bbb14b81bc0ff857734872/src/stories/Library/Modals/modal-loan/modal-loan.scss#L53

Good catch, I'll remove it from here.

src/apps/loan-list/modal/renew-loans-modal-content.tsx
sinejespersen: Definitely, will fix

src/apps/loan-list/modal/renew-loans-modal-content.tsx
sinejespersen: Definitely, will fix

kasperg (approved):
:+1: Excellent

PR #66, opened 2022-07-29
Add material page description to the material app

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFSOEG-184

#### Description
This PR adds all of the currently available static HTML to the material app with updated styling from the dpl-design-system package, and then slices the description part of the app into its own sub-component that makes a data call to populate itself based on the currently viewed material (based on PID). 
It also adds material description headline as one of the app mount props. This way the headline text can be adjusted from the CMS interface when mounted in. 
Outside of the ticket scope, I also added the option to set xlarge option for Cover component to match our design system package, and the Figma design, as I was already working with the whole material app anyways.

#### Screenshot of the result
image

#### Checklist

### DIFF
--- a/src/apps/material/material.dev.tsx
+++ b/src/apps/material/material.dev.tsx
@@ -12,28 +12,28 @@ export default {
       control: { type: "text" }
     },
     materialHeaderAuthorByText: {
-      name: "Af forfatter",
+      name: "By (author)",
       defaultValue: "Af ",
       control: { type: "text" }
     },
     periodikumSelectYearText: {
-      name: "År",
+      name: "Year",
       defaultValue: "År",
       control: { type: "text" }
     },
     periodikumSelectWeekText: {
-      name: "Uge",
+      name: "Week",
       defaultValue: "Uge",
       control: { type: "text" }
     },
     reserveBookText: {
-      name: "Reserve book",
+      name: "Reserve",
       defaultValue: "RESERVER BOG",
       control: { type: "text" }
     },
-    fineOnBookshelfText: {
-      name: "Fine on bookshelf",
-      defaultValue: "FINE PÅ HYLDEN",
+    findOnBookshelfText: {
+      name: "Find on bookshelf",
+      defaultValue: "FIND PÅ HYLDEN",
       control: { type: "text" }
     }
   }
--- a/src/apps/material/material.dev.tsx
+++ b/src/apps/material/material.dev.tsx
@@ -35,6 +35,11 @@ export default {
       name: "Find on bookshelf",
       defaultValue: "FIND PÅ HYLDEN",
       control: { type: "text" }
+    },
+    descriptionHeadlineText: {
+      name: "Description headline",
+      defaultValue: "Beskrivelse",
+      control: { type: "text" }
     }
   }
 } as ComponentMeta<typeof MaterialEntry>;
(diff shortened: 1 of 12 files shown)


yepzdk (commented):
Added a couple of comments.
I can't really give you any qualified feedback on your graphql implementation.
Maybe another reviewer should have a look also.
But if this is blocking you, I don't se any problems with merging.

src/components/cover/cover.tsx
```diff
@@ -21,7 +21,12 @@ export const Cover = ({
   tint,
   pid
 }: CoverProps) => {
-  const dataSize: CoverProps["size"] = size === "xsmall" ? "small" : size;
+  let dataSize: CoverProps["size"] = size;
+  if (dataSize === "xsmall") {
```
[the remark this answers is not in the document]
Adamik10: It's because for now we don't receive xsmall for example from the API. There the smallest is small. But xsmall keyword is used for css classes.

src/apps/material/material.tsx
```diff
@@ -22,9 +27,147 @@ const Material: React.FC<MaterialProps> = ({ pid }) => {
   }
 
   return (
-    <main>
+    <div className="material-page">
       <MaterialHeader pid={pid} work={data.work} />
-    </main>
+      <MaterialDescription pid={pid} />
+      <details className="disclosure text-body-large">
```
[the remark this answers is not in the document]
Adamik10: It is :) 
The end product of the design system is our storybook where we can copy HTML markup from, and a css file with all the styling from the repository. We import this styling in the dpl-react repository as an npm package and then use it to style the application without having to write any css here.

src/apps/material/material.dev.tsx
```diff
@@ -12,28 +12,33 @@ export default {
...
+      name: "Reserve",
       defaultValue: "RESERVER BOG",
       control: { type: "text" }
     },
-    fineOnBookshelfText: {
-      name: "Fine on bookshelf",
-      defaultValue: "FINE PÅ HYLDEN",
+    findOnBookshelfText: {
+      name: "Find on bookshelf",
+      defaultValue: "FIND PÅ HYLDEN",
```
kasperg, line 36: Consider if we should capitalize the text using CSS instead of static text.

Note that this outside the scope of this PR and also applies to other buttons.

src/apps/material/material.tsx
```diff
@@ -22,9 +27,147 @@ const Material: React.FC<MaterialProps> = ({ pid }) => {
   }
 
   return (
-    <main>
+    <div className="material-page">
```
kasperg, line 30: I see that this is according to the design system. I general I think it is OK to use the `<main>` tag here for an app we assume will be the main content of a page.

kasperg (approved):
:+1: I general I would have suggested an approach where we gradually builds from nothing to a fully dynamic page. This gradually works from a fully static page to a fully static page.

src/apps/material/material.dev.tsx
Adamik10: You're right. Very good suggestion!. 
I have added a utility class that does this in the other PR I have open in the design system.

src/apps/material/material.tsx
Adamik10: This makes sense. I corrected it :)

Adamik10:
Follow this PR further at #67

PR #67, opened 2022-08-01
Feature/topic tags

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFSOEG-158
https://reload.atlassian.net/browse/DDFSOEG-157
image
image

#### Description
This PR focuses on the material page description, connected series information and topic tags.

I also added all of the currently available static HTML to the material app with updated styling from the dpl-design-system package. Apart from new components I also added some new app props so that the libraries can use the CMS to specify translations for these texts. As a part of the material page, I also added the option to set xlarge option for Cover component to match our design system package, and the Figma design, none of which were updated yet.
Both series info and topic tags blocks are only shown if there is any data present. A search is performed upon clicking the individual identifiers using the base url and the identifier value. 

#### Screenshot of the result
image

#### Checklist

+ #66 is included in this - everything up until commit 88d6cba should NOT be reviewed as a part of this PR.


src/components/material-description/MaterialDescription.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+export interface MaterialDescriptionProps {
+  pid: Pid;
+}
+
+const MaterialDescription: React.FC<MaterialDescriptionProps> = ({ pid }) => {
+  const { data, isLoading } = useGetManifestationQuery({ pid });
+  const t = useText();
+  const baseUrl = t("baseUrlText");
+  const identifierText = t("identifierText");
```
kasperg, line 15: Please to not declare this texts separately.

If these texts are not used multiple places I think it is fine just to inline them in the JSX. The name of the `t()` function has been chosen because it is sufficiently short.

src/components/material-description/MaterialDescription.tsx
```diff
@@ -0,0 +1,73 @@
...
+                <span key={generateMapId(index)}>
+                  <a
+                    href={`${baseUrl}?q=${identifier.value}`}
+                    className="link-tag"
+                  >
+                    {identifier.value}
+                  </a>
+                </span>
+              );
+            })}
```
kasperg, line 65: Consider if the HTML structure can be improved.

I know that this change belongs in the design system but for future reference I would have expected something like:

```html
<h2>Indentifiers</h4>
<ul>
  <li>
    <a href="search url">Identifier</a>
  </li>
</ul>
```

src/apps/material/material.dev.tsx
```diff
@@ -12,28 +12,43 @@ export default {
...
+      control: { type: "text" }
+    },
+    descriptionHeadlineText: {
+      name: "Description headline",
+      defaultValue: "Beskrivelse",
+      control: { type: "text" }
+    },
+    baseUrlText: {
+      name: "Base search url",
+      defaultValue: "https://bibliotek.dk/search",
```
kasperg, line 46: Consider if the default value be the url of the search app in storyboook.

src/core/utils/helpers.ts
```diff
@@ -135,3 +135,8 @@ export const dateMatchesUsFormat = (date: string | null) => {
   const returnValue = dateFound && dateFound.length > 0 ? dateFound[0] : null;
   return returnValue;
 };
+
+export function generateMapId(index: number | string) {
+  const randomId = Math.random().toString(36);
+  return randomId.concat(index.toString());
+}
```
kasperg, line 142: Please reconsider this approach. [Random does not seem to be a good approach to React keys](https://www.developerway.com/posts/react-key-attribute).

src/apps/material/material.dev.tsx
```diff
@@ -12,28 +12,43 @@ export default {
...
+      control: { type: "text" }
+    },
```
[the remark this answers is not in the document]
Adamik10: Done :)

src/components/material-description/MaterialDescription.tsx
```diff
@@ -0,0 +1,73 @@
...
+import { Pid } from "../../core/utils/types/ids";
+
+export interface MaterialDescriptionProps {
+  pid: Pid;
+}
+
+const MaterialDescription: React.FC<MaterialDescriptionProps> = ({ pid }) => {
+  const { data, isLoading } = useGetManifestationQuery({ pid });
+  const t = useText();
+  const baseUrl = t("baseUrlText");
```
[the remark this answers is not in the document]
Adamik10: done :) great point!

src/components/material-description/MaterialDescription.tsx
Adamik10: 👍 💯

src/components/material-description/MaterialDescription.tsx
Adamik10: Good point! I need to focus on this a bit more in the future.

src/apps/material/material.dev.tsx
Adamik10: That could actually be nice! I added it :)

src/core/utils/helpers.ts
Adamik10: Uuuu very interesting! I was living in the dark for so long 😝 Thanks!

kasperg (approved):
:+1: Excellent.

PR #68, opened 2022-08-02
Improve accessibility for search header

kasperg:
#### Link to issue

https://reload.atlassian.net/browse/DDFSOEG-176

#### Description

Accessibility checks at danskernesdigitalebibliotek/dpl-cms#118 shows a problem where an ARIA property is added to the form element where it is not valid.

The ARIA role comes from Downshift `getInputProps()`. This PR moves it from the `<form>` to a `<div/>`  which wraps the `<input/>` element. This is in accordance with [Downshift documentation and examples](https://www.downshift-js.com/use-combobox).

As far as I remember we have experienced problems with using a form as the outer element. My initial testing shows that this is not the case.

The PR also includes a bit of boyscouting in relation to the search header.

#### Screenshot of the result

https://user-images.githubusercontent.com/73966/182362468-53bf3374-5757-46a9-9575-63b738ce57c4.mp4

#### Checklist

### DIFF
--- a/src/apps/search-header/search-header.tsx
+++ b/src/apps/search-header/search-header.tsx
@@ -182,11 +182,7 @@ const SearchHeader: React.FC = () => {
   });
 
   return (
-    <form
-      className="header__menu-second"
-      action={t("searchHeaderUrlText")}
-      id="autosuggestForm"
-    >
+    <form className="header__menu-second" action={t("searchHeaderUrlText")}>
       {/* The downshift combobox uses prop spreading by design */}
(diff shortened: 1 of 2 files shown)


Adamik10 (approved):
Perfect. Thanks for fixing this :) I think it all looks as should + it also seems to work so I don't have any comments. 
But maybe one thing you could try is to update the dpl-design-system package (1.5.0). If none of the design breaks, we should be fine :)

kasperg:
> But maybe one thing you could try is to update the dpl-design-system package (1.5.0). If none of the design breaks, we should be fine :)

I have checked and I do not see any breakage.

PR #69, opened 2022-08-03
Only cross-compile code to modern browsers

kasperg:
#### Link to issue

https://reload.atlassian.net/browse/DDFSOEG-176

#### Description

Performance checks at https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/118 shows that we are running into performance problems due to the amount of JavaScript we deliver to clients.

One way to reduce this is to not cross compile our code to legacy browsers as this will require a lot of polyfills. [This is also a suggestion mentioned by Lighthouse](https://storage.googleapis.com/lighthouse-infrastructure.appspot.com/reports/1659519745506-91116.report.html#legacy-javascript). According to our coding guidelines we do not have to support legacy browsers.

This PR updates our `.browserslists` configuration to match our coding guidelines.

In general we should be aware of our bundle size. The current tool included in the project is not in actual use and also seems to be abandoned. Consequently I have removed it. 

#### Checklist

### DIFF
--- a/.browserslistrc
+++ b/.browserslistrc
@@ -1,17 +1,9 @@
-# https://github.com/browserslist/browserslist#full-list
-# Browsers that we support
-# cover 90% in DK 06-11-2019
-and_chr 76
-chrome 77
-chrome 76
-edge 18
-firefox 69
-ie 11
-ios_saf 13.0-13.1
-ios_saf 12.2-12.3
-ios_saf 12.0-12.1
-ios_saf 11.3-11.4
-safari 13
-safari 12.1
-samsung 10.1
-samsung 9.2
+# These browsers match our coding guidelines
+# https://github.com/danskernesdigitalebibliotek/dpl-react/blob/main/docs/code_guidelines.md#general
+last 2 chrome versions
+last 2 edge versions
+last 2 firefox versions
+last 2 safari versions
+last 2 ios_saf versions
+last 2 and_chr versions
+last 2 samsung versions
--- a/package.json
+++ b/package.json
@@ -76,6 +76,7 @@
     "babel-loader": "8.2.5",
     "babel-plugin-istanbul": "^6.1.1",
     "babel-plugin-transform-react-remove-prop-types": "^0.4.24",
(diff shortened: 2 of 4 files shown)


kasperg:
The result of this has reduced the size of our `dist.zip` build from 613KB in [the latest version of release 3](https://github.com/danskernesdigitalebibliotek/dpl-react/releases/tag/release-release%2F3) to 506KB in [the version for the branch release](https://github.com/reload/dpl-react/releases/tag/release-feature%2Freduce-bundle-size).

.browserslistrc
```diff
@@ -1,17 +1,9 @@
...
-samsung 9.2
+# These browsers match our coding guidelines
+# https://github.com/danskernesdigitalebibliotek/dpl-react/blob/main/docs/code_guidelines.md#general
+last 2 chrome versions
+last 2 edge versions
+last 2 firefox versions
+last 2 safari versions
+last 2 ios_saf versions
+last 2 and_chr versions
+last 2 samsung versions
```
Adamik10, line 9: Beautiful that you checked up on this and we got rid of so many unnecessary browsers 💯

package.json
```diff
@@ -77,6 +76,7 @@
     "babel-loader": "8.2.5",
     "babel-plugin-istanbul": "^6.1.1",
     "babel-plugin-transform-react-remove-prop-types": "^0.4.24",
+    "caniuse-lite": "^1.0.30001373",
```
Adamik10, line 79: So caniuse-lite is the new @bundle-analyzer/webpack-plugin, right? Because it's more lightweight.

Adamik10 (approved):
Good stuff! Have we also measured how much it helps? Or that can only happen after we set it into the CMS? I have a couple of comments but they are just retorical questions or praise.

package.json
kasperg: Can `caniuse-lite` is a database that Babel with `babel/preset-env` uses during the Webpack build process.

kasperg:
> Have we also measured how much it helps?

Not in the sense that we can compare Lighthouse results with and without this change.

PR #70, opened 2022-08-03
Search header fixes

kasperg:
#### Description

This fixes a couple of issues found during testing of the search header. 

Please check out the individual commits for more commentary.

#### Checklist


Adamik10 (approved):
This is wonderful! Thanks for the quick fix!

PR #72, opened 2022-08-05
Feature/material details

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFSOEG-171

#### Description
This PR adds a new Disclosure showing material details. 

For now, we show the data that is available in a work. but later logic must be made to determine which manifestation will show the missing fields. 

#### Screenshot of the result
Skærmbillede 2022-08-05 kl  11 22 36

#### Checklist

### DIFF
--- a/src/apps/material/material.tsx
+++ b/src/apps/material/material.tsx
@@ -11,6 +11,11 @@ import ListDescription, {
   ListData
 } from "../../components/list-description/list-description";
 import { useText } from "../../core/utils/text";
+import {
+  creatorsToString,
+  filterCreators,
+  flattenCreators
+} from "../../core/utils/helpers";
 
 export interface MaterialProps {
   pid: Pid;
@@ -41,15 +46,20 @@ const Material: React.FC<MaterialProps> = ({ pid, searchUrl }) => {
     workYear
   } = data.work;
 
+  const creatorsText = creatorsToString(
+    flattenCreators(filterCreators(creators, ["Person"])),
+    t
+  );
+
   const listDescriptionData = {
     Type: {
-      value: String(materialTypes?.[0]?.specific),
+      value: materialTypes?.[0]?.specific,
       type: "standard"
     },
-    Sprog: { value: String(mainLanguages?.[0].display), type: "standard" },
-    Bidragsydere: { value: String(creators?.[0].display), type: "link" },
+    Sprog: { value: mainLanguages?.[0].display, type: "standard" },
+    Bidragsydere: { value: creatorsText, type: "link" },
     Originaltitel: {
-      value: String(`${titles.original} ${workYear}`),
+      value: `${titles?.original} ${workYear}`,
       type: "standard"
     }
     // TODO: Logic must be created to select the manifestation to be presented for the rest of listDescriptionData
--- a/src/apps/material/material.tsx
+++ b/src/apps/material/material.tsx
@@ -52,23 +52,26 @@ const Material: React.FC<MaterialProps> = ({ pid, searchUrl }) => {
   );
 
   const listDescriptionData = {
-    Type: {
+    [t("typeText")]: {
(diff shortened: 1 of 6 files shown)


src/apps/material/material.tsx
```diff
@@ -28,6 +37,40 @@ const Material: React.FC<MaterialProps> = ({ pid, searchUrl }) => {
...
+    flattenCreators(filterCreators(creators, ["Person"])),
+    t
+  );
+
+  const listDescriptionData = {
+    Type: {
+      value: materialTypes?.[0]?.specific,
+      type: "standard"
+    },
+    Sprog: { value: mainLanguages?.[0].display, type: "standard" },
```
kasperg, line 59: Consider if we should show all values for language and types.

As far as I can tell this will only show the first value is shown but our schema indicates that there may be more.

src/apps/material/material.tsx
```diff
@@ -48,6 +91,16 @@ const Material: React.FC<MaterialProps> = ({ pid, searchUrl }) => {
           );
         })}
       </Disclosure>
+      <Disclosure
+        mainIconPath={Receipt}
+        title="Detaljer"
```
kasperg, line 96: Remember to make texts translatable.

src/apps/material/material.tsx
```diff
@@ -28,6 +37,40 @@ const Material: React.FC<MaterialProps> = ({ pid, searchUrl }) => {
...
+    workYear
+  } = data.work;
+
+  const creatorsText = creatorsToString(
+    flattenCreators(filterCreators(creators, ["Person"])),
+    t
+  );
+
+  const listDescriptionData = {
+    Type: {
```
kasperg, line 55: Consider if these keys should be translatable and managed with `t()`.

I now realize that they here and elsewhere are in fact part of the user interface static texts. In that case they should be managed with `t()`.

kasperg (changes requested):
:+1: Please look into handling or single/multiple values for list data and translations.

src/components/material/MaterialMainfestationItem.tsx
```diff
@@ -50,19 +50,29 @@ const MaterialMainfestationItem: FC<MaterialMainfestationItemProps> = ({
...
+    [t("typeText")]: { value: materialTypes?.[0]?.specific, type: "standard" },
+    [t("languageText")]: {
+      value: allLanguages,
+      type: "standard"
+    },
+    [t("contributorsText")]: { value: allContributors, type: "link" },
+    [t("originalTitleText")]: { value: titles?.original, type: "standard" },
+    [t("isbnText")]: { value: identifiers?.[0].value, type: "standard" },
+    [t("editionText")]: { value: edition?.summary, type: "standard" },
+    [t("ScopeText")]: {
```
kasperg, line 66:
```suggestion
    [t("scopeText")]: {
```

For consistency

src/components/material/MaterialMainfestationItem.tsx
```diff
@@ -50,19 +50,29 @@ const MaterialMainfestationItem: FC<MaterialMainfestationItemProps> = ({
...
       value: physicalDescriptions[0].numberOfPages,
       type: "standard"
     },
-    Forlag: { value: hostPublication?.publisher, type: "standard" },
-    Målgruppe: { value: audience?.generalAudience, type: "standard" }
+    [t("publisherText")]: {
+      value: hostPublication?.publisher,
+      type: "standard"
+    },
+    [t("AudienceText")]: { value: audience?.generalAudience, type: "standard" }
```
kasperg, line 74:
```suggestion
    [t("audienceText")]: { value: audience?.generalAudience, type: "standard" }
```

kasperg (changes requested):
:+1: on improving translateability. A few consistency comments. Otherwise we should be good to go.

PR #74, opened 2022-08-05
See this ! https://github.com/danskernesdigitalebibliotek/dpl-design-system/pull/120

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFSOEG-177

This must be done at the same time as https://github.com/danskernesdigitalebibliotek/dpl-design-system/pull/120 !!!

Remember to change "@danskernesdigitalebibliotek/dpl-design-system": to the new version

#### Description
Material has changed its name to Cover in the dpl-design-system so we have to change class names here too

#### Checklist


kasperg (approved):
:+1: LGTM

PR #75, opened 2022-08-08
Material reviews

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFSOEG-173
image

#### Description
This PR adds components related to material app reviews, and graphql queries are adjusted accordingly. For each review the following shows if the data exists:
- metadata (author and date)
- amount of hearts given to the work
- headline
- body text
- link to external review

There is a special kind of review - infomedia review - where we have to make a data call to fetch the content separately. In order for the user to be able to see these infomedia reviews they need to be logged in and their account needs to be connected to the municipality compatible with the review. The API service checks for whether this is true. In this PR, we were unable to retrieve infomedia reviews due to an internal server error. DDF was contacted and is investigating this, but we might have to get back to this in the future. I was able to see what the data looks like from FBI API documentation based on my own user token from beta.bibliotek.dk, but in our app I can't get it to work.

#### Screenshot of the result
image

#### Checklist

- What do you think of the implementation?


src/components/material/ReviewExternal.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+const ReviewExternal: React.FC<ReviewExternalProps> = ({ review }) => {
+  return (
+    <li className="review text-small-caption">
+      {(review.author || review.date) && (
+        <ReviewMetadata author={review.author} date={review.date} />
+      )}
+      {review.rating && <ReviewHearts amountOfHearts={review.rating} />}
+      {review.urls &&
+        review.urls.length >= 1 &&
```
kasperg, line 18: Consider if this condition is necessary. If `urls` is an empty array then it should just be `map`'ed 0 times.

src/components/material/ReviewExternal.tsx
```diff
@@ -0,0 +1,30 @@
...
+import ReviewHearts from "./ReviewHearts";
+import ReviewMetadata from "./ReviewMetadata";
+
+export interface ReviewExternalProps {
+  review: ExternalReview;
+}
+
+const ReviewExternal: React.FC<ReviewExternalProps> = ({ review }) => {
+  return (
+    <li className="review text-small-caption">
```
kasperg, line 12: Consider changing this to `<div>`. 

This is the classic list/item question. Should the item (`ReviewExternal`) be aware that it is always used within a list and thus use `<li>` as a wrapper or not and then use a `<div>` which then will result in more markup as the structure will become.

```html
<ul>
  <li>
     <div class="review"/>
  </li>
</ul>
 ```

File: src/components/material/ReviewHearts.tsx [hunk 3]
```diff
@@ -0,0 +1,35 @@
...
+  );
+  const emptyHeartsArray = Array.from(
+    { length: Number(heartArray[1]) - Number(heartArray[0]) },
+    (_, i) => i + 1
+  );
+  return (
+    <div
+      className="mb-4"
+      role="figure"
+      aria-label="Rating of this item is 5 out of 6"
```
kasperg, line 23: Please make this text translatable and vary according to `filledHeartsArray.length` and `heartArray.length`.

src/components/material/ReviewInfomedia.tsx
```diff
@@ -0,0 +1,52 @@
...
+        </div>
+      )}
+      {data?.infomedia.article?.text && (
+        <p className="review__body mb-8">{data?.infomedia.article?.text}</p>
+      )}
+      {review.origin && (
+        <a href={review.origin} className="link-tag text-small-caption mb-8">
+          {`${review.author}${
+            review.author && review.date ? ", " : ""
+          }${review.date?.split("-").reverse().join(".")}`}
```
kasperg, line 45: Consider whether proper date handling can replace string manipulation.

I would expect something like creating a `Date` object from the string and then formatting the resulting object according to our wishes. We may need a library to help us out here.

I recommend https://date-fns.org/ but @spaceo may also have an opinion here.

Whether you do please move handling into a separate function as this occurs multiple places.

spaceo:
> ... DDF was contacted and is investigating this, but we might have to get back to this in the future.
I guess it is certain that we have to get back to this functionality in the future if the Infomedia reviews does not work. You were not able to see how the data was structured then, right? And the Infomedia data is supposed to be the most "complex" data, as far as I remember?

kasperg (changes requested):
:+1: Well done. I have a single issue I would like you to fix and a couple of things for you to consider.

src/components/material/ReviewInfomedia.tsx
```diff
@@ -0,0 +1,52 @@
...
+const ReviewInfomedia: React.FC<ReviewInfomediaProps> = ({ review }) => {
+  const { id } = review;
+  const { data } = useGetInfomediaQuery({
+    id
+  });
+  const t = useText();
+  return (
+    <li className="review text-small-caption">
+      {(review.author || review.date) && (
+        <ReviewMetadata author={review.author} date={review.date} />
```
spaceo, line 23: I think we should use date objects as properties whenever possible. It is much easier to do date calculations and test like that. So the date should be transformed into a date object before sending it to the ReviewMetadata.
You could create a helper function, eg. fromReviewDateToDateObject in a file called: src/components/material/helpers.ts.
Then you can use datejs to create a proper date object. Like Sine does here:
https://github.com/danskernesdigitalebibliotek/dpl-react/pull/55/files#diff-5796af082eb3b5376779c529940e1f83e906738f33dedb630a37d5205e4d7f89R20

src/components/material/ReviewInfomedia.tsx
spaceo: Yes. I think you should use dayjs as described earlier. But I also think that you could use `ReviewMetadata` in both places. It seems like the same thing is happening with checking author and date and formatting date.

spaceo (changes requested):
Looks fab. I have a few comments though.

src/components/material/ReviewExternal.tsx
Adamik10: Great comment! Thanks :) I removed it

src/components/material/ReviewExternal.tsx
Adamik10: That is a very fair point. 
For now though, we don't need to reuse these reviews any other places in different contexts. So I will take this with me forward as a point to consider, but in this case the reviews are list items :)

src/components/material/ReviewInfomedia.tsx
Adamik10: Perf, yes ;) Thanks for the suggestion 💯 Fixing it.

src/components/material/ReviewInfomedia.tsx
Adamik10: Thanks a lot for the extensive recommendation! 🏅 💯 I adjusted it so that we work with Date objects instead ;)

src/core/utils/helpers.ts
```diff
@@ -135,3 +135,16 @@ export const dateMatchesUsFormat = (date: string | null) => {
   const returnValue = dateFound && dateFound.length > 0 ? dateFound[0] : null;
   return returnValue;
 };
+
+export const usDateStringToEurDateObj = (date: string | null | undefined) => {
+  if (!date || date === null) {
+    return date;
+  }
```
kasperg, line 142: Consider if you can drop the ` | null | undefined` and the following check. It seems as if you can move the call to `usDateStringToEurDateObj` to `ReviewMetadata` and let the `date` argument` accept `string | null | undefined`.

Also: I am not sure there is any such thing as a Eur(o) date object. It's just a date object.

src/core/utils/helpers.ts
```diff
@@ -135,3 +135,16 @@ export const dateMatchesUsFormat = (date: string | null) => {
...
+  const dateArray = date?.split("-");
+  const euDate = new Date(
+    Number(dateArray[0]),
+    Number(dateArray[1]) - 1,
+    Number(dateArray[2])
+  );
+  return euDate;
```
kasperg, line 149: Consider if this can be replaced by

```js
var customParseFormat = require('dayjs/plugin/customParseFormat');
dayjs.extend(customParseFormat);
return dayjs(date, 'YYYY-MM-DD');
```
https://day.js.org/docs/en/plugin/custom-parse-format

kasperg (changes requested):
I appreciate the intention but I think the result could use push.

src/components/material/ReviewInfomedia.tsx
kasperg: Consider if you can use the the new helper here.

Adamik10: Snap, forgot to use it here too 🎖️ 🎖️ 🎖️ Thx

src/core/utils/helpers.ts
Adamik10: 👍

src/core/utils/helpers.ts
Adamik10: Perf! Thanks for the suggestion 💯

src/components/material/ReviewMetadata.tsx
```diff
@@ -15,14 +16,29 @@ export const usDateStringToDateObj = (date: string): Date | null => {
...
-      }`}
-    </div>
-  );
+const ReviewMetadata: React.FC<ReviewMetadataProps> = ({
+  author,
+  date,
+  link
+}) => {
+  const metaDataText = (
+    returnAuthor: string | null | undefined,
```
spaceo, line 22: This is nitty gritty, but why the "return" prefix on the parameters? Are they necesarry?
Also, you `metaDataText` is actually a functional component. So you could also move that out to a separate `<MetaDataText />` component.

spaceo (changes requested):

kasperg (approved):
:+1: LGTM

src/components/material/ReviewMetadata.tsx
Adamik10: The "return" prefix is there because there cannot be duplicates in one file, and we already have both Author and Date. And It is not a component because it returns just a string. Even if we put it into its own component and return that string in a React fragment, Eslint and typescript will complain. I am adamant in keeping it as is in this case.

src/components/material/ReviewMetadata.tsx
```diff
@@ -15,14 +16,29 @@ export const usDateStringToDateObj = (date: string): Date | null => {
...
+    returnDate: Date | null | undefined
+  ) => {
+    return `${returnAuthor || ""}${returnAuthor && returnDate ? ", " : ""}${
+      returnDate ? dayjs(returnDate).format("DD.MM.YYYY") : ""
+    }`;
+  };
+
+  if (link) {
+    return (
+      <a href={link} className="link-tag text-small-caption mb-8">
```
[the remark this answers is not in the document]
Adamik10: Thanks! Am using it now ;)

src/components/material/ReviewInfomedia.tsx
```diff
@@ -25,14 +25,12 @@ const ReviewInfomedia: React.FC<ReviewInfomediaProps> = ({ review }) => {
...
+  const date = review.date ? usDateStringToDateObj(review.date) : null;
   if (infomedia.error) {
     return (
       <li className="review text-small-caption">
         {(review.author || review.date) && (
-          <ReviewMetadata
-            author={review.author}
-            date={usDateStringToDateObj(review.date as string)}
-          />
+          <ReviewMetadata author={review.author} date={date} />
```
spaceo, line 33: Nice! More clean :)

src/components/material/ReviewMetadata.tsx
spaceo: Since it is parameters in a function they have their own scope and won't conflict with the variables outside of it.
Weird that eslint/typescript complains about a string in a fragment. What kind of error does it throw?

src/components/material/ReviewMetadata.tsx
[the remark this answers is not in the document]
spaceo: Sweet 👍

src/components/material/ReviewInfomedia.tsx
Adamik10: ❤️

kasperg (approved):
:+1: LGTM

src/components/material/ReviewMetadata.tsx
Adamik10: image

spaceo (approved):
LGTM 👍

PR #76, opened 2022-08-08
Make search header linkable to search result and material app

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFSOEG-185

#### Description
This PR makes sure that linking from the search header app to search result app/material app works when the apps are set into the CMS.  
This needed to be distinguished based on the autosuggest item the user selected. If it is one of the string suggestions - we redirect to the search result; if it is a material suggestion - we redirect to the material app page. 

#### Screenshot of the result
n/a

#### Checklist

n/a

### DIFF
--- a/src/apps/search-header/search-header.dev.tsx
+++ b/src/apps/search-header/search-header.dev.tsx
@@ -7,9 +7,9 @@ export default {
   title: "Apps / Search Header",
   component: SearchHeaderEntry,
   argTypes: {
-    searchHeaderUrlText: {
+    baseUrl: {
       name: "Search header base URL",
-      defaultValue: "https://bibliotek.dk/search",
+      defaultValue: "https://bibliotek.dk",
       control: { type: "text" }
     },
     altText: {
--- a/src/apps/search-header/search-header.tsx
+++ b/src/apps/search-header/search-header.tsx
@@ -96,12 +96,17 @@ const SearchHeader: React.FC<SearchHeaderProps> = ({ baseUrl }) => {
 
   // downshift prevents the default form submission event when the autosuggest
   // is open - that's why in some cases we have to simulate form sumbission
-  function manualRedirect(inputValue: string) {
-    const baseUrl = t("searchHeaderUrlText");
-    const params = inputValue;
-    if (window.top) {
-      window.top.location.href = `${baseUrl}?q=${params}`;
+  function manualRedirect(itemValue: Suggestion) {
+    const inputValue = itemToString(itemValue);
+    if (!window.top) {
+      return;
     }
+    // if this item is one of the work suggestions, redirect to material page
+    if (materialData.includes(itemValue)) {
+      window.top.location.href = `${baseUrl}/work/${itemValue.work?.manifestations.first.pid}`;
+      return;
+    }
+    window.top.location.href = `${baseUrl}/search?q=${inputValue}`;
   }
 
   function handleHighlightedIndexChange(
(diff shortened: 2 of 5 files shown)


src/apps/search-header/search-header.tsx
```diff
@@ -98,12 +96,17 @@ const SearchHeader: React.FC = () => {
...
-    if (window.top) {
-      window.top.location.href = `${baseUrl}?q=${params}`;
+  function manualRedirect(itemValue: Suggestion) {
+    const inputValue = itemToString(itemValue);
+    if (!window.top) {
+      return;
+    }
+    // if this item is one of the work suggestions, redirect to material page
+    if (materialData.includes(itemValue)) {
+      window.top.location.href = `${baseUrl}/work/${itemValue.work?.manifestations.first.pid}`;
```
kasperg, line 106: Please clear this path with @spaceo .

I think it is a mistake. The data we are showing on the work page is a work - not a material. If that is is the case the path should be generated from `itemValue.work?.workId`.

src/apps/search-header/search-header.dev.tsx
```diff
@@ -7,9 +7,9 @@ export default {
   title: "Apps / Search Header",
   component: SearchHeaderEntry,
   argTypes: {
-    searchHeaderUrlText: {
+    baseUrl: {
       name: "Search header base URL",
-      defaultValue: "https://bibliotek.dk/search",
+      defaultValue: "https://bibliotek.dk",
```
kasperg, line 12: Consider introducing separate props from the material page and the search page.

The current approach integrates logic of routing structures of the mounting app into the React components. We would like to avoid this. One way to do this is to have separate props for search results and materials.

Another aspect of this would be placeholder support for querys, ids etc. so we can define where to put query, (work) id etc. in the prop value. [In the current React components this is implemented with `:placeholder`](https://github.com/danskernesdigitalebibliotek/ddb-react/blob/master/src/core/replacePlaceholders.js). This would make it possible to have the search prop be something `/search?q=:query` in the CMS but something like `https://danskernesdigitalebibliotek.github.io/dpl-react/iframe.html?args=query::query&id=apps-search-result--search-result&viewMode=story` in Storybook.

src/apps/search-header/search-header.tsx
```diff
@@ -98,12 +96,17 @@ const SearchHeader: React.FC = () => {
...
-  function manualRedirect(inputValue: string) {
-    const baseUrl = t("searchHeaderUrlText");
-    const params = inputValue;
-    if (window.top) {
-      window.top.location.href = `${baseUrl}?q=${params}`;
+  function manualRedirect(itemValue: Suggestion) {
+    const inputValue = itemToString(itemValue);
+    if (!window.top) {
+      return;
+    }
```
kasperg, line 103: Consider throwing an error if this occurs.

I do not know why `window.top` would not be set but with this approach it fails silently. This can make it hard to debug. Throwing an error may make it easier for us to debug in the future.

kasperg (changes requested):
:+1: One issue for you to check and a few things for you to consider.

src/apps/search-header/search-header.tsx
Adamik10: Mmmmm but it actually is a material. Try to check material.tsx file - the query is called useGetMaterial() and it accepts a pid, as opposed to workId. 
The reason for this that there always has to be one specific material selected (availability labels on the top left part of the material header)
image But I'll talk to Mikkel tomorrow ;)

Adamik10:
Follow PR #89 instead

PR #77, opened 2022-08-09
Add logic to display language in title

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFSOEG-148

#### Description

This change checks if the language contains "dansk" and if not it will be displayed together with the title.

## Uddrag fra mindstekrav:

c. Title-full - og såfremt værket er en bog, som ikke er på dansk - sprog i parentes efter Title-full

#### Screenshot of the result

**This is a reverse example because there are only languages that contain "dansk"**
Skærmbillede 2022-08-09 kl  14 00 20

#### Checklist


kasperg (approved):
:+1: LGTM

PR #78, opened 2022-08-10
Reference TypeScript instead of JavaScript in documentation

kasperg:
#### Description

This should clarify the primary language of the project.

#### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -418,7 +418,7 @@ this:
 </details>
 
 As a minimum you will need the `runtime.js` and `bundle.js`. For styling
-of atoms and components you will need to import `components.css`.
+of components you will need to import `components.css`.
 
 Each application also has its own JavaScript artifact and it might have a CSS
 artifact as well. Such as `add-to-checklist.js` and `add-to-checklist.css`.
@@ -515,7 +515,7 @@ can reuse various levels of infrastructure provided by the project such as:
 
 1. [Integration with various webservices](src/core)
 2. [User authentication and token management](src/core)
-3. [Visual atoms or components](#cross-application-components)
+3. [Visual components](#cross-application-components)
 4. Visual representations of [existing applications](src/apps)
 5. [Styling using SCSS](#style-your-application)
 6. Test infrastructure
--- a/README.md
+++ b/README.md
@@ -175,7 +175,7 @@ export default MyNewApplicationEntry;
 <details>
   <summary>3. Create the mount</summary>
 
-```javascript
+```typescript
 // ./src/apps/my-new-application/my-new-application.mount.ts
 import addMount from "../../core/addMount";
 import MyNewApplication from "./my-new-application.entry";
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -278,7 +275,7 @@ appropriate.
 
 ```scss
 // ./src/apps/my-new-application/my-new-application.scss
-.dpl-warm {
+.warm {
```
hrmoller, line 278: Doesn't the removal of `dpl-` contradict with what's stated in [`code_guidelines.md` line 94](https://github.com/reload/dpl-react/blob/docs-typescript/docs/code_guidelines.md?plain=1#L94)?

hrmoller (commented):
A few typos... I think:)

- Besides what I consider typos which has been commented on on the specific lines I have found a few places in the README where the term "atoms" is still present even though it seems as the usage of that term is discontinued.
  1. [On line 421](https://github.com/reload/dpl-react/blob/docs-typescript/README.md?plain=1#L421)
  2. [On line 518](https://github.com/reload/dpl-react/blob/docs-typescript/README.md?plain=1#L518)

- In the [Files section in code_guidelines.md](https://github.com/reload/dpl-react/blob/docs-typescript/docs/code_guidelines.md?plain=1#L94) there's being referred to `.js` and `.jsx` files but should that be changed to `.ts` and `.tsx`?

- It might be outside the scope of this PR but would it at a later point be worth considering how to define types if there's a need for defining types outside the scope of a component?

README.md
```diff
@@ -343,130 +337,11 @@ __Cowabunga!__ You now got styling in your application
 ### Cross application components
 
 If the component is simple enough to be a primitive you would use in multiple
-occasions it's called an 'atom'. Such as a button or a link. If it's more
-specific that that and to be used across apps we just call it a component. An
-example would be some type of media presented alongside a header and some text.
+occasions it's called n component.
```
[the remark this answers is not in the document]
kasperg: It is a type. I have changed it to "a component".

README.md
kasperg: It does. I have reverted this change in [e4f8186](https://github.com/danskernesdigitalebibliotek/dpl-react/pull/78/commits/e4f81862eee59c08b6a037e9744dee68762e90ef).

kasperg:
>  I have found a few places in the README where the term "atoms" is still present even though it seems as the usage of that term is discontinued

I agree about atom being discontinued in the project. I have removed references in [d1066ff](https://github.com/danskernesdigitalebibliotek/dpl-react/pull/78/commits/d1066ff476410b3c0daf6bfcd566feb5a56768bc).

> there's being referred to .js and .jsx files but should that be changed to .ts and .tsx?

I agree. This has been changed in [4c1731f](https://github.com/danskernesdigitalebibliotek/dpl-react/pull/78/commits/4c1731f1f8978be1d3221766ec0e56e593ab248a).

> [...] would it at a later point be worth considering how to define types if there's a need for defining types outside the scope of a component?

I agree. In current cross-project types are being added to `src/core`.

hrmoller (approved):
Look good to me😊

PR #79, opened 2022-08-10
Fix reloading covers for autosuggest - material suggestions

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFSOEG-191

#### Description
Using autosuggest material item workId as key for mapped suggestion items makes sure that mapped element's keys stay the same between specific items across reloads. Thus the items won't be re-rendered when they the items stay the same - thus the cover pictures won't flash.

#### Screenshot of the result
https://user-images.githubusercontent.com/28546954/183887537-343c6ca4-791f-4830-a65e-125ab3e3d9d5.mov

#### Checklist

n/a

### DIFF
--- a/src/components/autosuggest-material/autosuggest-material.tsx
+++ b/src/components/autosuggest-material/autosuggest-material.tsx
@@ -31,7 +31,7 @@ const AutosuggestMaterial: React.FC<AutosuggestMaterialProps> = ({
         <ul className="autosuggest__materials">
           {/* eslint-disable react/jsx-props-no-spreading */}
           {/* The downshift combobox works this way by design (line 50) */}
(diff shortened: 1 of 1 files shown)


src/components/autosuggest-material/autosuggest-material.tsx
```diff
@@ -47,7 +47,7 @@ const AutosuggestMaterial: React.FC<AutosuggestMaterialProps> = ({
                     ? "autosuggest__material--highlight"
                     : ""
                 }`}
-                key={generateItemId(item)}
+                key={item.work?.workId}
```
kasperg, line 50: Please remove the `generateItemId()` import which is no longer used.

kasperg (approved):
:+1: LGTM

PR #80, opened 2022-08-10
Fix button favourite

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFSOEG-34

#### Description
This PR is a review of button favorite compared to MK. 

The button has been changed so that it now uses a specific type of id instead of just a string

- Variation 3a has not been made

#### Checklist


kasperg (approved):
:+1: LGTM

PR #81, opened 2022-08-10
Linkability fra search result til materialevisning og til search result

spaceo:
#### Link to issue

https://reload.atlassian.net/browse/DDFSOEG-186

#### Description

Adds proper url handling at search result

- Adding useUrl concept that avoids prop drilling of urls
- Adding helper for constructing path with query parameters
- Organising global helpers in separate files since they are growing in numbers

#### Checklist

### DIFF
--- a/src/apps/material/material.entry.tsx
+++ b/src/apps/material/material.entry.tsx
@@ -1,7 +1,7 @@
 import * as React from "react";
 import { withText } from "../../core/utils/text";
 import { Pid } from "../../core/utils/types/ids";
-import { withUrl } from "../../core/utils/url";
+import { withUrls } from "../../core/utils/url";
 import Material from "./material";
 
 interface MaterialEntryTextProps {
@@ -50,4 +50,4 @@ const MaterialEntry: React.FC<MaterialEntryProps> = ({ pid }) => {
   return <Material pid={pid} />;
 };
 
-export default withUrl(withText(MaterialEntry));
+export default withUrls(withText(MaterialEntry));
--- a/src/apps/material/material.entry.tsx
+++ b/src/apps/material/material.entry.tsx
@@ -1,6 +1,7 @@
 import * as React from "react";
 import { withText } from "../../core/utils/text";
 import { Pid } from "../../core/utils/types/ids";
+import { withUrl } from "../../core/utils/url";
 import Material from "./material";
 
 interface MaterialEntryTextProps {
@@ -34,14 +35,19 @@ interface MaterialEntryTextProps {
   audienceText: string;
   fictionNonfictionText: string;
 }
-
-export interface MaterialEntryProps extends MaterialEntryTextProps {
-  pid: Pid;
+interface MaterialEntryUrlProps {
   searchUrl: string;
+  materialUrl: string;
 }
 
-const MaterialEntry: React.FC<MaterialEntryProps> = ({ pid, searchUrl }) => {
-  return <Material pid={pid} searchUrl={searchUrl} />;
+export interface MaterialEntryProps
+  extends MaterialEntryUrlProps,
+    MaterialEntryTextProps {
+  pid: Pid;
+}
+
+const MaterialEntry: React.FC<MaterialEntryProps> = ({ pid }) => {
+  return <Material pid={pid} />;
 };
 
-export default withText(MaterialEntry);
+export default withUrl(withText(MaterialEntry));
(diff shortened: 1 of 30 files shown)


src/apps/material/material.entry.tsx
```diff
@@ -29,14 +30,19 @@ interface MaterialEntryTextProps {
   audienceText: string;
   fictionNonfictionText: string;
 }
+interface MaterialEntryUrlProps {
+  searchUrl: string;
+  materialUrl: string;
```
kasperg, line 35: Consider using [the `URL` type](https://developer.mozilla.org/en-US/docs/Web/API/URL) instead of raw strings.

It seems as if this may help with some type safety and improved value handling. I do not know if it works with Storybook though.

src/components/search-result-list/search-result-list-item/search-result-list-item.tsx
```diff
@@ -34,17 +35,21 @@ const SearchResultListItem: React.FC<SearchResultListItemProps> = ({
...
   const creatorsText = creatorsToString(
     flattenCreators(filterCreators(creators, ["Person"])),
     t
   );
   const author = creatorsText || "[Creators are missing]";
   const datePublished = getFirstPublishedYear(manifestations);
   const manifestationPid = getManifestationPid(manifestations);
+  const firstInSeries = series?.[0];
+  const { title: seriesTitle, numberInSeries } = firstInSeries;
+  const materialFullPath = `${materialUrl}/${workId}`;
```
kasperg, line 48: Consider adding support for placeholders in urls.

We have previously worked with placeholders for values in props. If the material url from `useUrls()` is `/work/:workId` then we replace `:workId` with the actual value. This reduces the coupling between routing in the original application

src/core/utils/url.tsx
```diff
@@ -0,0 +1,49 @@
+import React, { useEffect, useState } from "react";
+import { RootState, store, useSelector } from "../store";
+import { addUrlEntries } from "../url.slice";
+
+export const useUrls = () => {
+  const { data } = useSelector((state: RootState) => state.url);
+  return data;
+};
+
+export const withUrl = <T,>(Component: React.ComponentType<T>) => {
```
kasperg, line 10: Consider renaming this to `withUrls` as there are multiple and our hook is called `useUrls`.

src/core/utils/url.tsx
```diff
@@ -0,0 +1,49 @@
...
+          return String(prop).match(pattern);
+        })
+      );
+      // and match all props that do NOT end with "Text".
+      const nonUrlEntries = Object.fromEntries(
+        Object.entries(props).filter(([prop]) => {
+          return !String(prop).match(pattern);
+        })
+      );
+      // If we do have props that are not text props
```
kasperg, line 28: There seems to be a wee bit of copy/pasting here.

src/core/utils/helpers/general.ts
```diff
@@ -1,12 +1,13 @@
...
-} from "../dbc-gateway/generated/graphql";
-import { FaustId, Pid } from "./types/ids";
+  ManifestationsSimpleFragment,
+  WorkSmallFragment
+} from "../../dbc-gateway/generated/graphql";
+import { UseTextFunction } from "../text";
+import { FaustId, Pid } from "../types/ids";
+import { getUrlQueryParam } from "./url";
 
 export const orderManifestationsByYear = (
```
kasperg, line 12: I appreciate the introduction of `date` and `url` helpers. Have you considered adding a `work` or `material` helper?

kasperg (changes requested):
:+1: Good stuff!

I have some questions for your consideration.

I general I think it worth considering Using `URL` instead of `string` as type for urls.

Please remember to follow up on CI results. There is `console.log` in the code.

src/core/utils/helpers/general.ts
spaceo: It depends if there is a direct link between those entities and the helper functions.
The core helpers are meant to be used on all areas of the application and that is why they are called generic names like url and date.
More specific helpers are also being used next to the various components.

src/core/utils/url.tsx
spaceo: you´re right

src/core/utils/url.tsx
spaceo: donno

src/components/search-result-list/search-result-list-item/search-result-list-item.tsx
spaceo: yep

kasperg (approved):
:+1: LGTM

PR #82, opened 2022-08-11
Tests for search header app

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFSOEG-106

#### Description
This PR introduces tests for search header app to check rendering and functionality. Tests make sure that the app:
- Allows user to write into the search field
- Doesn't show suggestions before 3 characters are typed
- Allows use of arrow keys to navigate autosuggest
- Matches text in the search field with highlighted item
- Shows both parts of the autosuggest
- Shows cover pictures for the material suggestions

All of the data is mocked in order for the tests to not have to wait for
real data to return and so that they can run in the Github environment.

#### Screenshot of the result
image

#### Checklist

- [x] My code complies with [our coding guidelines](../docs/code_guidelines.md).

N/A


src/apps/search-header/search-header.test.ts
```diff
@@ -0,0 +1,213 @@
+describe("Search header app", () => {
+  beforeEach(() => {
```
spaceo, line 2: There is a lot of mock data here. You could consider move the data to a fixture like here: https://github.com/danskernesdigitalebibliotek/dpl-react/pull/83/files#diff-2681eddd9ccddb0757dced492f1e67c5065af0d811d13c6d071a7b9d09124753R86

src/apps/search-header/search-header.test.ts
```diff
@@ -0,0 +1,213 @@
...
+    });
+    cy.intercept("GET", "**/covers**", {
+      statusCode: 200,
+      body: [
+        {
+          id: "870970-basis:25245784",
+          type: "pid",
+          imageUrls: {
+            small: {
+              url: "https://res.cloudinary.com/dandigbib/image/upload/t_ddb_cover_small/v1543886150/bogportalen.dk/9788702029444.jpg",
```
spaceo, line 152: Instead of serving an external image you can also mock it like here:
https://github.com/danskernesdigitalebibliotek/dpl-react/pull/83/files#diff-2681eddd9ccddb0757dced492f1e67c5065af0d811d13c6d071a7b9d09124753R92

src/core/dbc-gateway/graphql-fetcher.ts
```diff
@@ -9,9 +9,12 @@ export const fetcher = <TData, TVariables>(
     // First version is with a library token.
     const libraryToken = getToken(TOKEN_LIBRARY_KEY);
 
-    if (!libraryToken) {
-      throw new Error("Library token is missing!");
-    }
```
spaceo, line 14: I think we should not remove the missing token error handling.
You could argue if an error should be thrown or logged or whatever. But I think we need error handling because the FBI services will not work without a token.
I have solved the same problem (about Cypress failing because the token is not present) by setting the token in sessionStorage which we use in storybook to set tokens that need to be accessable:
https://github.com/danskernesdigitalebibliotek/dpl-react/pull/83/files#diff-2681eddd9ccddb0757dced492f1e67c5065af0d811d13c6d071a7b9d09124753R82-R84

src/apps/search-header/search-header.test.ts
```diff
@@ -0,0 +1,213 @@
...
+    // eslint-disable-next-line cypress/no-unnecessary-waiting
+    cy.wait(1000);
+    cy.get(".header__menu-search-input").focus().type("{downArrow}");
+    cy.get("#downshift-0-item-0").should("have.attr", "aria-selected", "true");
+  });
+
+  it("Matches text in the search field with highlighted item", () => {
+    cy.get(".header__menu-search-input").focus().type("har");
+    // eslint-disable-next-line cypress/no-unnecessary-waiting
+    cy.wait(1000);
```
spaceo, line 188: I am still in doubt wether it is not possible to solve this with .then instead. Have you investigated that?
https://docs.cypress.io/api/commands/then#Syntax

spaceo (changes requested):
Looks great.
I have a few comments. I think we need to find a solution for the library token that we can agree upon.
Take a look at how I solved it.

src/core/dbc-gateway/graphql-fetcher.ts
spaceo: Disrtegard this. We keep it like this ;)

Adamik10: talked about this live :)

src/apps/search-header/search-header.test.ts
Adamik10: Very good point! I am mocking the covers now too :)

src/apps/search-header/search-header.test.ts
Adamik10: Diddly done :)

src/apps/search-header/search-header.test.ts
Adamik10: Kasper found a nice way to make it work and not use wait :)

src/apps/search-header/search-header.tsx
```diff
@@ -24,7 +24,12 @@ const SearchHeader: React.FC = () => {
...
   const [q, setQ] = useState<string>("");
   const [qWithoutQuery, setQWithoutQuery] = useState<string>(q);
-  const [suggestItems, setSuggestItems] = useState<any[]>([]);
+  const [suggestItems, setSuggestItems] = useState<
+    SuggestionsFromQueryStringQuery["suggest"]["result"] | []
+  >([]);
+  // we need to convert between string and suggestion result object so
+  // that the value in the search field on enter click doesn't become [object][object]
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
   const [currentlySelectedItem, setCurrentlySelectedItem] = useState<any>("");
```
spaceo, line 33: If you need to convert between string and result object I guess it should look something like:
```
const [currentlySelectedItem, setCurrentlySelectedItem] = useState<string | SuggestionsFromQueryStringQuery["suggest"]["result"] >("");
```
...or what?

spaceo: We should avoid usage of `any` and , of course, usage of `eslint-disable-next-line`

Adamik10: Let's look at this as a separate task :)

spaceo (approved):

sinejespersen (approved):

PR #83, opened 2022-08-12
E2e test test af search result app

spaceo:
#### Link to issue

https://reload.atlassian.net/browse/DDFSOEG-189

#### Description

Adding cypress tests for search result

- Introducing usage of local images in fixtures. Will also increase speed.
- Adding fixture data for a cover
- Adding fixture data for the search result test itself

#### Checklist


src/apps/search-result/search-result.test.ts
```diff
@@ -0,0 +1,127 @@
...
+
+    cy.log("Check length of search result list:");
+    cy.get(".search-result-page__list").find("li").should("have.length", 2);
+
+    cy.log("Do the search results have images?");
+    cy.get(".search-result-page__list .search-result-item img")
+      .should("have.attr", "src")
+      .and("match", coverUrlPattern);
+
+    cy.log("Does the search result have favourite buttons?");
```
kasperg, line 17: I like this way of attaching something humanly readable to an assertion. It reminds me a bit of PHPUnit 😄 . We could also do this with code comments but this puts it in the Cypress UI which may make it easier to work with in GitHub Actions in the like.

src/apps/search-result/search-result.test.ts
```diff
@@ -0,0 +1,127 @@
...
+    cy.get(
+      ".search-result-page__list .search-result-item .button-favourite"
+    ).should("have.attr", "aria-label", "Add to favorites");
+
+    cy.log("Does the search result have a series line?");
+    cy.get(
+      ".search-result-page__list .search-result-item .horizontal-term-line"
+    ).should(
+      "contain.text",
+      "Nr. 1  i serienDummy Some SeriesNr. 1  i serienDummy Some Series"
```
kasperg, line 27: Consider adding a comment explaining why this text is somewhat odd.

An alternative approach would be to modify the mocked response to return something that is closer to what we expected.

src/apps/search-result/search-result.test.ts
```diff
@@ -0,0 +1,127 @@
...
+    cy.log("Does the search result have titles?");
+    cy.get(".search-result-page__list .search-result-item h2").should(
+      "contain.text",
+      "Dummy Some Title: Full"
+    );
+
+    cy.log("Does the search result have authors?");
+    cy.get(
+      ".search-result-page__list .search-result-item p.text-small-caption"
+    ).should("contain.text", "Af Dummy Jens Jensen (Dummy 1839)");
```
kasperg, line 39: Consider reducing the dependency on CSS selectors.

Cypress and others recommend testing based on what is visible to end users or assistive technologies to reduce frailty. In this regard it is sufficient that the UI contains the text "Af Dummy Jens Jensen (Dummy 1839)".

This also applies for other tests in this PR.

kasperg (commented):
:+1: Good stuff.

src/apps/search-result/search-result.test.ts
spaceo: 👍

src/apps/search-result/search-result.test.ts
spaceo: Yes UI agree. I think we need to revise the text when we get some meaningful data from DBC. I'll write a comment about that.

src/apps/search-result/search-result.test.ts
spaceo: Check

cableman (approved):

PR #85, opened 2022-08-15
Material app e2e + it also fixes the pathname problem

kasperbirch1:
#### Link to issue

https://reload.atlassian.net/browse/DDFSOEG-188

#### Description

This PR sets up a cypress test that checks the material app and it also fixes the pathname problem.

- The test uses mockup data to check if data is rendered in the right places

**This commit fixes the pathname bug 9306dfbfab7a8bcb0ac40c718358aaffdd32d7c9** 

#### Screenshot of the result
Skærmbillede 2022-08-16 kl  09 55 31

#### Checklist

### DIFF
--- a/src/apps/material/material.test.ts
+++ b/src/apps/material/material.test.ts
@@ -1,0 +1,69 @@
+describe("Material", () => {
+  it("Should render the material page as expected", () => {
+    cy.log("Does the Material have title?");
+    cy.contains("Dummy Some Title: Full");
+
+    cy.log("Check that cover has a src");
+    cy.get("img").should(
+      "have.attr",
+      "src",
+      "https:res.cloudinary.com/dandigbib/image/upload/t_ddb_cover_large/v1605727140/bogportalen.dk/9781848485532.jpg"
+    );
+
+    cy.log("Does the material have favourite buttons?");
+    cy.get(".button-favourite").should(
+      "have.attr",
+      "aria-label",
+      "Add to favorites"
+    );
+
+    cy.log("Does the material have horizontal lines?");
+    cy.contains("Nr 1 i serien");
+    cy.contains("Dummy Some Series");
+
+    cy.log("Does the material have authors?");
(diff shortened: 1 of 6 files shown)


src/apps/material/material.test.ts
```diff
@@ -0,0 +1,78 @@
+describe("Material app e2e", () => {
+  it("Should render the material page as expected", () => {
+    cy.log("Does the Material have title?");
+    cy.get(".text-header-h1").should("have.length.gte", 1);
```
kasperg, line 4: In general please check that the expected data is visible/accessible somewhere on the page instead of checking for the presence of specific HTML tags or classes.

This is recommended as best practice by Cypress to avoid frail tests.

This applies throughout the test.

src/apps/material/material.test.ts
```diff
@@ -0,0 +1,78 @@
+describe("Material app e2e", () => {
```
kasperg, line 1:
```suggestion
describe("Material", () => {
```

There is no reason to add app or e2e to the test name. This is not something we do for other tests as well.

src/apps/material/material.test.ts
```diff
@@ -0,0 +1,78 @@
...
+        cy.intercept("GET", "**/availability/v3?recordid=**", result);
+      })
+      .as("Availability service");
+
+    // Intercept like button
+    cy.intercept(
+      "HEAD",
+      "https://prod.materiallist.dandigbib.org/list/default/work-of:870970-basis:52557240",
+      {
+        statusCode: 401
```
kasperg, line 72: Please reconsider this mock.

Is there a reason why we want to return 401 (Unauthorized) for this request? I would have expected a 404 (the material is not on the list) or a 200 (Material is on list).

Also there is no reason to use a very specific url. It will easily break if something else changes.

kasperg (approved):
:+1: LGTM

PR #87, opened 2022-08-15
Update cover handling in Storybook

kasperg:
#### Description

Update handling of pids for covers in Storybook

- Change materialId argument to pid. It was changed to `pid` for the component in 69daa5e5b77a88549cefdc19f8714d5ff8747343 so the `pid` control would appear anyway. Now it is just a matter of configuring the control
- Move the pid argument to the top
- Make the cover component a component instead of an atom

See individual commits for more commentary

#### Screenshot of the result

Before:

Atoms : Cover - Item ⋅ Storybook 2022-08-15 16-35-14

After:

Components : Cover - Item ⋅ Storybook 2022-08-15 16-33-54

#### Checklist

### DIFF
--- a/src/components/cover/cover.dev.tsx
+++ b/src/components/cover/cover.dev.tsx
@@ -4,7 +4,7 @@ import { Cover } from "./cover";
 import { getCurrentLocation } from "../../core/utils/helpers/url";
 
 export default {
-  title: "Atoms / Cover",
+  title: "Components / Cover",
   component: Cover,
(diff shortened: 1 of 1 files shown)


Adamik10 (approved):
Great initiative :) It gets a thumbs up from me 😉 
But I believe there is something going on with the integration test. Please look into that.

spaceo (approved):

sinejespersen (approved):

PR #88, opened 2022-08-15
WIP: Publizon integration

cableman:
#### Link to issue

#### Description

Integration with digital loans from publizon... 

#### Screenshot of the result

#### Checklist


sinejespersen (approved):

spaceo (approved):
LGTM 👍

src/core/fbs/mutator/fetcher.ts
```diff
@@ -52,7 +52,7 @@ export const fetcher = async <ResponseType>({
   data?: BodyType<unknown>;
   signal?: AbortSignal;
 }) => {
-  const libraryToken = getToken(TOKEN_LIBRARY_KEY);
+  const libraryToken = getToken(TOKEN_USER_KEY);
```
kasperg, line 55: This cannot be added in the current form. This change will mean that requests against FBS for e.g. availability information will fail because of missing tokens without a user being logged in.

One way to address this would be to add the user token if that is available and fall back to the library token which should always be available.

Also the variable name `libraryToken` is misleading.

src/core/publizon/mutator/fetcher.ts
```diff
@@ -0,0 +1,97 @@
...
+  data
+}: {
+  url: string;
+  method: "get" | "post" | "put" | "delete" | "patch" | "head";
+  headers?: object;
+  params?: unknown;
+  data?: BodyType<unknown>;
+  signal?: AbortSignal;
+}) => {
+  const libraryToken = getToken(TOKEN_USER_KEY);
```
kasperg, line 55: See comment regarding the FBS fetcher.

kasperg (changes requested):
Please look into the fetcher token changes.

src/core/fbs/mutator/fetcher.ts
cableman: Fixed

src/core/publizon/mutator/fetcher.ts
cableman: Fixed

sinejespersen (approved):

spaceo (approved):
The idea of consolidating the fetchers has been moved out of scope on our side for now.
With that in mind I think it is ok to repeat the code as it is.
It could also be an idea to create a file called /src/core/utils/fetcher.ts and create some functions for resolving the token etc. It does hurt a bit to violate the principle like that.
Otherwise it is approved.

PR #90, opened 2022-08-17
Use workid in material app

spaceo:
#### Description

Since we have decided to use full work id when using the material app we need to change the inner workings, query etc.

#### Checklist

As it is now we use the newest manifestation pid as the pid for choosing cover etc. That will change when we have implemented using the type parameter


kasperbirch1 (approved):
LGTM 👍

cableman (approved):

PR #91, opened 2022-08-17
Fix checks

kasperg:
#### Description

Remove imports and props that are no longer used. This addresses issues that are currently popping up on other unrelated PR.

#### Checklist


spaceo (approved):
Is fixed in one of Adams PR's but heck let's get it in :)
LGTM +1

PR #92, opened 2022-08-17
Linkability from material page to search results & material page

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFSOEG-187
image
-> refers to image below
image

#### Description
This PR adjust the Horizontal Term Line component (that's all those underscored lines of text on image above) to not create links itself, but instead receives them ready made. This "dumb-ifying" of the component lets us decide where to redirect before calling the component. ("I samme serie" items redirect to material page, as opposed to the rest.)
Minor updates to the graphQL schema adding workId as one of the keys for the array of "in the same series as". We need the WorkId to create links to material page.

#### Screenshot of the result
n/a

#### Checklist

n/a

### DIFF
--- a/src/components/horizontal-term-line/HorizontalTermLine.tsx
+++ b/src/components/horizontal-term-line/HorizontalTermLine.tsx
@@ -1,12 +1,13 @@
 import React from "react";
-import { constructSearchUrl } from "../../core/utils/helpers/url";
-import { useUrls } from "../../core/utils/url";
 import { Link } from "../atoms/link";
 
 export interface HorizontalTermLineProps {
   title: string;
   subTitle?: string;
-  linkList: string[];
+  linkList: {
+    url: URL;
+    term: string;
+  }[];
 }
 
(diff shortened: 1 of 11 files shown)


src/components/horizontal-term-line/HorizontalTermLine.tsx
```diff
@@ -25,13 +24,11 @@ const HorizontalTermLine: React.FC<HorizontalTermLineProps> = ({
         )}
       </p>
 
-      {linkList.map((term) => {
-        const termUrl = constructSearchUrl(searchUrl, term);
-
+      {linkList.map((item) => {
         return (
```
spaceo, line 28: You can also destruct the values:
```
const { term, url } = item; 
```
instead of using `item.xxx`

spaceo (approved):
LGTM +1
Approved with non-blocking comment.

sinejespersen (approved):

PR #93, opened 2022-08-17
Select work manifestation 

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFSOEG-190

#### Description
This PR adds select work manifestation logic.

- Material type is now specified in the URL and will be set to first published manifestation if not. 
- A currentManifestation state has been added so that we can choose based on it how the material page should be displayed, for example, cover and buttons.
- Availability labels can be clicked on to change both state and URL
- Availability labels show as selected if the type of material text is the same

There is also added a translation of Reviews

#### Screenshot of the result
**type=Dummy+bog**
Skærmbillede 2022-08-22 kl  13 47 42
**type=e-bog**  (not selected)
Skærmbillede 2022-08-22 kl  13 47 35

#### Checklist

### DIFF
--- a/src/apps/material/material.tsx
+++ b/src/apps/material/material.tsx
@@ -117,7 +117,7 @@ const Material: React.FC<MaterialProps> = ({ wid }) => {
         />
       </Disclosure>
       {data.work.reviews && data.work.reviews.length >= 1 && (
-        <Disclosure title="Anmeldelser" mainIconPath={CreateIcon}>
+        <Disclosure title={t("reviewsText")} mainIconPath={CreateIcon}>
           <MaterialReviews
             listOfReviews={
               data.work.reviews as Array<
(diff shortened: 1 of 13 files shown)


src/apps/material/material.tsx
```diff
@@ -36,38 +37,54 @@ export interface MaterialProps {
...
   const t = useText();
-  const [typeState, setTypeState] = useState<null | string>(null);
+
+  const [currentManifestationState, setCurrentManifestationState] = useState<
+    ManifestationsSimpleFragment["latest"] | null
+  >(null);
+
   const { data, isLoading } = useGetMaterialQuery({
     wid
   });
```
spaceo, line 67: I think we should have the loading and error handling up here before the useEffect.
So when we are past all that:
// TODO: handle erro
We can destruct the work object.

spaceo (commented):
Some comments...

src/apps/material/material.tsx
kasperbirch1: I get this eslint error if I move it above useEfffect:

React Hook "useEffect" is called conditionally. React Hooks must be called in the exact same order in every component render

PR #97, opened 2022-08-22
Select work manifestation + Material cover change

kasperbirch1:
#### Link to issues
https://reload.atlassian.net/browse/DDFSOEG-154
https://reload.atlassian.net/browse/DDFSOEG-190

#### Description
This PR adds select work manifestation logic and display the cover image for the selected manifestation

- Material type is now specified in the URL and will be set to first published manifestation if not. 
- A currentManifestation state has been added so that we can choose based on it how the material page should be displayed, for example, cover and buttons.
- Availability labels can be clicked on to change both state and URL
- Availability labels show as selected if the type of material text is the same

#### Screenshot of the result
Skærmbillede 2022-08-22 kl  14 30 34
Skærmbillede 2022-08-22 kl  14 30 27

#### Checklist

**There is also added a translation of Reviews**

### DIFF
--- a/src/apps/material/material.tsx
+++ b/src/apps/material/material.tsx
@@ -117,7 +117,7 @@ const Material: React.FC<MaterialProps> = ({ wid }) => {
         />
       </Disclosure>
       {data.work.reviews && data.work.reviews.length >= 1 && (
-        <Disclosure title="Anmeldelser" mainIconPath={CreateIcon}>
+        <Disclosure title={t("reviewsText")} mainIconPath={CreateIcon}>
           <MaterialReviews
             listOfReviews={
               data.work.reviews as Array<
--- a/src/apps/material/material.tsx
+++ b/src/apps/material/material.tsx
@@ -10,7 +10,7 @@ import {
   useGetMaterialQuery,
   ManifestationsSimpleFieldsFragment
 } from "../../core/dbc-gateway/generated/graphql";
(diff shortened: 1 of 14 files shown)


src/apps/material/material.tsx
```diff
@@ -117,7 +159,7 @@ const Material: React.FC<MaterialProps> = ({ wid }) => {
         />
       </Disclosure>
       {data.work.reviews && data.work.reviews.length >= 1 && (
-        <Disclosure title="Anmeldelser" mainIconPath={CreateIcon}>
+        <Disclosure title={t("reviewsText")} mainIconPath={CreateIcon}>
```
cableman, line 162: 👍

src/components/material/MaterialHeader.tsx
```diff
@@ -37,10 +46,6 @@ const MaterialHeader: React.FC<MaterialHeaderProps> = ({
     t
   );
 
-  // TODO: Temporary way to get a pid we can use for showing a cover for the material.
-  // It should be replaced with some dynamic feature
-  // that follows the current type of the material.
-  const pid = getManifestationPid(manifestations) as Pid;
   const author = creatorsText || "[Creators are missing]";
```
cableman, line 49: Should "[Creators are missing]" be translatable?

cableman (changes requested):

PR #101, opened 2022-08-23
indloggetudlogget funktionalitet DDFSOEG 178

spaceo:
This PR is dealing with the scenario where a user wants to perform an action that is prohibited for anonymous users.
As it is now It has been implemented on the favourite button on search results and on the material page.

The flow that is covers:
* An anonymous user clicks the favourite button
* The request that should be fired for storing the material on the users favourite list is persisted via redux-persist
* The user is being redirected to dpl-cms via the authUrl that has been provided to the app. The authUrl gets a `current-path` url parameter attached so the user is redirected back to place where the favourite button was clicked.
* User logs in.
* User is being rediredted back to where the user came from
* The persisted request is fired and the app is being blocked until the request has been fired.
* The application runs after the request has been fires and the user can see the material has been added to the list by the heart that should be filled

The functionality consists of two major areas:
1. A redux (persisted) store that holds a request that could not be called before the user has been authenticated
2. A GuardedApp component that holds back the app if a user has been authenticated and has a persisted request. The app is withheld until the request has been executed.

 #### Link to issue

https://reload.atlassian.net/browse/DDFSOEG-178

#### Screenshot of the result

#### Checklist

Another PR is soon to be released on dpl-cms that handles the auth handling which this PR is depending upon:
https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/122

Codecov is complaining. Please ignore it for now. It is not following the coverage we have agreed on in this project.

### DIFF
--- a/cypress/support/index.ts
+++ b/cypress/support/index.ts
@@ -6,6 +6,10 @@ import "@cypress/code-coverage/support";
 const TOKEN_USER_KEY = "user";
 
 Cypress.Commands.add("createFakeAuthenticatedSession", () => {
+  // Since the user token is shared in storybook by setting it in sessionStorage
+  // we can use that and fake that we have a inlogged user session
+  // by using the same principle.
+  // See userToken handling in .storybbok/preview.js.
   window.sessionStorage.setItem(TOKEN_USER_KEY, "999");
 });
 
@@ -16,7 +20,7 @@ declare global {
        * Pretend that a user is logged in.
        * @example cy.createFakeAuthenticatedSession()
        */
-      createFakeAuthenticatedSession(): Chainable<Element>;
+      createFakeAuthenticatedSession(): void;
     }
   }
 }
--- a/cypress/support/index.ts
+++ b/cypress/support/index.ts
@@ -1,0 +1,22 @@
+// According to the documentation of types and Cypress commands
+// the namespace is declared like it is done here. Therefore we'll bypass errors about it.
+/* eslint-disable @typescript-eslint/no-namespace */
+import "@cypress/code-coverage/support";
+
+const TOKEN_USER_KEY = "user";
+
+Cypress.Commands.add("createFakeAuthenticatedSession", () => {
+  window.sessionStorage.setItem(TOKEN_USER_KEY, "999");
+});
+
+declare global {
+  namespace Cypress {
+    interface Chainable {
+      /**
+       * Pretend that a user is logged in.
+       * @example cy.createFakeAuthenticatedSession()
+       */
+      createFakeAuthenticatedSession(): Chainable<Element>;
+    }
+  }
+}
--- a/src/apps/search-header/search-header.entry.tsx
+++ b/src/apps/search-header/search-header.entry.tsx
@@ -1,5 +1,4 @@
 import * as React from "react";
-import GuardedApp from "../../components/guarded-app";
 import { withText } from "../../core/utils/text";
 import { withUrls } from "../../core/utils/url";
 import SearchHeader from "./search-header";
@@ -22,9 +21,8 @@ export interface SearchHeaderEntryProps
   extends SearchHeaderTextProps,
     SearchHeaderUrlProps {}
 
-const WrappedSearchHeader: React.FC<SearchHeaderEntryProps> = () => (
-  <GuardedApp app="search-header">
-    <SearchHeader />
-  </GuardedApp>
-);
-export default withUrls(withText(WrappedSearchHeader));
+const SearchHeaderEntry: React.FC<SearchHeaderEntryProps> = () => {
+  return <SearchHeader />;
+};
+
+export default withUrls(withText(SearchHeaderEntry));
--- a/src/apps/search-header/search-header.entry.tsx
+++ b/src/apps/search-header/search-header.entry.tsx
@@ -1,4 +1,5 @@
 import * as React from "react";
+import GuardedApp from "../../components/guarded-app";
 import { withText } from "../../core/utils/text";
 import { withUrls } from "../../core/utils/url";
 import SearchHeader from "./search-header";
@@ -13,23 +14,17 @@ export interface SearchHeaderTextProps {
 }
 
 export interface SearchHeaderUrlProps {
-  searchUrl: string;
-  materialUrl: string;
+  searchUrl?: string;
+  materialUrl?: string;
 }
 
 export interface SearchHeaderEntryProps
   extends SearchHeaderTextProps,
     SearchHeaderUrlProps {}
 
-const SearchHeaderEntry: React.FC<SearchHeaderEntryProps> = ({
-  altText = "search icon",
-  inputPlaceholderText = "Search here",
-  stringSuggestionAuthorText = "author",
-  stringSuggestionWorkText = "work",
-  stringSuggestionTopicText = "topic",
-  etAlText = "et al."
-}) => {
-  return <SearchHeader />;
-};
-
-export default withUrls(withText(SearchHeaderEntry));
+const WrappedSearchHeader: React.FC<SearchHeaderEntryProps> = () => (
+  <GuardedApp app="search-header">
+    <SearchHeader />
+  </GuardedApp>
+);
+export default withUrls(withText(WrappedSearchHeader));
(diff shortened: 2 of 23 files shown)


src/components/button-favourite/button-favourite.dev.tsx
```diff
@@ -18,7 +21,14 @@ export default {
 
 const Template: ComponentStory<typeof ButtonFavourite> = (
   args: ButtonFavouriteProps
-) => <ButtonFavourite {...args} />;
+) => {
+  // This is a fake situation where we just need to give the button a handler.
+  // The handler does nothing.
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  const addToListRequest = (id: ButtonFavouriteId) => {};
```
kasperg, line 28: :+1:

src/components/button-favourite/button-favourite.tsx
```diff
@@ -1,23 +1,27 @@
 import React, { useCallback, useEffect, useState } from "react";
+import { useDispatch } from "react-redux";
```
kasperg, line 2: Please remove unused dependencies. This also applies elsewhere in this PR.

src/components/guarded-app.tsx
```diff
@@ -0,0 +1,91 @@
...
+// until the persisted request has been executed.
+const GuardedApp = ({ app, children }: GuardedAppProps) => {
+  const dispatch = useDispatch();
+  const { request: persistedRequest } = useSelector(
+    (state: RootState) => state.guardedRequests
+  );
+  const isApplicationBlocked = persistedRequest && !userIsAnonymous();
+  const didAuthenticate = getUrlQueryParam(AUTH_PARAM);
+
+  console.log("PERSISTED REQUEST:", persistedRequest);
```
kasperg, line 36: Please silence these errors with an explanation regarding why we think it is necessary to use the `console`.

It would probably make sense to create a function for this as it occurs multiple times in this component.

Also consider using [`console.debug()`](https://developer.mozilla.org/en-US/docs/Web/API/console/debug) to indicate that is in fact something that we intend to keep around for debugging purposes.

src/components/guarded-app.tsx
```diff
@@ -0,0 +1,91 @@
...
+    );
+    // Remove auth parameter from url so we don't accidentally
+    // repeat the functionality related to it.
+    replaceCurrentLocation(currentUrlWithoutAuthParam);
+    dispatch(removeRequest());
+  }, [app, didAuthenticate, dispatch, isApplicationBlocked, persistedRequest]);
+
+  if (isApplicationBlocked) {
+    return <div />;
+  }
```
kasperg, line 83:
```suggestion
  if (isApplicationBlocked) {
    return;
  }
```

Consider if could we just not return anything?

src/core/guardedRequests.slice.ts
```diff
@@ -0,0 +1,143 @@
...
+        if (authUrl) {
+          const { pathname, search } = appendQueryParametersToUrl(
+            new URL(getCurrentLocation()),
+            {
+              [AUTH_PARAM]: "1"
+            }
+          );
+          const redirectUrl = appendQueryParametersToUrl(authUrl, {
+            "current-path": `${pathname}${search}`
+          });
```
kasperg, line 113: Consider if this can be moved to the mounting application in the `authUrl` prop.

It seems as if this will create a hard dependency between whatever is located at `authUrl` (the authenticating app) and the guarded app. The authenticating app has to store and redirect to whatever is provided in the current path query parameter.

How about instead stating that it is the responsibility of the mounting app to ensure that whatever url is provided as `authUrl` will return the user to an appropriate state. In our case Drupal can provide `/oauth/adgangsplatformen/redirect?destination=/search?q=harry` as auth url and this will ensure that the browser ends up at `/search?q=harry` after authentication. When the guarded app can pick up where it left off.

src/core/utils/helpers/url.ts
```diff
@@ -102,3 +102,12 @@ export const turnUrlStringsIntoObjects = (urls: { [key: string]: string }) => {
...
 };
+
+export const replaceCurrentLocation = (replacementUrl: URL) => {
+  window.history.replaceState(null, "", replacementUrl);
+};
+
+export const removeQueryParametersFromUrl = (url: URL, parameter: string) => {
+  url.searchParams.delete(parameter);
+  return url;
+};
```
kasperg, line 113: Consider if this function makes sense.

It contains a single line of code without much complexity, additional values etc. To me it seems as if developers might as well call `searchParams.delete` directly.

src/core/utils/types/ids.ts
```diff
@@ -1,3 +1,4 @@
 export type FaustId = `${number}`;
 export type Pid = `${number}-${string}:${FaustId}`;
 export type WorkId = `work-of:${number}-${string}:${FaustId}`;
+export type App = "material" | "search-header" | "search-result";
```
kasperg, line 4: Consider if loan list should be included in this list. You could also rename it to only contain ids for guarded apps.

kasperg (commented):
:+1: lots of code, lots of comments.

cypress/support/index.ts
```diff
@@ -0,0 +1,22 @@
+// According to the documentation of types and Cypress commands
+// the namespace is declared like it is done here. Therefore we'll bypass errors about it.
+/* eslint-disable @typescript-eslint/no-namespace */
+import "@cypress/code-coverage/support";
+
+const TOKEN_USER_KEY = "user";
+
+Cypress.Commands.add("createFakeAuthenticatedSession", () => {
+  window.sessionStorage.setItem(TOKEN_USER_KEY, "999");
```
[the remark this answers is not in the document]
spaceo: 👍

src/apps/search-header/search-header.entry.tsx
```diff
@@ -13,23 +14,17 @@ export interface SearchHeaderTextProps {
...
-  return <SearchHeader />;
-};
-
-export default withUrls(withText(SearchHeaderEntry));
+const WrappedSearchHeader: React.FC<SearchHeaderEntryProps> = () => (
+  <GuardedApp app="search-header">
+    <SearchHeader />
+  </GuardedApp>
+);
+export default withUrls(withText(WrappedSearchHeader));
```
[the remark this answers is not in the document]
spaceo: Agree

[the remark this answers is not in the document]
spaceo: ..in that the guard wrap should be removed

src/components/button-favourite/button-favourite.tsx
spaceo: 👍

src/components/guarded-app.tsx
spaceo: Its because the guarded app needs to return a jsx element. That is also what the stories in storybook expect.

src/core/guardedRequests.slice.ts
spaceo: It's a part of the openid_connect auth flow to delete the destination, that's why I added the current-path functionality.
I don't consider the current-path principle to be a hard dependency between the app and Drupal. Current path is just an extra feature the Drupal endpoint exposes that makes it possible to redirect to it after ended authentication. Actually it is a replacement of the Drupal destination feature since it was removed by the openid_connect module.

src/core/utils/helpers/url.ts
spaceo: Yeah I had the same thought. But since we have abstracted all URL handling away in other cases I ended up thinking we should do that here as well.

src/core/utils/types/ids.ts
spaceo: I have removed the search header since it has no guarded requests.
This should have a small effect on performance as well. If it is measurable I don't know.

cypress/support/index.ts
```diff
@@ -0,0 +1,22 @@
...
+});
+
+declare global {
+  namespace Cypress {
+    interface Chainable {
+      /**
+       * Pretend that a user is logged in.
+       * @example cy.createFakeAuthenticatedSession()
+       */
+      createFakeAuthenticatedSession(): Chainable<Element>;
```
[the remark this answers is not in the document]
spaceo: Yes. Using `void`now.

sinejespersen (approved):

PR #102, opened 2022-08-23
Info text under buttons in the material header.

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFSOEG-149

#### Description
# This PR adds logic for the info text below buttons on the material header.

### MaterialAvailabilityText is a wrapper component
That checks whether the manifestation accessTypes is PHYSICAL or ONLINE where after it shows **MaterialAvailabilityTextPhysical**, **MaterialAvailabilityTextOnline** or both

### MaterialAvailabilityTextPhysical
Takes a PID id which it then turns into a FaustId and uses useGetHoldingsV3 to get total materials and total reservations

### MaterialAvailabilityTextOnline
Takes an ISBN number and uses useGetV1ProductsIdentifier to  check if manifestation is costFree (blue title)

### Still missing.
useGetV1UserLoansIdentifier must be used to find user's totalEbookLoans and ebookLoansRemaining

#### Screenshot of the result
Skærmbillede 2022-08-29 kl  13 53 51
Skærmbillede 2022-08-29 kl  13 53 30
Skærmbillede 2022-08-29 kl  13 52 34

#### Checklist

### DIFF
--- a/.nycrc
+++ b/.nycrc
@@ -5,6 +5,7 @@
     "src/core/cover-service-api/**/*.*",
     "src/core/dbc-gateway/**/*.*",
     "src/core/fbs/**/*.*",
-    "src/core/material-list-api/**/*.*"
+    "src/core/material-list-api/**/*.*",
+    "src/core/publizoni/**/*.*"
   ]
 }
--- a/.nycrc
+++ b/.nycrc
@@ -6,6 +6,6 @@
     "src/core/dbc-gateway/**/*.*",
     "src/core/fbs/**/*.*",
     "src/core/material-list-api/**/*.*",
-    "src/core/publizoni/**/*.*"
+    "src/core/publizon/**/*.*"
   ]
 }
--- a/src/components/material/MaterialAvailabilityText/MaterialAvailabilityText.tsx
+++ b/src/components/material/MaterialAvailabilityText/MaterialAvailabilityText.tsx
@@ -19,7 +19,7 @@ const MaterialAvailabilityText: React.FC<Props> = ({ manifestation }) => {
         if (item.code === "ONLINE" && manifestation.identifiers)
           return (
             <MaterialAvailabilityTextOnline
-              isbn={manifestation.identifiers?.[0].value}
+              isbn={manifestation.identifiers[0].value}
             />
           );
         return null;
--- a/src/components/material/MaterialAvailabilityText/MaterialAvailabilityText.tsx
+++ b/src/components/material/MaterialAvailabilityText/MaterialAvailabilityText.tsx
@@ -1,0 +1,26 @@
+import * as React from "react";
+import { ManifestationsSimpleFieldsFragment } from "../../../core/dbc-gateway/generated/graphql";
+import { Pid } from "../../../core/utils/types/ids";
+import MaterialAvailabilityTextOnline from "./online/MaterialAvailabilityTextOnline";
+import MaterialAvailabilityTextPhysical from "./physical/MaterialAvailabilityTextPhysical";
+
+interface Props {
+  manifestation: ManifestationsSimpleFieldsFragment;
+}
+
+const MaterialAvailabilityText: React.FC<Props> = ({ manifestation }) => {
+  return (
+    <>
+      {manifestation.accessTypes.map((item) => {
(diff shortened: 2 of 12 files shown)


.nycrc
```diff
@@ -5,6 +5,7 @@
     "src/core/cover-service-api/**/*.*",
     "src/core/dbc-gateway/**/*.*",
     "src/core/fbs/**/*.*",
-    "src/core/material-list-api/**/*.*"
+    "src/core/material-list-api/**/*.*",
+    "src/core/publizoni/**/*.*"
```
kasperg, line 9:
```suggestion
    "src/core/publizon/**/*.*"
```

There seems to be a typo here? 😄

src/components/material/MaterialAvailabilityText/MaterialAvailabilityText.tsx
```diff
@@ -0,0 +1,31 @@
...
+    <>
+      {manifestation.accessTypes.map((item) => {
+        if (item.code === "PHYSICAL")
+          return (
+            <MaterialAvailabilityTextPhysical pid={manifestation.pid as Pid} />
+          );
+        if (item.code === "ONLINE" && manifestation.identifiers)
+          return (
+            <MaterialAvailabilityTextOnline
+              isbn={manifestation.identifiers?.[0].value}
```
kasperg, line 22:
```suggestion
              isbn={manifestation.identifiers[0].value}
```

Please remove the optional chaining operator. It should not be necessary given the check for `manifestation.identifiers` on line 19. It suggests an uncertainty about the data which should not be there.

src/components/material/MaterialAvailabilityText/online/MaterialAvailabilityTextOnline.tsx
```diff
@@ -0,0 +1,58 @@
...
+interface MaterialAvailabilityTextOnlineProps {
+  isbn: string;
+}
+
+const MaterialAvailabilityTextOnline: React.FC<
+  MaterialAvailabilityTextOnlineProps
+> = ({ isbn }) => {
+  const t = useText();
+  // const costfreeID = "9788711321683";
+  // const notCostfreeID = "9788740047905";
```
kasperg, line 18: Please add a `TODO` comment explaining why is nice to have this code around.

Normally I would argue against having code in comments but in this situation it is necessary since we do not get proper ids in the data.

src/components/material/MaterialAvailabilityText/online/MaterialAvailabilityTextOnline.tsx
```diff
@@ -0,0 +1,58 @@
...
+    return (
+      <MaterialAvailabilityTextParagraph>
+        {`${t("youHaveBorrowedText")} X ${t(
+          "outOfText"
+        )} Y [materialte-type] ${t("thisMonthText")}`}
+      </MaterialAvailabilityTextParagraph>
+    );
+  }
+
+  return null;
```
kasperg, line 55: Consider restructuring this code.

Your check for whether the product is cost free or not should focus on building the proper string. Then the string could finally be wrapped in a `MaterialAvailabilityTextParagraph` component.

With the current structure I cannot see how this function could ever end up with the final `return null`. Either `costFree` is truthy and then we end up with the branch on L38. If it is not then we end up on L46.

src/components/material/MaterialHeader.tsx
```diff
@@ -1,4 +1,4 @@
-import React from "react";
+import * as React from "react";
```
kasperg, line 1: I wonder why this change is necessary.

kasperg (changes requested):
:+1: There are a couple of things that I would like you to look into here.

kasperg (approved):
:+1: LGTM

kasperg (approved):
:+1: LGTM

src/core/publizon/mutator/fetcher.ts
```diff
@@ -1,6 +1,11 @@
 import { getToken, TOKEN_USER_KEY, TOKEN_LIBRARY_KEY } from "../../token";
```
sinejespersen, line 2: is this supposed to be here?

PR #111, opened 2022-08-30
Add MaterialPeriodical

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFSOEG-182

#### Description

# This PR is introducing MaterialPeriodical

### MaterialPeriodical
Is a wrapper component that is only rendered if a manifestation has a source that matches types that needs a PeriodicalSelect. it also fetches data with useGetHoldingsV3 and manipulates it before it renders MaterialPeriodicalSelect

### MaterialPeriodikumSelect 
Has a setState function (selectPeriodicalSelect)  that comes from the root in the Material app. This will be used later...

** groupBy  global function needs to be rewritten to typeScript **

#### Screenshot of the result
Skærmbillede 2022-08-31 kl  14 39 16
Skærmbillede 2022-08-31 kl  14 39 06

#### Checklist

### DIFF
--- a/src/apps/material/material.tsx
+++ b/src/apps/material/material.tsx
@@ -100,7 +100,7 @@ const Material: React.FC<MaterialProps> = ({ wid }) => {
           currentManifestation as ManifestationsSimpleFieldsFragment
         }
         selectManifestationHandler={setCurrentManifestation}
-        selectPeriodikumSelect={setPriodikumSelect}
+        selectPeriodicalSelect={setPriodikumSelect}
       />
       <MaterialDescription pid={pid} work={data.work} />
       <Disclosure
--- a/src/apps/material/material.tsx
+++ b/src/apps/material/material.tsx
@@ -39,6 +39,10 @@ const Material: React.FC<MaterialProps> = ({ wid }) => {
   const [currentManifestation, setCurrentManifestation] =
     useState<ManifestationsSimpleFieldsFragment | null>(null);
 
+  // periodicalSelect must be used later to change the UI and reservation when you have chosen a specific periodical
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  const [periodicalSelect, setPriodikumSelect] = useState<string | null>(null);
+
   const { data, isLoading } = useGetMaterialQuery({
     wid
   });
@@ -96,6 +100,7 @@ const Material: React.FC<MaterialProps> = ({ wid }) => {
           currentManifestation as ManifestationsSimpleFieldsFragment
         }
         selectManifestationHandler={setCurrentManifestation}
+        selectPeriodikumSelect={setPriodikumSelect}
       />
       <MaterialDescription pid={pid} work={data.work} />
       <Disclosure
--- a/src/apps/material/material.tsx
+++ b/src/apps/material/material.tsx
@@ -41,7 +41,7 @@ const Material: React.FC<MaterialProps> = ({ wid }) => {
 
   // periodicalSelect must be used later to change the UI and reservation when you have chosen a specific periodical
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
-  const [periodicalSelect, setPriodikumSelect] = useState<string | null>(null);
+  const [periodicalSelect, setPeriodicalSelect] = useState<string | null>(null);
 
   const { data, isLoading } = useGetMaterialQuery({
     wid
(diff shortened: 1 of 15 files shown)


src/apps/material/material.tsx
```diff
@@ -45,6 +45,8 @@ const Material: React.FC<MaterialProps> = ({ wid }) => {
   const [currentManifestation, setCurrentManifestation] =
     useState<ManifestationsSimpleFieldsFragment | null>(null);
 
+  const [periodikumSelect, setPriodikumSelect] = useState<string | null>(null);
```
spaceo, line 48: Lijs lint says you never use `periodikumSelect` later in the code 🤔

src/core/utils/helpers/general.ts
```diff
@@ -149,3 +149,37 @@ export const getRenewableMaterials = (list: LoanV2[]) => {
...
+//         return acc;
+//       }
+//       if (propertyValue in acc) {
+//         return { ...acc, [propertyValue]: [...acc[propertyValue], curr] };
+//       }
+//       return { ...acc, [propertyValue]: [curr] };
+//     },
+//     {} as Record<string, T[]> | Record<string, unknown>
+//     // This empty object is cache
+//   );
```
spaceo, line 185:
```suggestion
export const groupObjectArrayByProperty = <
  P extends string,
  T extends Record<P, string> & Record<string, unknown>
>(
  arr: T[],
  property: P
) =>
  arr.reduce<Record<string, T[]>>(
    (result: { [key: string]: T[] }, current: T) => {
      const groupBy = current[property];
      // Merge into result if the property already exist.
      if (groupBy in result) {
        return {
          ...result,
          [groupBy]: [...result[groupBy], current]
        };
      }
      // Otherwise create new property.
      return { ...result, [groupBy]: [current] };
    },
    {}
  );
```

src/apps/material/material.tsx
```diff
@@ -45,6 +45,10 @@ const Material: React.FC<MaterialProps> = ({ wid }) => {
   const [currentManifestation, setCurrentManifestation] =
     useState<ManifestationsSimpleFieldsFragment | null>(null);
 
+  // periodikumSelect must be used later to change the UI and reservation when you have chosen a specific periodical
+  // eslint-disable-next-line
+  const [periodikumSelect, setPriodikumSelect] = useState<string | null>(null);
```
kasperg, line 50: Please update this code to use periodical instead of periodikum (including typos).

:+1: for adding the explanation for the line. [Please refer to the specific rule being disabled instead of disabling eslint entirely for the line](https://eslint.org/docs/latest/user-guide/configuring/rules#using-configuration-comments-1).

src/components/material/MaterialHeader.tsx
```diff
@@ -42,8 +45,12 @@ const MaterialHeader: React.FC<MaterialHeaderProps> = ({
     workId: wid
   },
   manifestation,
-  selectManifestationHandler
+  selectManifestationHandler,
+  selectPeriodikumSelect
 }) => {
+  // THIS MUST BE DELETED (START
+  const isDemoPeriodikum = getUrlQueryParam("periodikum");
+  // THIS MUST BE DELETED (END)
```
kasperg, line 53: Please prefix such comments with `TODO` or remove this feature entirely since we have already demoed this.

The convention of using `TODO` makes it easier for us to search through the codebase for issues that needs to be addressed in the future.

src/components/material/MaterialHeader.tsx
```diff
@@ -74,6 +81,10 @@ const MaterialHeader: React.FC<MaterialHeaderProps> = ({
   const coverPid =
     (manifestation?.pid as Pid) || getManifestationPid(manifestations);
 
+  const faustId = manifestation
+    ? convertPostIdToFaustId(manifestation?.pid as Pid)
```
kasperg, line 85: Please remove the null chaining if it is not necessary. 

As far as I can tell we know that `manifestation` will always be set in this codebranch.

src/components/material/MaterialPeriodical.tsx
```diff
@@ -0,0 +1,48 @@
...
+}) => {
+  const { data, isLoading, isError } = useGetHoldingsV3({
+    recordid: [String(faustId)]
+  });
+
+  if (isLoading || isError || !data) return null;
+
+  const holdings = data[0]?.holdings;
+  if (!holdings) {
+    return null;
```
kasperg, line 25: Please remove null chaining and checks.

As far as I can tell they should not be necessary.

kasperg (changes requested):
:+1: Some more comments for you to look into.

Please ensure that all TS code does not use the periodikum term. We might also want to remove it from our CSS classes in the design system at some point but for now it should be OK to keep naming in the TS code consistent.

cableman (approved):

kasperg (approved):
:+1: LGTM

PR #114, opened 2022-08-31
Add SearchResultHeader

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFSOEG-120

#### Screenshot of the result
Skærmbillede 2022-08-31 kl  16 22 17

#### Checklist

### DIFF
--- a/src/apps/search-result/search-result.test.ts
+++ b/src/apps/search-result/search-result.test.ts
@@ -1,25 +1,33 @@
 const coverUrlPattern = /^https:\/\/res\.cloudinary\.com\/.*\.(jpg|jpeg|png)$/;
 
 describe("Search Result", () => {
-  it("Should render the search results as expected", () => {
+  it("should render the site", () => {
     cy.visit(
       "/iframe.html?id=apps-search-result--search-result&args=pageSizeDesktop:2;pageSizeMobile:2"
     );
+  });
 
-    cy.log("Check length of search result list:");
+  it("Check search title", () => {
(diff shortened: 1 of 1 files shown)


src/components/search-bar/search-result-header/SearchResultHeader.tsx
```diff
@@ -0,0 +1,21 @@
+import * as React from "react";
+import { useText } from "../../../core/utils/text";
+
+export interface SearchResultHeaderProps {
+  hitcount?: string;
+  q?: string;
```
spaceo, line 6: What are you thoughts about the props being required/optional?
The way the rendering in the component is constructed it does not handle if one of them are empty/falsy

src/apps/search-result/search-result.test.ts
```diff
@@ -1,25 +1,33 @@
...
 describe("Search Result", () => {
-  it("Should render the search results as expected", () => {
+  it("should render the site", () => {
     cy.visit(
       "/iframe.html?id=apps-search-result--search-result&args=pageSizeDesktop:2;pageSizeMobile:2"
     );
+  });
+
+  it("Check search title", () => {
+    cy.contains("Viser resultater for");
```
spaceo, line 11: what about search string and hitcount?

spaceo (changes requested):
Niceness.
I have a few remarks

sinejespersen (approved):

spaceo (approved):

PR #117, opened 2022-09-05
Feature/5411 pagination

sinejespersen:
#### Link to issue

https://platform.dandigbib.org/issues/5411

#### Description

- Rename SearchResultPager to ResultPager (this pr is friends with this pr: https://github.com/danskernesdigitalebibliotek/dpl-design-system/pull/133)
- Add pagesize config
- Create pagination component, the pagination we have is both for single elements and also for "stacked" elements, so the pagination depends on view

#### Screenshot of the result

image

#### Checklist


src/apps/loan-list/utils/pagination.tsx
```diff
@@ -0,0 +1,92 @@
...
+        <>
+          <LoanListItems
+            dueDates={dueDates}
+            loans={displayedLoans}
+            view={view}
+            openModalDueDate={openModalDueDate}
+            selectModalMaterial={selectModalMaterial}
+          />
+          <ResultPager
+            searchItemsShown={displayedLoans.length}
```
spaceo, line 82: SearchItems? :)

src/apps/loan-list/utils/pagination.tsx
```diff
@@ -0,0 +1,92 @@
...
+  }, [dueDates, loans, searchItemsShown, view]);
+
+  useEffect(() => {
+    // When view is changed (from list to stacks or stacks to list)
+    // The items shown are reset to pagesize from config
+    setSearchItemsShown(pageSize);
+  }, [view]);
+
+  return (
+    // eslint-disable-next-line react/jsx-no-useless-fragment
```
spaceo, line 70: I think that either we should make a global consideration wether we want the eslint rule or not.
Until now I have accepted that the rule comes from a recommended set of rules and used div's instead. What do you think?

src/apps/loan-list/utils/helpers.ts
```diff
@@ -53,4 +53,26 @@ export const queryMatchesFaust = (query: string | null) => {
   return returnValue;
 };
 
+export const getStackedSearchItems = (
+  view: string,
```
spaceo, line 57: I think that view should be a uniontype. If it is a global type it could be moved to: src/core/utils/types

src/apps/loan-list/materials/stackable-material.tsx
```diff
@@ -22,7 +26,9 @@ const StackableMaterial: FC<StackableMaterialProps & MaterialProps> = ({
 }) => {
   const t = useText();
   const dispatch = useDispatch();
-
+  const [additionalMaterials] = useState(
+    amountOfMaterialsWithDueDate ? amountOfMaterialsWithDueDate - 1 : 0
```
spaceo, line 30: I haven't see useState been used without a getter before. But I guess it can be useful :)

src/apps/loan-list/utils/pagination.tsx
```diff
@@ -0,0 +1,92 @@
...
+          view,
+          loans,
+          searchItemsShown,
+          dueDates
+        );
+
+        setDisplayedLoans([...stackedLoans]);
+      }
+    }
+  }, [dueDates, loans, searchItemsShown, view]);
```
spaceo, line 61: `searchItemsShown` should be renamed I guess?

src/apps/loan-list/utils/pagination.tsx
```diff
@@ -0,0 +1,92 @@
...
+      const currentPage = page + 1;
+      const itemsOnPage = (currentPage + 1) * pageSize;
+      setPage(currentPage);
+      setSearchItemsShown(itemsOnPage);
+    }
+  };
+
+  useEffect(() => {
+    if (loans) {
+      if (view === "list") {
```
spaceo, line 48: Should be union type I think

src/apps/loan-list/utils/pagination.tsx
```diff
@@ -0,0 +1,92 @@
...
+  const [page, setPage] = useState<number>(0);
+
+  const setPageHandler = () => {
+    if (loans) {
+      const currentPage = page + 1;
+      const itemsOnPage = (currentPage + 1) * pageSize;
+      setPage(currentPage);
+      setSearchItemsShown(itemsOnPage);
+    }
+  };
```
spaceo, line 44: Isn't it possible to make this part generic as well? I am just freestyling here, but could it be possible to create a custom hook that returns the pager component with a contract like this:
```
const [itemsShown, hitcount, pagehandler, ResultPager] = usePager(pageSize);
```
and later on in the rendering:
```
return (
  <div>
    <AlistOfSomething />
    <ResultPager itemsShown={itemsShown} hitcount={hitcount} pagehandler={pagehandler} />
  </div>
);

**Review by spaceo — CHANGES_REQUESTED** [review]
Looks great. I have however some suggestion I would like you to look at. Especially the suggestion of how to make the pageHandler a part of generic functionality

File: src/apps/loan-list/utils/helpers.ts (hunk 3 above)
Comment (sinejespersen) [reply src/apps/loan-list/utils/helpers.ts:57]: I agree

File: src/apps/loan-list/utils/pagination.tsx (hunk 1 above)
Comment (sinejespersen) [reply src/apps/loan-list/utils/pagination.tsx:82]: heh, I'll rename

File: src/apps/loan-list/utils/pagination.tsx (hunk 5 above)
Comment (sinejespersen) [reply src/apps/loan-list/utils/pagination.tsx:61]: heh, yes

File: src/apps/loan-list/utils/pagination.tsx (hunk 7 above)
Comment (sinejespersen) [reply src/apps/loan-list/utils/pagination.tsx:44]: I am not entirely sure what you mean, create a hook that returns the pager component _instead of_ pagination.tsx? And to cover the search results as well?

File: src/apps/loan-list/utils/pagination.tsx (hunk 2 above)
Comment (sinejespersen) [reply src/apps/loan-list/utils/pagination.tsx]: I think its a silly rule, but I've fixed it in a better way. Perhaps thats why the silly rule exists.

File: src/apps/loan-list/utils/pagination.tsx [hunk 8]
```diff
@@ -76,11 +67,7 @@ const Pagination: FC<PaginationProps> = ({
         openModalDueDate={openModalDueDate}
         selectModalMaterial={selectModalMaterial}
       />
-      <ResultPager
-        itemsShown={displayedLoans.length}
-        hitcount={hitcount}
-        setPageHandler={setPageHandler}
-      />
+      {PagerComponent}
```
spaceo, line 70: Isn't it possible to write?:
```
<Pagercomponent />

File: src/apps/search-result/helpers.ts [hunk 9]
```diff
@@ -9,8 +9,8 @@ export const getPageSizeFromDataAttributes = ({
   desktop,
   mobile
 }: PageSizeDataAttributes) => {
-  const { pageSize } = getDeviceConf("search", {
-    search: {
+  const { pageSize } = getDeviceConf("dplConfig", {
```
spaceo, line 12: I don't think dplConfig is the right configuration domain name. Maybe we could use `paginatedLists`?

src/apps/search-result/search-result.tsx
```diff
@@ -64,21 +55,13 @@ const SearchResult: React.FC<SearchResultProps> = ({ q, pageSize }) => {
...
       <SearchResultHeader hitcount={String(hitcount)} q={q} />
       {worksAreLoaded && <SearchResultList resultItems={resultItems} />}
-      {moreWorksToBeLoaded && (
-        <ResultPager
-          itemsShown={searchItemsShown}
-          hitcount={hitcount}
-          setPageHandler={setPageHandler}
-        />
-      )}
+      {PagerComponent}
```
spaceo, line 64: Isn't it possible to write?:
```
<Pagercomponent />

File: src/apps/loan-list/list/loan-list.tsx [hunk 11]
```diff
@@ -223,7 +224,7 @@ const LoanList: FC = () => {
             <Pagination
               dueDates={dueDates}
               loans={loans}
-              view={view}
+              view={view as ListView}
```
spaceo, line 227: It should be possible to set the ListView type [here](https://github.com/danskernesdigitalebibliotek/dpl-react/blob/2637492e9d9db2729ed5a651823c0296c657f3c8/src/apps/loan-list/list/loan-list.tsx#L60) instead. And then you wouldn't need the type assertion.

src/apps/loan-list/modal/due-date-loans-modal.tsx
```diff
@@ -56,7 +56,6 @@ const DueDateLoansModal: FC<DueDateLoansModalProps> = ({
             )}
             <RenewLoansModalContent
               loansModal={loansModal}
-              renewable={renewable}
```
spaceo, line 59: should be removed from props also?

spaceo (changes requested):

src/apps/loan-list/utils/pagination.tsx
sinejespersen: image

src/apps/search-result/helpers.ts
sinejespersen: it also contains "coverTints"

src/apps/loan-list/list/loan-list.tsx
sinejespersen: Done

src/apps/search-result/search-result.tsx
sinejespersen: See above

src/apps/loan-list/utils/pagination.tsx
spaceo: weird

src/core/configuration/index.ts
```diff
@@ -33,5 +32,5 @@ export const getDeviceConf = (
 };
 
 export default {
-  search
+  dplConfig
```
spaceo, line 35: The idea was to put the different configuration scopes into this: search, colors, modal-ids, status-threshold. And have one way of fetching the configuration. We have now two ways of doing it. And calling the scope `dplConfig` is, in my eyes, the same as calling the scope "stuff" or "whatever".
So in my view the right solution here is getting all the configuration here and use the same way of retrieving the configuration.
Regarding the comment about that cover tints is also defined along the pager settings could be solved with separating the two types of configuration into the scopes: "pager" and "covers"

spaceo (changes requested):

PR #121, opened 2022-09-06
Wiremock CORS issues

cableman:
#### Link to issue

#### Description

It turns out the wiremock has support for automatically handle CORS headers (so this PR switches over to use that).

#### Screenshot of the result

N/A

#### Checklist

N/A


sinejespersen (approved):

Taskfile.yml
```diff
@@ -47,6 +47,20 @@ tasks:
       # causes the process to fail. Wait and retry if that is the case
       - docker compose {{ .DOCKER_COMPOSE_FILES }} down --volumes --remove-orphans || (sleep 3 && docker-compose down)
 
+  dev:mocks:start:
+    summary: Start wiremocks
+    cmds:
+      - docker compose {{ .DOCKER_COMPOSE_FILES }} up wiremock-fbs --detach
```
kasperg, line 53: It may be worthwhile using [Docker Compose profiles](https://docs.docker.com/compose/profiles/#enabling-profiles) to control which services are started when instead of relying on multiple `docker-compose.yml` files, (somewhat) complex tasks etc. I only recently discovered this feature myself.

kasperg (approved):
:+1: LGTM

Taskfile.yml
cableman: Have looked into profiles some time back and it's usage was not was i expected...

PR #124, opened 2022-09-08
All props needs to follow the same format

cableman:
#### Link to issue

#### Description

When using dpl-react in the CMS the props need to follow this format.

#### Screenshot of the result

N/A

#### Checklist

N/A


src/apps/loan-list/helpers.ts
```diff
@@ -1,6 +1,6 @@
 import get from "lodash.get";
 import dayjs from "dayjs";
-import { LoanV2 } from "../../core/fbs/model/loanV2";
+import { LoanV2 } from "../../core/fbs/model";
```
sinejespersen, line 4: why this change

cableman: The IDE that i uses says that this is the correct way to import it...

sinejespersen (approved):

PR #129, opened 2022-09-08
Add possibility for using user token in env file

spaceo:
Why not adding that feature to the Dev Experience™ ? [Loud echo]

#### Checklist


Adamik10 (approved):
Great improvements :)

Adamik10 (changes requested):
The integration tests are failing - could you please look into that? 
Otherwise I like this effort! :)

Adamik10 (approved):

cableman (approved):

PR #131, opened 2022-09-09
Reservation modal

spaceo:
#### Link to issue
https://reload.atlassian.net/browse/DDFSOEG-203

### This PR adds a modal that handles the reservation of a material
The modal is opened after the reservation button has been pressed. Here shows the material and user information.
if the user confirms, and if it is approved it shows ReservationSucces  else is shows ReservationError

Here is added: 
- markup and CSS classes from design-system
- a reusable ReservationFormListItem
- ReservationSucces + ReservationError components 
- add all the translations that are needed

**2 help functions:**
- **getNoInterestAfter**: 
rewrite days into months
- **getPreferredLocation**: 
find branch name with id

#### Screenshot of the result
Skærmbillede 2022-09-07 kl  15 24 38
Skærmbillede 2022-09-12 kl  11 14 23

#### Checklist

### DIFF
--- a/src/components/reservation/ReservationFormListItem.tsx
+++ b/src/components/reservation/ReservationFormListItem.tsx
@@ -1,0 +1,33 @@
+import * as React from "react";
+import { useText } from "../../core/utils/text";
+
+interface ReservationFormListItemProps {
+  icon: string;
+  title: string;
+  text: string;
+}
+
+const ReservationFormListItem: React.FC<ReservationFormListItemProps> = ({
+  icon,
+  title,
+  text
+}) => {
+  const t = useText();
+  return (
+    <div className="reservation-modal-list-item">
+      <img src={icon} alt="" />
+      <div className="reservation-modal-list-item-text">
+        <h3 className="text-header-h5">{title}</h3>
+        <p className="text-small-caption">{text ?? `Mangler data`}</p>
+      </div>
+      {/* <button
+        type="button"
+        className="link-tag text-small-caption cursor-pointer"
+      >
+        {t("shiftText")}
+      </button> */}
+    </div>
+  );
+};
+
+export default ReservationFormListItem;
--- a/src/components/reservation/ReservationFormListItem.tsx
+++ b/src/components/reservation/ReservationFormListItem.tsx
@@ -5,12 +5,14 @@ interface ReservationFormListItemProps {
   icon: string;
   title: string;
   text: string;
+  changeHandler?: () => void;
 }
 
 const ReservationFormListItem: React.FC<ReservationFormListItemProps> = ({
   icon,
   title,
-  text
+  text,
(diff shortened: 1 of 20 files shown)


src/components/reservation/ReservationFormListItem.tsx
```diff
@@ -0,0 +1,33 @@
...
+      <div className="reservation-modal-list-item-text">
+        <h3 className="text-header-h5">{title}</h3>
+        <p className="text-small-caption">{text ?? `Mangler data`}</p>
+      </div>
+      {/* <button
+        type="button"
+        className="link-tag text-small-caption cursor-pointer"
+      >
+        {t("shiftText")}
+      </button> */}
```
kasperg, line 28: Please go though this code again.

- `t` is declared but never used
- There should not be commented out code in the codebase unless there is a reason to. If you think the code is should then leave it be with a `TODO` comment for the click handler or the like.
- "Mangler data" is not translated. I am not sure how it can be false anyway. Perhaps if the string is empty? Are we sure we want to print "Mangler data" here anyway?

src/components/reservation/reservation-modal.tsx
```diff
@@ -0,0 +1,183 @@
...
+  const { data: branchData } = branchResponse as { data: AgencyBranch[] };
+  const { data: userData } = userResponse as { data: AuthenticatedPatronV6 };
+  const { data: holdingsData } = holdingsResponse as {
+    data: HoldingsForBibliographicalRecordV3[];
+  };
+  const { reservations, holdings } = holdingsData[0];
+  const { patron } = userData;
+
+  // TODO move to material/helper.ts because it is used in multiple places (Info text under buttons in the material header)
+  const totalReservations = reservations;
```
kasperg, line 83: Consider whether the `totalReservations` variable is needed.

As far as I can tell `reservations`as not used elsewhere.

kasperg (changes requested):
:+1: Good job. There are a few things that I think needs to be looked into.

src/components/reservation/reservation-modal.tsx
```diff
@@ -0,0 +1,173 @@
...
+      onSuccess: () => setReservationDidSuccess(true),
+      onError: () => setReservationDidSuccess(false)
+    });
+  };
+
+  return (
+    <Modal
+      modalId={reservationModalId(faustId)}
+      screenReaderModalDescriptionText={t("screenReaderModalDescriptionText")}
+      closeModalAriaLabelText={t("ariaLabelModalTwoText")}
```
sinejespersen, line 106: these two are missing: screenReaderModalDescriptionText, ariaLabelModalTwoText

sinejespersen (changes requested):

sinejespersen (approved):

PR #136, opened 2022-09-09
Restructure development setup documentation

kasperg:
#### Description

Move token information to the top as these are steps you will need  to perform before you can install the project with the design system.

Restructure steps with and without Docker

Move information about the Adgangsplatformen story to the part of the documentation regarding access tokens as that is where tokens and  this story is mentioned otherwise.

#### Checklist


spaceo (approved):
LGTM 👍

README.md
```diff
@@ -6,5 +6,738 @@
 A set of React components and applications providing self-service features for
 Danish public libraries.
 
+<<<<<<< HEAD
```
sinejespersen, line 9: 🤖

spaceo: Fixed

sinejespersen (approved):

PR #137, opened 2022-09-09
Fix/material-buttons - don't show "find on shelf" button twice on work manifestations

Adamik10:
#### Description
This bug was created during merge conflicts when merging in my last PR.

#### Screenshot of the result
Before:
image

After:
image

#### Checklist

n/a


kasperg (changes requested):
Please check out the warnings raised by this change.

kasperg (approved):

PR #141, opened 2022-09-13
Confirming the reservation

kasperbirch1:
#### Description
This PR use response from mutate (useAddReservationsV2) to determine whether to show success or error component after confirming the reservation.

**ReservationSucces**  shows
- title
- numberInQueue
- preferredPickupBranch

#### Screenshot of the result
Skærmbillede 2022-09-14 kl  11 37 37

if reservationResult === "already_reserved"  **ReservationError**  shows
Skærmbillede 2022-09-14 kl  11 37 50

#### Checklist

### DIFF
--- a/src/components/Buttons/Button.tsx
+++ b/src/components/Buttons/Button.tsx
@@ -12,9 +12,9 @@ export type ButtonProps = {
   size: ButtonSize;
   variant: "outline" | "filled";
   onClick?: () => void;
-  classNames?: string;
+  iconClassNames?: string;
   id?: string;
-  className?: string;
+  classNames?: string;
 };
 
 export const Button: React.FC<ButtonProps> = ({
@@ -25,12 +25,12 @@ export const Button: React.FC<ButtonProps> = ({
   size,
   variant,
   onClick,
-  classNames,
+  iconClassNames,
   id,
-  className
+  classNames
 }) => {
   const iconClassName = `btn-icon ${clsx({ "btn-collapsible": collapsible }, [
-    classNames
+    iconClassNames
   ])}`;
 
   const Icon = React.useCallback(() => {
@@ -57,7 +57,7 @@ export const Button: React.FC<ButtonProps> = ({
     <button
       type="button"
       className={`btn-primary btn-${variant} btn-${size} arrow__hover--right-small ${
-        className ?? ""
+        classNames ?? ""
       }`}
       disabled={disabled}
       onClick={onClick}
--- a/src/components/Buttons/Button.tsx
+++ b/src/components/Buttons/Button.tsx
@@ -14,6 +14,7 @@ export type ButtonProps = {
   onClick?: () => void;
   classNames?: string;
   id?: string;
+  className?: string;
 };
 
 export const Button: React.FC<ButtonProps> = ({
@@ -25,7 +26,8 @@ export const Button: React.FC<ButtonProps> = ({
   variant,
   onClick,
   classNames,
-  id
+  id,
+  className
 }) => {
   const iconClassName = `btn-icon ${clsx({ "btn-collapsible": collapsible }, [
     classNames
@@ -54,7 +56,9 @@ export const Button: React.FC<ButtonProps> = ({
   return (
     <button
       type="button"
-      className={`btn-primary btn-${variant} btn-${size} arrow__hover--right-small`}
+      className={`btn-primary btn-${variant} btn-${size} arrow__hover--right-small ${
+        className ?? ""
+      }`}
       disabled={disabled}
       onClick={onClick}
       id={id}
--- a/src/components/reservation/reservation-modal.tsx
+++ b/src/components/reservation/reservation-modal.tsx
@@ -107,13 +107,10 @@ const ReservationModal = ({ manifestation }: ReservationModalProps) => {
     });
   };
 
-  const reservationResult = reservationResponse?.reservationResults
-    ? reservationResponse.reservationResults[0].result
-    : null;
-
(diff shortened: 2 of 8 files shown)


src/components/reservation/reservation-modal.tsx
```diff
@@ -94,11 +98,23 @@ const ReservationModal = ({ manifestation }: ReservationModalProps) => {
       }
     };
     mutate(batch, {
-      onSuccess: () => setReservationDidSuccess(true),
-      onError: () => setReservationDidSuccess(false)
+      onSuccess: (res) => {
+        // this state is used to show the success or error modal
+        setReservationResponse(res);
+        // because after a successful reservation the holdings (reservations) are updated
+        queryClient.invalidateQueries(getGetHoldingsV3QueryKey());
```
kasperg, line 105: :+1: for explaining the relationship here.

src/components/reservation/reservation-modal.tsx
```diff
@@ -94,11 +98,23 @@ const ReservationModal = ({ manifestation }: ReservationModalProps) => {
...
     });
   };
 
+  const reservationResult = reservationResponse?.reservationResults
+    ? reservationResponse.reservationResults[0].result
+    : null;
+
+  const reservationDetails = reservationResponse?.reservationResults
+    ? reservationResponse.reservationResults[0].reservationDetails
+    : null;
```
kasperg, line 116: Consider if this can be simplified with more null chaining e.g. 
`reservationResult =reservationResponse?.reservationResults[0]?.reservationDetails;`

src/components/reservation/reservation-modal.tsx
```diff
@@ -107,67 +123,73 @@ const ReservationModal = ({ manifestation }: ReservationModalProps) => {
...
     >
-      {reservationDidSuccess && patron && (
+      {reservationResult === "success" && reservationDetails && patron && (
         <ReservationSucces
           modalId={reservationModalId(faustId)}
           title={titles.main[0]}
           preferredPickupBranch={getPreferredLocation(
             patron.preferredPickupBranch,
             branchData
           )}
```
kasperg, line 133: Please get the branch from the reservation `reservationDetails.pickupBranch` instead of from the patron.

src/components/reservation/reservation-modal.tsx
```diff
@@ -94,11 +98,23 @@ const ReservationModal = ({ manifestation }: ReservationModalProps) => {
...
+        setReservationResponse(res);
+        // because after a successful reservation the holdings (reservations) are updated
+        queryClient.invalidateQueries(getGetHoldingsV3QueryKey());
+      }
     });
   };
 
+  const reservationResult = reservationResponse?.reservationResults
+    ? reservationResponse.reservationResults[0].result
+    : null;
```
kasperg, line 112: Please get the result of the reservation operation as a boolean from `reservationResponse.success`.

This is more clean.

src/components/reservation/reservation-modal.tsx
```diff
@@ -107,67 +123,73 @@ const ReservationModal = ({ manifestation }: ReservationModalProps) => {
...
 
-      {reservationDidSuccess === null && (
-        <section className="reservation-modal">
-          <header className="reservation-modal-header">
-            <Cover pid={pid as Pid} size="medium" animate />
-            <div className="reservation-modal-description">
-              <div className="reservation-modal-tag">
-                {materialTypes[0].specific}
+      {reservationResult !== "success" &&
+        reservationResult !== "already_reserved" && (
```
kasperg, line 147: Please revisit this. As mentioned there are many more results apart from "success" and "already_reserved" which would cause the modal to appear again. As I see it we only want to display the form if there no request has been made.

Also consider moving the form into a separate component like we have for error and success. The modal component is growing quite large.

src/components/reservation/reservation-modal.tsx
```diff
@@ -107,67 +123,73 @@ const ReservationModal = ({ manifestation }: ReservationModalProps) => {
...
-      {reservationDidSuccess && patron && (
+      {reservationResult === "success" && reservationDetails && patron && (
         <ReservationSucces
           modalId={reservationModalId(faustId)}
           title={titles.main[0]}
           preferredPickupBranch={getPreferredLocation(
             patron.preferredPickupBranch,
             branchData
           )}
+          numberInQueue={String(reservationDetails.numberInQueue)}
```
kasperg, line 134: Please use the original types from the response if possible.

As I see it is typed as [an optional number from the API spec](https://github.com/reload/dpl-react/blob/f46afa9f427e45f37afe3725e52322e4c849a5d2/src/core/fbs/model/reservationDetailsV2.ts#L16-L16). I do not see a reason to change this.

src/components/Buttons/Button.tsx
```diff
@@ -14,6 +14,7 @@ export type ButtonProps = {
   onClick?: () => void;
   classNames?: string;
   id?: string;
+  className?: string;
```
kasperg, line 17: Please rework these prop names.

It is quite confusing to have two similar sounding names `classNames` and `className` which does different things. The original `classNames` is applies to the icon while the new `className` is applied to the entire button.

I suggest renaming these to `classNames` for the button and `iconClassNames` for the icon.

kasperg (changes requested):
This PR has some misunderstandings regarding how reservations and errors should be handled and reported. We can look into this together.

sinejespersen (approved):

kasperg (approved):
:+1: LGTM

PR #142, opened 2022-09-13
Make FBS and Publizon URLs configurable 

cableman:
#### Link to issue

#### Description

Make FBS/Publizon mocks available for tests in cypress in actions. 

#### Screenshot of the result

N/A

#### Checklist

N/A


src/core/utils/helpers/fetcherHelper.ts
```diff
@@ -0,0 +1,19 @@
...
+
+interface ConfigPossibilities {
+  fbs: string;
+  publizon: string;
+}
+
+export const configTypes: ConfigPossibilities = {
+  fbs: "fbsBaseUrl",
+  publizon: "publizonBaseUrl"
+};
```
spaceo, line 11:
```suggestion
export const configTypes: ConfigPossibilities = {
  fbs: "fbsBaseUrl",
  publizon: "publizonBaseUrl"
} as const;
```

spaceo: If you write `as const` you make sure that the object is immutable and emphasize that it is meant for static "configuration" usage.

src/core/publizon/publizonConfig.ts
```diff
@@ -0,0 +1,21 @@
+/**
```
spaceo, line 1: This file is deprecated now, right?

src/core/fbs/fbsConfig.ts
```diff
@@ -0,0 +1,21 @@
+/**
```
spaceo, line 1: This file is deprecated now, right?

src/core/utils/helpers/fetcherHelper.ts
```diff
@@ -0,0 +1,19 @@
+import { store } from "../../store";
```
spaceo, line 1: I think you should rename the file to fetcher.ts. That is the naming convention we have used for other helpers.
Alternatively we need to rename all helper files 😈

spaceo (changes requested):
Great. This is more simple.
I have a few comments

src/core/publizon/publizonConfig.ts
sinejespersen: deleted

src/core/fbs/fbsConfig.ts
sinejespersen: deleted

src/core/utils/helpers/fetcherHelper.ts
sinejespersen: Sure

spaceo (approved):

PR #144, opened 2022-09-13
Improve build/watch workflow

kasperg:
#### Description

This PR should make it easier for developers to notice problems in their code related to eslint before they pass along a pull request for review. It does so in two steps:

1. Make the eslint GitHub Action fail for both warnings and errors. There are multiple occations where reviewers have had to reject a PR because there are warnings shown as annotations in the Files Changed tab. These are likely missed because the eslint job succeeded. We do not want warnings in our code. Either they should be addressed through changes/fixes or eslint updates.
2. Display eslint problems in the developer console. Warnings can be hard to miss if you do not have eslint integrated with your IDE or if your IDE does not notify you clearly about problems which may occur outside the current scope you are working in (e.g. imports). Showing the warnings in your console should make it easier to notice.

Please check the commit comments for more commentary.

[927ea0a](https://github.com/danskernesdigitalebibliotek/dpl-react/pull/144/commits/927ea0af6eab51f36994c05b9e9cfe117ede9fac) changes 1) the reporting in GitHub. The rest is 2). I have tested this both with and without Docker on OSX.

If you do not agree with the entire change I am fine with continuing with 1).

#### Screenshot of the result

Demo without Docker:

https://user-images.githubusercontent.com/73966/190082287-2d627657-bc95-4afb-8395-a1026b79aa94.mp4

Screenshot with Docker:

Monosnap dpl-react – MaterialAvailabilityText tsx 2022-09-14 09-12-02

#### Checklist


Adamik10 (approved):
I like this very much. Both of the parts you describe. It moves issues from the "you gotta remember to check for this" category to "you will be notified if your code is not as it should be". And I'm a big fan of that. I tested it in VSCode and it seems to be working as you describe :) 
Yay for automatisation! 🎉

spaceo:
Great improvement. Looking forward to work with it.

cableman (approved):

PR #146, opened 2022-09-14
Test modal reservationsflow 

kasperbirch1:
#### Description

We tested:
- that we can open the reservation modal and there is user data 
- that we can approve the reservation and get a receipt with our number in the queue

#### Checklist


src/apps/material/material.test.ts
```diff
@@ -35,9 +35,53 @@ describe("Material", () => {
...
+  it("Open modal by clicking on reserver button (reserver Dummy bog) and close it with the x bottom", () => {
+    cy.contains("button:visible", "reserver Dummy bog").click();
+    cy.contains("Afhentes på");
+    cy.contains("Hovedbiblioteket");
+    cy.contains("22403525");
+    cy.contains("kasperbirch1@gmail.com");
+    cy.get(`[aria-label="Luk reservation modal"]`).click({
+      multiple: true,
+      force: true
+    });
```
kasperg, line 47: Consider how you can drop the `multiple/force` element. You probably want to use the visible selector here again.

sinejespersen (approved):

PR #147, opened 2022-09-14
Cleanup types and destruct props

spaceo:
#### Description

Cleaning up work and manifestation types and destructing more variables to cleanup code

#### Checklist

### DIFF
--- a/src/apps/material/helper.ts
+++ b/src/apps/material/helper.ts
@@ -23,10 +23,10 @@ export const getManifestationFromType = (
   const allManifestations = orderManifestationsByYear(manifestations);
 
   const allManifestationsThatMatchType = allManifestations.filter(
-    (item) => getManifestationType(item as Manifestation) === type
+    (item) => getManifestationType(item) === type
   );
 
-  return allManifestationsThatMatchType.shift() as Manifestation;
+  return allManifestationsThatMatchType.shift();
 };
 
 export const getWorkDescriptionListData = ({
--- a/src/apps/material/helper.ts
+++ b/src/apps/material/helper.ts
@@ -1,8 +1,4 @@
 import { ListData } from "../../components/material/MaterialDetailsList";
-import {
-  ManifestationsSimpleFieldsFragment,
-  WorkMediumFragment
-} from "../../core/dbc-gateway/generated/graphql";
 import { AgencyBranch, HoldingsV3 } from "../../core/fbs/model";
 import {
   creatorsToString,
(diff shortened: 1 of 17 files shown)


src/components/material/material-buttons/physical/MaterialButtonsPhysical.tsx
```diff
@@ -3,23 +3,22 @@ import { ManifestationsSimpleFieldsFragment } from "../../../../core/dbc-gateway
...
-import { FaustId, Pid } from "../../../../core/utils/types/ids";
+import { Pid } from "../../../../core/utils/types/ids";
 import MaterialButtonCantReserve from "../generic/MaterialButtonCantReserve";
 import MaterialButtonLoading from "../generic/MaterialButtonLoading";
 import MaterialButtonUserBlocked from "../generic/MaterialButtonUserBlocked";
 import MaterialButtonReservePhysical from "./MaterialButtonPhysical";
 
 export interface MaterialButtonsPhysicalProps {
-  manifestation: ManifestationsSimpleFieldsFragment;
+  manifestation: ManifestationsSimpleFieldsFragment & { pid: Pid };
```
kasperg, line 13: Should this not use the new `Manifestation` type?

src/components/search-result-list/search-result.list.tsx
```diff
@@ -12,7 +13,10 @@ const SearchResultList: React.FC<SearchResultListProps> = ({ resultItems }) => {
     <ul className="search-result-page__list my-32">
       {resultItems.map((item, i) => (
         <li key={item.workId}>
-          <SearchResultListItem item={item} coverTint={getCoverTint(i)} />
+          <SearchResultListItem
+            item={item as Work}
```
kasperg, line 17: Could `resultItems` also be instances of `Work` instead of `WorkSmallFragment` so we avoid the `as`?

src/core/utils/types/entities.ts
```diff
@@ -0,0 +1,17 @@
...
+  ManifestationsSimpleFieldsFragment,
+  WorkMediumFragment
+} from "../../dbc-gateway/generated/graphql";
+import { Pid, WorkId } from "./ids";
+
+export type Manifestation = Omit<ManifestationsSimpleFieldsFragment, "pid"> & {
+  pid: Pid;
+};
+export type Work = Omit<WorkMediumFragment, "workId"> & {
+  workId: WorkId;
```
kasperg, line 11: Excellent!

kasperg (approved):
:+1: I love the proper typings of these ids.

I am not a fan of these destructing function arguments.

src/apps/material/helper.ts
```diff
@@ -11,35 +7,35 @@ import {
...
   type: string,
-  work: WorkMediumFragment
+  { manifestations: { all: manifestations } }: Work
 ) => {
-  const allManifestations = orderManifestationsByYear(work.manifestations);
+  const allManifestations = orderManifestationsByYear(manifestations);
 
   const allManifestationsThatMatchType = allManifestations.filter(
-    (item) => getManifestationType(item) === type
+    (item) => getManifestationType(item as Manifestation) === type
```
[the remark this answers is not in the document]
spaceo: Agreed. Sometimes the ts server in vs code is acting weird. I opened up and closed some files and then the `as` was not necessary anymore

src/apps/material/helper.ts
```diff
@@ -11,35 +7,35 @@ import {
...
   );
 
-  return allManifestationsThatMatchType.shift();
+  return allManifestationsThatMatchType.shift() as Manifestation;
```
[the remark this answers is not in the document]
spaceo: You're right

src/components/material/material-buttons/physical/MaterialButtonsPhysical.tsx
spaceo: Yes it should

src/components/search-result-list/search-result.list.tsx
spaceo: Yes skipped it initially because I ended up in a rabbit hole. But now I found my way out

PR #150, opened 2022-09-19
Only render SearchResultHeader + SearchResultList if worksAreLoaded

kasperbirch1:
#### Description

This PR does 3 things

- Hide pager component if no results
- Only render SearchResultHeader + SearchResultList if worksAreLoaded
This prevents that there will be a flash of a zero result search page before the result are ready
- Adding debug snippet

#### Checklist


sinejespersen (approved):

spaceo (approved):
Approved - but there is a couple of warnings

PR #154, opened 2022-09-20
Makes it possible to change the pick-up point for a reservation

kasperbirch1:
#### Description
This PR gives the opportunity to change where you want to collect the material
 

A useState (selectedBranch) has been added in ReservationModal which sets the user's preferredPickupBranch with a useEffect. 

setState is prop drilling until ModalReservationFormSelect from where it can be changed

#### Screenshot of the result
Skærmbillede 2022-09-20 kl  15 57 55
Skærmbillede 2022-09-20 kl  15 57 59
Skærmbillede 2022-09-20 kl  15 58 05

#### Checklist

### DIFF
--- a/src/components/reservation/helper.ts
+++ b/src/components/reservation/helper.ts
@@ -1,6 +1,21 @@
 import { UseConfigFunction } from "../../core/utils/config";
+import { AgencyBranch } from "../../core/fbs/model";
 
 export const smsNotificationsIsEnabled = (config: UseConfigFunction) =>
   config("smsNotificationsForReservationsEnabledConfig") === "1";
+
+export const getPreferredLocation = (id: string, array: AgencyBranch[]) => {
+  const locationItem = array.find((item) => item.branchId === id);
+  return locationItem ? locationItem.title : id;
+};
+
+export const getPreferredLocationText = (
+  preferred: string,
+  selected: string,
+  branches: AgencyBranch[]
+) =>
+  selected
+    ? getPreferredLocation(selected, branches)
+    : getPreferredLocation(preferred, branches);
 
 export default {};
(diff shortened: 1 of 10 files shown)


src/components/reservation/helper.ts
```diff
@@ -0,0 +1,17 @@
...
+};
+
+export const getPreferredLocationText = (
+  preferred: string,
+  selected: string,
+  branches: AgencyBranch[]
+) =>
+  selected
+    ? getPreferredLocation(selected, branches)
+    : getPreferredLocation(preferred, branches);
```
kasperg, line 15: Consider what types are used for the arguments of this function.

The use of types here confused me quite a bit. We test for `selected` which is a string. This should only evaluate to `false` if the string is empty. However if the user does not have a preferred branch then I would expect `preferred` to be `null`. Perhaps this goes back to the API response and there is nothing we can do about this.

src/components/reservation/helper.ts
```diff
@@ -0,0 +1,17 @@
+import { AgencyBranch } from "../../core/fbs/model";
+
+export const getPreferredLocation = (id: string, array: AgencyBranch[]) => {
+  const locationItem = array.find((item) => item.branchId === id);
+  return locationItem ? locationItem.title : id;
+};
+
+export const getPreferredLocationText = (
```
kasperg, line 8: Consider renaming location to branch in these function names.

Branch is a concept used by FBS so unless we have a good reason we could just continue using that concept instead of introducing our own.

kasperg (changes requested):
:+1: Good stuff. I have a few things for you to consider.

Most importantly remember translate texts.

src/components/reservation/helper.ts
kasperbirch1: eslint-disable-next-line react/jsx-props-no-spreading

kasperg (approved):
:+1: LGTM

sinejespersen (approved):

PR #157, opened 2022-09-21
Ved biblioteks fravalgt sms ddfsoeg 209

spaceo:
#### Link to issue

https://reload.atlassian.net/browse/DDFSOEG-209

#### Description

This PR handles two things:
* Introduces withConfig/useConfig for handlig config specific data attributes
* Adds the first config data attribute for controlling wether an agency can send sms notifications about reservations

#### Checklist

I principle I don't like to reuse the concept from withUrl and withText. But my thought was: this will do for now.
I general I would like to change two things regarding the with*/use* systems:
* Connect them into one slice and one hoc that divides the data into the same slice of redux under the keys: "url", "text", "config". Data-storing-wise the three contexts does the same.
* Make one data attribute for eeach context with json serialized data. The amount of data attributes is getting ridiculously high :)


kasperg (approved):
:+1: LGTM

We should probably consider adding some documentation about these more or less elaborate systems we have for passing data into components - texts, urls, configurations etc.

I like how you wrap the configuration `1` in a function `smsNotificationsIsEnabled` to ensure we get a proper boolean.

We could consider if we could use the same approach for urls.

sinejespersen (approved):

PR #160, opened 2022-09-22
Parallel reservering

spaceo:
#### Link to issue

https://reload.atlassian.net/browse/DDFSOEG-212

#### Description

Adds parallel reservation for fictional material

#### Checklist


src/core/utils/UseAvailableManifestations.tsx
```diff
@@ -0,0 +1,88 @@
...
+      if (type) {
+        setAvailableManifestations(filterManifestationsByType(type, available));
+        setUnAvailableManifestations(
+          filterManifestationsByType(type, unavailable)
+        );
+        return;
+      }
+      // Otherwise set the available and unavailable manifestations as is.
+      setAvailableManifestations(available);
+      setUnAvailableManifestations(unavailable);
```
kasperg, line 72: Consider filtering by type before fetching availability information instead of after.

As I see it there is no reason to fetch information for types that are not relevant to us.

src/components/reservation/ReservationModalBody.tsx
```diff
@@ -34,24 +30,27 @@ import {
...
+const ReservationModalBody = ({
+  mainManifestation,
+  mainManifestation: { pid, materialTypes, titles, edition },
+  parallelManifestations
 }: ReservationModalProps) => {
+  const mainManifestationType = getManifestationType(mainManifestation);
+  const { availableManifestations } = UseAvailableManifestations({
+    manifestations: parallelManifestations ?? [mainManifestation],
+    type: mainManifestationType
+  });
```
kasperg, line 53: Please consider if filtering the manifestations by availability is actually what we want.

As I see it: An available manifestation is a manifestation where there is at least one holding at a library which can be picked up by a patron. However when we create a parallel reservation we should not limit ourselves to available manifestations. If we do so then patrons cannot reserve manifestations where all the holdings are out on loan already.

I tried to go through the specs but could not find any mention of this myself.

kasperg (approved):
:+1: A few considerations for you. Otherwise LGTM

src/components/reservation/helper.ts
```diff
@@ -0,0 +1,48 @@
...
+  materialIsFiction
+} from "../../core/utils/helpers/general";
+import { UseTextFunction } from "../../core/utils/text";
+import { Manifestation } from "../../core/utils/types/entities";
+
+const constructReservation = ({ pid }: Manifestation) => {
+  const faustId = convertPostIdToFaustId(pid);
+
+  return { recordId: faustId };
+};
```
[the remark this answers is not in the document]
spaceo: Good point ;)

src/core/utils/UseAvailableManifestations.tsx
spaceo: Another good point

src/components/reservation/ReservationModalBody.tsx
spaceo: Replaced with reservable and refactored...

sinejespersen (approved):

PR #161, opened 2022-09-22
Fix broken material test

kasperbirch1:
#### Fix broken material test
- cy.visit() is moved to the bottom of beforeEach 
- instead of opening all the details, only the last one is now selected, which is the one for the material

#### Checklist


Adamik10 (approved):
Glad you fixed this 🥇 🥇 🥇 
Great job! :)

sinejespersen (approved):

PR #163, opened 2022-09-23
Find on shelf - open branch disclosure if it's the only one present

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFSOEG-223

#### Description
This PR makes changes to the Find on shelf modal and the disclosure component. 

#### Checklist

It is only the last two commits that you need to address (b52d2a3 and e4a7ec9).
The rest is from find-on-shelf-manifestation-info branch that has its own PR.


kasperg (approved):
:+1: LGTM

sinejespersen (approved):

PR #164, opened 2022-09-23
Feature/find on shelf/main branch first

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFSOEG-224

#### Description
In this PR we sort find on shelf branch disclosures - showing the main branch first if there is one. 

#### Screenshot of the result
image

#### Checklist

n/a

### DIFF
--- a/src/components/find-on-shelf/FindOnShelfModal.tsx
+++ b/src/components/find-on-shelf/FindOnShelfModal.tsx
@@ -75,6 +75,18 @@ const FindOnShelfModal: FC<FindOnShelfModalProps> = ({
     );
   });
 
+  // Sorting of the final data below.
+  const finalDataMainBranchFirst = [
+    ...finalData.filter((manifestationHoldings) => {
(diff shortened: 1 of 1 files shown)


spaceo (approved):

src/components/find-on-shelf/FindOnShelfModal.tsx
```diff
@@ -75,6 +75,17 @@ const FindOnShelfModal: FC<FindOnShelfModalProps> = ({
     );
   });
 
+  // Sorting of the final data below.
```
sinejespersen, line 78: maybe a comment, cause this is not very selfexplanatory. why "00"?

sinejespersen (approved):

src/components/find-on-shelf/FindOnShelfModal.tsx
Adamik10: Very good. I am being more specific in the comments in the components now :) Thanks!

PR #169, opened 2022-09-23
Feature/5379 reservation list files created

sinejespersen:
#### Link to issue

#### Description

add reservation list files, obviously there is a bit more work to do on this

#### Screenshot of the result

image

#### Checklist

### DIFF
--- a/src/apps/reservation-list/list/reservation-list.tsx
+++ b/src/apps/reservation-list/list/reservation-list.tsx
@@ -1,0 +1,33 @@
+import React, { useEffect, useState, FC } from "react";
+import { useText } from "../../../core/utils/text";
+import { useGetReservationsV2 } from "../../../core/fbs/fbs";
+import { ReservationDetailsV2 } from "../../../core/fbs/model/reservationDetailsV2";
+
+const ReservationList: FC = () => {
+  const t = useText();
+  const { isSuccess, data } = useGetReservationsV2();
+  const [reservations, setReservations] = useState<ReservationDetailsV2[]>([]);
+
+  useEffect(() => {
+    if (isSuccess && data) {
+      setReservations(data);
+    }
+  }, [isSuccess, data]);
+
+  return (
+    <>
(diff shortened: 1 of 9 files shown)


src/apps/reservation-list/list/reservation-list.tsx
```diff
@@ -0,0 +1,33 @@
+import React, { useEffect, useState, FC } from "react";
+import { useText } from "../../../core/utils/text";
+import { useGetReservationsV2 } from "../../../core/fbs/fbs";
+import { ReservationDetailsV2 } from "../../../core/fbs/model/reservationDetailsV2";
+
+const ReservationList: FC = () => {
+  const t = useText();
+  const { isSuccess, data } = useGetReservationsV2();
+  const [reservations, setReservations] = useState<ReservationDetailsV2[]>([]);
```
spaceo, line 9: You could also use null as initial value.
Then you could distinguish between initial value and value with the response.
If the response returns an empty array, you would know that `reservations ` is an empty array because the service said it was.
Does it make sense?

src/apps/reservation-list/list/reservation-list.tsx
```diff
@@ -0,0 +1,33 @@
...
+    <>
+      <h1 className="text-header-h1 m-32">{t("reservationListHeaderText")}</h1>
+      <div className="dpl-list-buttons m-32">
+        <h2 className="dpl-list-buttons__header">
+          {t("reservationListPhysicalLoansTitleText")}
+          <div className="dpl-list-buttons__power">{reservations.length}</div>
+        </h2>
+      </div>
+      {/* Todo */}
+      {reservations.length > 0 &&
```
spaceo, line 27: If you implement what I suggested above you could boil the check down to:
```
{reservations &&
```

spaceo (commented):
LGTM. Have a suggestion...

src/apps/reservation-list/list/reservation-list.tsx
sinejespersen: yes, but I think I'll be using the [] for the ```<emptylist``` component

PR #170, opened 2022-09-26
Sort find on shelf modal library branches alphabetically

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFSOEG-226

#### Description
This PR orders find on shelf modal library branches alphabetically (with the exception of the main branch, which is always first).

#### Screenshot of the result
image

#### Checklist

n/a

### DIFF
--- a/src/components/find-on-shelf/FindOnShelfModal.tsx
+++ b/src/components/find-on-shelf/FindOnShelfModal.tsx
@@ -78,8 +78,13 @@ const FindOnShelfModal: FC<FindOnShelfModalProps> = ({
   });
 
   // Sorting of the data below to show branches & manifestations in the correct order.
+  const finalDataAlphabetical = finalData.sort(
+    (a: ManifestationHoldings, b: ManifestationHoldings) => {
+      return a[0].holding.branch.title > b[0].holding.branch.title ? 1 : -1;
+    }
+  );
   // "00" is the ending of beanchIds for branches that are considered main.
-  const finalDataMainBranchFirst = finalData.sort(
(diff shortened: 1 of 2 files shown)


src/components/find-on-shelf/FindOnShelfModal.tsx
```diff
@@ -78,8 +78,13 @@ const FindOnShelfModal: FC<FindOnShelfModalProps> = ({
   });
 
   // Sorting of the data below to show branches & manifestations in the correct order.
+  const finalDataAlphabetical = finalData.sort(
+    (a: ManifestationHoldings, b: ManifestationHoldings) => {
+      return a[0].holding.branch.title > b[0].holding.branch.title ? 1 : -1;
```
kasperg, line 83: Consider using [`localeCompare()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare) when comparing strings for sorting.

kasperg (approved):
:+1: LGTM

src/components/find-on-shelf/FindOnShelfModal.tsx
Adamik10: Great suggestion, thx!
After some research and considerations, I opted in to follow it :) There are not that many branches we need to sort through so the [not-greatest-speed](https://stackoverflow.com/questions/14677060/400x-sorting-speedup-by-switching-a-localecompareb-to-ab-1ab10) of .localeCompare() won't burden us.

sinejespersen (approved):

PR #171, opened 2022-09-26
Feature/find on shelf/show unavailable branches last

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFSOEG-225

#### Description
This PR introduces sorting library branches in "find on shelf modal" in **groups** - branches that have at least one available manifestation are shown first and branches without any available specimens are shown last (with the exception of any main branches that are shown on the top regardless).

#### Screenshot of the result
image

#### Checklist

n/a

### DIFF
--- a/src/components/find-on-shelf/FindOnShelfModal.tsx
+++ b/src/components/find-on-shelf/FindOnShelfModal.tsx
@@ -1,6 +1,6 @@
 import * as React from "react";
 import { FC } from "react";
-import { partition } from "lodash.partition";
+import partition from "lodash.partition";
 import { isAnyManifestationAvailableOnBranch } from "../../apps/material/helper";
 import { useGetHoldingsV3 } from "../../core/fbs/fbs";
 import {
--- a/src/components/find-on-shelf/FindOnShelfModal.tsx
+++ b/src/components/find-on-shelf/FindOnShelfModal.tsx
@@ -138,7 +138,7 @@ const FindOnShelfModal: FC<FindOnShelfModalProps> = ({
             <div className="text-small-caption modal-find-on-shelf__caption">
               {`${finalData.length} ${t("librariesHaveTheMaterialText")}`}
             </div>
-            {finalDataToShow.map((libraryBranch) => {
+            {finalDataToShow.map((libraryBranch: ManifestationHoldings) => {
               return (
                 <Disclosure
                   key={libraryBranch[0].holding.branch.branchId}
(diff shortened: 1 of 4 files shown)


src/components/find-on-shelf/FindOnShelfModal.tsx
```diff
@@ -78,15 +78,38 @@ const FindOnShelfModal: FC<FindOnShelfModalProps> = ({
...
+  const availableManifestationHoldings = finalData.filter(
+    (manifestationHolding: ManifestationHoldings) => {
+      return isAnyManifestationAvailableOnBranch(manifestationHolding);
+    }
+  );
+  const unavailableManifestationHoldings = finalData.filter(
+    (manifestationHolding: ManifestationHoldings) => {
+      return !isAnyManifestationAvailableOnBranch(manifestationHolding);
     }
   );
```
kasperg, line 99: Instead of filtering the same array twice we could use something like `_.partition()` to split the array in into available and unavailable parts.

The performance gained from avoiding this is negligeable so the primary point would be to shorten the code. @spaceo has opinions on when/if we should use Lodash but [`_partition()` is at least not one of the functions that are easily replaced](https://github.com/you-dont-need/You-Dont-Need-Lodash-Underscore).

src/components/find-on-shelf/FindOnShelfModal.tsx
```diff
@@ -78,15 +78,38 @@ const FindOnShelfModal: FC<FindOnShelfModalProps> = ({
   });
 
   // Sorting of the data below to show branches & manifestations in the correct order.
```
kasperg, line 80: Consider elaborating what the "correct" order is in a comment to help provide an overview of the following code:

1. Main library branch first
2. Library branches sorted alphabetically with available holdings second
3. Library branches sorted alphabetically without available holdings third

You could even consider moving sorting into a separate function to reduce the complexity of the component code itself.

kasperg (approved):
:+1: LGTM

src/components/find-on-shelf/FindOnShelfModal.tsx
Adamik10: Very good call with more precise comment. 
I was toying with the idea of throwing it out of the component, but then I ended up with the thought that it's so specific to this component that I think it would make sense to keep it in.

src/components/find-on-shelf/FindOnShelfModal.tsx
Adamik10: @sinejespersen, @spaceo will talk to you about this :)

sinejespersen: @Adamik10 I've talked to Mikkel, Lodash partition is fine by me

sinejespersen (approved):

PR #172, opened 2022-09-28
Add reserve a periodical logic

kasperbirch1:
#### This PR makes it possible to reserve a periodical material 
To find out if it is a periodical we need to look for manifestation materialTypes and not source as we previously thought

Other changes:
- adds an extra story to be able to display a periodical
- change periodicalSelect (useState) to type GroupListItem instead of string

#### Screenshot of the result
Skærmbillede 2022-09-26 kl  13 57 01
Skærmbillede 2022-09-26 kl  13 57 06
Skærmbillede 2022-09-26 kl  13 57 29

#### Checklist

### DIFF
--- a/src/apps/material/material.dev.tsx
+++ b/src/apps/material/material.dev.tsx
@@ -515,6 +515,14 @@ export default {
   }
 } as ComponentMeta<typeof MaterialEntry>;
 
-export const Material: ComponentStory<typeof MaterialEntry> = (
+const Template: ComponentStory<typeof MaterialEntry> = (
   args: MaterialEntryProps
 ) => <MaterialEntry {...args} />;
+
+export const Default = Template.bind({});
+Default.args = {};
+
+export const Periodical = Template.bind({});
+Periodical.args = {
+  wid: "work-of:870970-basis:06373674"
+};
--- a/src/apps/material/material.test.ts
+++ b/src/apps/material/material.test.ts
@@ -2,14 +2,23 @@ const coverUrlPattern = /^https:\/\/res\.cloudinary\.com\/.*\.(jpg|jpeg|png)$/;
 
 describe("Material", () => {
   it("Does the Material have title?", () => {
+    cy.visit(
+      "/iframe.html?args=&id=apps-material--default&viewMode=story&type=bog"
+    );
     cy.get(".text-header-h1").should("be.visible");
   });
 
   it("Check that cover has a src", () => {
+    cy.visit(
+      "/iframe.html?args=&id=apps-material--default&viewMode=story&type=bog"
+    );
     cy.get("img").should("have.attr", "src").and("match", coverUrlPattern);
   });
 
   it("Does the material have favourite buttons?", () => {
+    cy.visit(
+      "/iframe.html?args=&id=apps-material--default&viewMode=story&type=bog"
+    );
     cy.get(".button-favourite").should(
       "have.attr",
       "aria-label",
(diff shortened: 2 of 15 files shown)


src/apps/material/material.dev.tsx
```diff
@@ -18,7 +18,8 @@ export default {
     },
     wid: {
       name: "Work ID",
-      defaultValue: "work-of:870970-basis:52557240",
+      defaultValue: "work-of:870970-basis:06373674",
+      // defaultValue: "work-of:870970-basis:52557240",
```
kasperg, line 22: Please do not leave code in comments without context.

If this is something that we want to keep around then at least add an explanation of its relevance. An alternative to the current approach would be to define two stories: One for a "standard" material and one for a periodical and set the standard values accordingly.

src/components/material/MaterialHeader.tsx
```diff
@@ -90,7 +90,9 @@ const MaterialHeader: React.FC<MaterialHeaderProps> = ({
           />
         </div>
 
-        {manifestation?.source?.includes("bibliotekskatalog") && (
+        {manifestation?.materialTypes
+          ?.map((i) => i.specific.toLowerCase())
+          .includes("periodikum") && (
```
kasperg, line 95: Please move this logic outside of the return value.

It is somewhat complex and should either be moved to an `isPeriocial` boolean variable at the top of this function or create a new `isPeriodical()` helper function.

src/components/material/MaterialPeriodical.tsx
```diff
@@ -32,12 +34,26 @@ const MaterialPeriodical: FC<MaterialPeriodicalProps> = ({
     })
     .flat();
 
-  return (
-    <MaterialPeriodicalSelect
-      groupList={groupObjectArrayByProperty(materialsPeriodical, "volumeYear")}
-      selectPeriodicalSelect={selectPeriodicalSelect}
-    />
+  const groupByvolumeYear = groupObjectArrayByProperty(
```
kasperg, line 37:
```suggestion
  const groupByVolumeYear = groupObjectArrayByProperty(
```

Proper casing makes the code a bit easier to read.

src/components/material/MaterialPeriodical.tsx
```diff
@@ -32,12 +34,26 @@ const MaterialPeriodical: FC<MaterialPeriodicalProps> = ({
...
+  const lastEdition = lastYear ? groupByvolumeYear[lastYear].shift() : null;
+
+  if (lastYear && lastEdition) {
+    return (
+      <MaterialPeriodicalSelect
+        lastEdition={lastEdition as GroupListItem}
+        groupList={groupByvolumeYear as GroupList}
+        lastYear={lastYear}
+        selectPeriodicalSelect={selectPeriodicalSelect}
+      />
```
kasperg, line 52: Please move the `lastYear` and `lastEdition` inside the `MaterialPeriodicalSelect` component.

I do not understand why we need to define the last year and edition outside of the component and pass it as props. All the information is already available within the `groupList` prop.

kasperg (changes requested):

src/apps/material/material.test.ts
```diff
@@ -2,14 +2,23 @@ const coverUrlPattern = /^https:\/\/res\.cloudinary\.com\/.*\.(jpg|jpeg|png)$/;
 
 describe("Material", () => {
   it("Does the Material have title?", () => {
+    cy.visit(
+      "/iframe.html?args=&id=apps-material--default&viewMode=story&type=bog"
```
spaceo, line 6: All the storybook urls have an empty arg. Is that itentional?

src/components/reservation/ReservationModalBody.tsx
```diff
@@ -36,23 +36,29 @@ import {
...
   mainManifestation: { pid, materialTypes, titles, edition },
-  parallelManifestations
+  parallelManifestations,
+  periodicalSelect
 }: ReservationModalProps) => {
   const mainManifestationType = getManifestationType(mainManifestation);
   const { reservableManifestations } = UseReservableManifestations({
-    manifestations: parallelManifestations ?? [mainManifestation],
+    manifestations:
+      parallelManifestations && parallelManifestations.length > 0
```
spaceo, line 59: The type `ReservationModalProps` says that `ReservationModalProps.parallelManifestations` either is `undefined` or `Manifestation[]`. It should not be necessary to check for: `parallelManifestations.length > 0`

src/components/reservation/ReservationModalBody.tsx
```diff
@@ -134,7 +141,10 @@ const ReservationModalBody = ({
               <div className="reservation-modal-tag">
                 {materialTypes[0].specific}
               </div>
-              <h2 className="text-header-h2 mt-22 mb-8">{titles.main[0]}</h2>
+              <h2 className="text-header-h2 mt-22 mb-8">
+                {titles.main[0]}{" "}
```
spaceo, line 145: You can also destruct it in the function definition:
```
mainManifestation: { pid, materialTypes, titles: { [main: mainTitle] }, edition }
```
Haven't tested so there could be mistakes.
If you are not fond of the destructing I think at least you should create a variable or a function for it since you refer to it twice in this component. It would make sense to create a funcvtion for it since the main title is also used in other places of the codebase.

src/components/reservation/helper.ts
```diff
@@ -54,48 +55,74 @@ export const getFutureDateString = (num: number) => {
...
+  periodical
 }: {
   manifestations: Manifestation[];
   selectedBranch: string | null;
   expiryDate: string | null;
+  periodical: GroupListItem | null | undefined;
 }): CreateReservationBatchV2 => {
+  const selectedPeriodical = periodical
+    ? {
+        volumeNumber: periodical.volumeNumber,
```
spaceo, line 115: Destruct?

kasperbirch1: not possible as periodical can be null

src/components/reservation/ReservationModalBody.tsx
kasperbirch1: I think the empty array is the same as true and therefore it will not select mainManifestation

sinejespersen (approved):

PR #173, opened 2022-09-29
Read article from infomedia

kasperbirch1:
#### This PR makes it possible to read articles from Infomedia.

That has been added a new modal that is opened from MaterialButtonOnlineInfomediaArticle 

It gets the data from useGetInfomediaQuery inside InfomediaModal

#### Screenshot of the result
Skærmbillede 2022-10-03 kl  15 01 06
Skærmbillede 2022-10-03 kl  15 01 11

#### Checklist


src/components/material/infomedia/InfomediaModalBody.tsx
```diff
@@ -0,0 +1,25 @@
+import React from "react";
+import { GetInfomediaQuery } from "../../../core/dbc-gateway/generated/graphql";
+
+export interface InfomediaModalBodyProps {
+  infomediaArticle: GetInfomediaQuery["infomedia"]["article"];
```
kasperg, line 5: Please use general types needed by the modal for props e.g. `title: string` and `text: ` instead of referencing the generated types.

This decouples the component.

Also it seems as if there are a lot of optional values involved in the article. It would be preferable if we ensured that we actually have what resembles an actual article (e.g. `text` must be set, perhaps `headline` is optional) before we consider it something worth rendering.

src/apps/material/material.tsx
```diff
@@ -101,9 +103,9 @@ const Material: React.FC<MaterialProps> = ({ wid }) => {
     t
   });
 
-  if (!currentManifestation) {
-    return null;
-  }
+  const infomediaId = currentManifestation.access.find(
+    (item) => item.__typename === "InfomediaService"
+  ) as InfomediaService;
```
kasperg, line 108: Please rename the variable `infomediaId`. I suggest just calling it `infomedia` or `infomediaAccess`.

The variable does not (only) represent an id. Accessing via `infomediaId.id` (double id) looks odd.

However I do think this is an example where it is OK to use `as`.

src/apps/material/material.dev.tsx
```diff
@@ -526,3 +526,8 @@ export const Periodical = Template.bind({});
 Periodical.args = {
   wid: "work-of:870970-basis:06373674"
 };
+
+export const Infomedia = Template.bind({});
+Infomedia.args = {
+  wid: "work-of:870971-avis:35731733"
+};
```
kasperg, line 533: :+1: Good to add this as an separate variation.

src/components/material/infomedia/InfomediaModalBody.tsx
```diff
@@ -0,0 +1,25 @@
...
+}
+
+const InfomediaModalBody: React.FunctionComponent<InfomediaModalBodyProps> = ({
+  infomediaArticle
+}) => {
+  return (
+    <article>
+      {infomediaArticle?.headLine && <h2>{infomediaArticle?.headLine}</h2>}
```
kasperg, line 13:
```suggestion
      {infomediaArticle?.headLine && <h2>{infomediaArticle.headLine}</h2>}
```

The second null chaining should not be necessary with the first check in mind.

src/components/material/infomedia/InfomediaModalBody.tsx
```diff
@@ -0,0 +1,25 @@
...
+      {infomediaArticle?.text && (
+        <div
+          className=""
```
kasperg, line 16: Please remove `className` if we do not intend to provide any value for it.

kasperg (changes requested):
Some comments, some good stuff 👍 

In general I wonder about the outcome on the provided screenshot:

- The availability button shows the item as unavailable. I wonder why that is. Perhaps it has something to do with a future split/addition of availability buttons based on physical and online materials. The article is not available physically but would be available online.
- There are three button showing "Unreservable", "Out on loan" and "Read article". I had not expected the two first buttons to show. I wonder why they are there.
- The text "We have bought 0 copies" is shown. I had not expected it to be since this is not a physical material.

src/components/material/infomedia/InfomediaModal.tsx
```diff
@@ -0,0 +1,56 @@
...
+  if (!data || error) {
+    return null;
+  }
+
+  const headline = data?.infomedia?.article?.headLine;
+  const text = data?.infomedia?.article?.text;
+
+  return (
+    <Modal
+      modalId={infomediaModalId(convertPostIdToFaustId(pid))}
```
kasperg, line 43: Consider changing `infomediaModalId()` to accept a `pid` instead of a `faust`.

It seems as if everywhere we use the function we convert a pid to a faust. Perhaps we can just let the function accept a pid instead? Then we can either convert it to a faust internally or perhaps work with it be as a pid.

sinejespersen (approved):

PR #174, opened 2022-09-30
Feature/find on shelf periodical dropdowns

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFSOEG-230

#### Description
This PR introduces periodical dropdowns to the find-on-shelf modal. These dropdowns control which edition of a given periodical material the modal shows data for. 

#### Screenshot of the result
https://user-images.githubusercontent.com/28546954/193258470-53ea94aa-de98-4dd6-87a0-b52ee5f1bcdb.mov

#### Checklist


src/components/find-on-shelf/FindOnShelfModal.tsx
```diff
@@ -130,6 +158,16 @@ const FindOnShelfModal: FC<FindOnShelfModalProps> = ({
         <h2 className="text-header-h2 modal-find-on-shelf__headline">
           {`${title} / ${author}`}
         </h2>
+        {manifestations.some((manifestation) => {
+          return manifestation.materialTypes.some((materialType) => {
+            return materialType.specific.includes("periodikum");
+          });
+        }) && (
```
kasperg, line 165: Consider moving this logic out of the template.

See [this](https://github.com/danskernesdigitalebibliotek/dpl-react/pull/172/files#diff-a77066a65eabdc34863f418f9a8581b8c603f5c709b44ac0253d38c4d78617b8R78) for inspiration.

src/components/find-on-shelf/FindOnShelfPeriodicalDropdowns.tsx
```diff
@@ -0,0 +1,81 @@
...
+        <select
+          id="find-on-shelf-periodical-year"
+          className="dropdown__select"
+          aria-label="Choose periodical year"
+          defaultValue={selectedYear}
+          onChange={(e) => setSelectedYear(e.target.value)}
+        >
+          {Object.keys(sortedperiodicalEditionsArray)
+            .sort()
+            .map((item) => (
```
kasperg, line 41: Consider how you can improve the readability of this:

1. We `sort()` something that based on the variable name `sortedperiodicalEditionsArray` already seems to be sorted. You might say that it has not been sorted - it has been grouped.
2. The name `item` says nothing about what the value contains. It is in fact a (volume) year.
3. `Array` can become a superfluous or misleading suffix. Giving the variable a plural name sortedperiodicalEdition*s* indicate that we have many. This will usually be in the from of an array or set. Also in this case the value is in fact not an array. It is an object used as a [map](https://javascript.info/map-set).

One approach would be to move most of this out of the JSX and colocate it with the grouping.

src/components/find-on-shelf/FindOnShelfPeriodicalDropdowns.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+  const [selectedYear, setSelectedYear] = useState<string>(
+    String(Object.keys(sortedperiodicalEditionsArray).sort().pop())
+  );
+
+  return (
+    <div className="modal-find-on-shelf__periodical-dropdowns">
+      <div className="dropdown dropdown--grey-borders">
+        <select
+          id="find-on-shelf-periodical-year"
```
kasperg, line 33: Consider whether this `id` value is needed.

An `id` value must be unique. I am not sure how this code guarantees this. I do not see if/how the id is needed either.

src/components/find-on-shelf/FindOnShelfPeriodicalDropdowns.tsx
```diff
@@ -0,0 +1,81 @@
...
+          className="dropdown__select"
+          aria-label="Choose periodical year"
```
kasperg, line 35: Please remember to make the label translatable..

src/components/find-on-shelf/FindOnShelfPeriodicalDropdowns.tsx
```diff
@@ -0,0 +1,81 @@
...
+        <div className="dropdown__arrows">
+          <img className="dropdown__arrow" src={ExpandMoreIcon} alt="" />
+        </div>
+      </div>
+      {selectedYear && (
+        <div className="dropdown dropdown--grey-borders">
+          <select
+            id="find-on-shelf-periodical-edition"
+            className="dropdown__select"
+            aria-label="Choose periodical week"
```
kasperg, line 56: See comments above.

src/components/find-on-shelf/FindOnShelfPeriodicalDropdowns.tsx
```diff
@@ -0,0 +1,81 @@
+import * as React from "react";
+import { FC, useState } from "react";
+import ExpandMoreIcon from "@danskernesdigitalebibliotek/dpl-design-system/build/icons/collection/ExpandMore.svg";
+import { HoldingsForBibliographicalRecordV3 } from "../../core/fbs/model";
+import makePeriodicalEditionsFromHoldings from "../material/helper";
+import { groupObjectArrayByProperty } from "../../core/utils/helpers/general";
+
+export interface FindOnShelfPeriodicalDropdownProps {
+  manifestationsHoldings: HoldingsForBibliographicalRecordV3[];
+  setSelectedPeriodical: (selectedWeek: string | null) => void;
```
kasperg, line 10: Consider updating the signature here:

1. The value selected is probably not the week.
2. Do we need to allow null to be passed?

src/components/material/MaterialHeader.tsx
```diff
@@ -89,8 +89,11 @@ const MaterialHeader: React.FC<MaterialHeaderProps> = ({
             selectManifestationHandler={selectManifestationHandler}
           />
         </div>
-
-        {manifestation?.source?.includes("bibliotekskatalog") && (
+        {manifestation?.materialTypes.some(
+          (materialType: Manifestation["materialTypes"][0]) => {
+            return materialType.specific.includes("periodikum");
+          }
```
kasperg, line 95: Please refactor this into a shared function as mentioned above.

kasperg (changes requested):
:+1: Good work but I think there are a few places where the code needs to be tuned a bit.

src/apps/material/material.dev.tsx
```diff
@@ -19,7 +19,12 @@ export default {
     wid: {
       name: "Work ID",
       defaultValue: "work-of:870970-basis:52557240",
-      control: { type: "text" }
+      options: [
+        "work-of:870970-basis:52557240",
+        "work-of:870970-basis:52077478",
+        "work-of:870970-basis:06373674"
+      ],
+      control: { type: "select" }
```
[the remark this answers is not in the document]
Adamik10: Fair enough :) Thanks for the explanation. It makes complete sense ;)
I am reverting these changes.

src/components/find-on-shelf/FindOnShelfModal.tsx
```diff
@@ -82,6 +86,32 @@ const FindOnShelfModal: FC<FindOnShelfModalProps> = ({
...
+          return {
+            ...manifestationHoldings,
+            holding: {
+              ...manifestationHoldings.holding,
+              materials: manifestationHoldings.holding.materials.filter(
+                (material) => {
+                  return (
+                    material.periodical?.volumeNumber ===
+                    selectedPeriodicalItemNumber
+                  );
```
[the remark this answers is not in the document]
Adamik10: Oh, of course they aren't unique. Great eyes and sharp brains! I am now checking whether both the year and volume(week) match.

src/components/find-on-shelf/FindOnShelfModal.tsx
Adamik10: Great for improving readability of the HTML part of the component. Thanks!

src/components/find-on-shelf/FindOnShelfPeriodicalDropdowns.tsx
Adamik10: Yeah, I'm not sure if I used enough time to fully understand this dropdown code before I reused it in find on shelf.
The ids are not needed as far as I can see, and I removed them now.

src/components/find-on-shelf/FindOnShelfPeriodicalDropdowns.tsx
Adamik10: Very very good comment. Thanks @kasperg :)

kasperg (approved):
:+1: LGTM

sinejespersen (approved):

PR #175, opened 2022-09-30
Feature/5379 refactor before reservation material

sinejespersen:
#### Link to issue

#### Description

This is a clean up to make (stackable)material-component sharable. 
I have: 
- moved the stackable-component-file into a folder called "stackable-component"
- created additional-materials-button, that shows if the component is stacked (and reservations are not stacked so making this a component makes it easy to _not use_ 
- created material-info - the left side og the material, which has type/year/dadada and will be used in reservations
- created material-overdue-link which is a link that shows if the material is overdue 
- created material-status, which is the right side, that is only used by loan-materials
- add config fbs/publizon to reservation list

#### Screenshot of the result

n/a

#### Checklist


src/apps/loan-list/materials/stackable-material/material-info.tsx
```diff
@@ -0,0 +1,58 @@
...
+}) => {
+  const t = useText();
+  const { creators, materialType, year, materialTitle, pid, description } =
+    getMaterialInfo(loanMetaData, material);
+
+  return (
+    <div className="list-reservation__material">
+      <div>
+        <Cover
+          pid={pid as Pid}
```
spaceo, line 28: I think the `pid` coming from `getMaterialInfo()` should be of the `Pid` type that we have globally.
Then you would not have to tranform it like this-

src/apps/loan-list/materials/stackable-material/material-info.tsx
```diff
@@ -0,0 +1,58 @@
...
+        <div>
+          <div className="status-label status-label--outline">
+            {materialType}
+          </div>
+        </div>
+        <div className="list-reservation__about">
+          <h3 className="text-header-h4">{materialTitle}</h3>
+          <p className="text-small-caption color-secondary-gray">
+            {creators &&
+              getAuthorNames(
```
spaceo, line 44: This is just a note. I did not look into the format of getAuthorNames. But my feeling is that we have several ways of doing this in the project now.

src/apps/loan-list/materials/stackable-material/material-info.tsx
```diff
@@ -0,0 +1,58 @@
+import React, { FC, ReactNode } from "react";
```
spaceo, line 1: I should have told you earlier on. But we decided some time ago to use Pascal case in our component naming.
Which leaves us with a mixed naming setup, but that could be cleaned up in the future.
I know it is not optimal.

src/apps/loan-list/materials/stackable-material/stackable-material.tsx
```diff
@@ -0,0 +1,119 @@
...
+}) => {
+  const t = useText();
+  const { open } = useModalButtonHandler();
+  const [additionalMaterials] = useState(
+    amountOfMaterialsWithDueDate ? amountOfMaterialsWithDueDate - 1 : 0
+  );
+  const [showModal, setShowModal] = useState(false);
+  const { dueDate, id } = loanMetaData;
+
+  function stopPropagationFunction(e: Event | MouseEvent) {
```
spaceo, line 32: I think this should be global so we could use it everywhere

src/apps/loan-list/materials/stackable-material/stackable-material.tsx
```diff
@@ -0,0 +1,119 @@
...
+    if (openModal) {
+      openDueDateModal();
+    }
+  }, [openDueDateModal, openModal]);
+
+  const selectListMaterial = useCallback(
+    (e: MouseEvent) => {
+      stopPropagationFunction(e);
+      if (selectMaterial) {
+        selectMaterial({
```
spaceo, line 64: Just a suggestion:
I usually name callbacks: "*Handler" to emphasize the purpose of the prop.
So in this case it would be: "selectMaterialHandler"

src/apps/loan-list/utils/helpers.ts
```diff
@@ -136,4 +137,40 @@ export const mapFBSRenewedLoanToLoanMetaDataType = (
   });
 };
 
+export const getMaterialInfo = (
```
spaceo, line 140: I like the idea of this function. That you can make sure that the object is constructed as you want.
If you could add types for the values it would be an extra benefit.
So you wouldn't have to use the `as Pid` construction.

spaceo (commented):
Approved, but please look at my comments ;)

src/apps/loan-list/materials/stackable-material/material-info.tsx
sinejespersen: I am thinking this will be changed after the string interpolation fix

src/apps/loan-list/materials/stackable-material/stackable-material.tsx
sinejespersen: can move

PR #184, opened 2022-10-05
Dpl react interagering med dpl cms proxy endpoint ddfsoeg 236

spaceo:
#### Link to issue

#### Description

#### Screenshot of the result

#### Checklist


orval.config.ts
```diff
@@ -98,5 +98,31 @@ export default defineConfig({
         indent: 2
       }
     }
+  },
+  dplCmsApi: {
+    output: {
+      mode: "split",
+      target: "src/core/dpl-cms-api/dpl.ts",
```
kasperg, line 105: Consider if this should be `cms` or `dpl-cms` instead of `dpl`.

It is the CMS we are interacting with not DPL :smile:

src/core/dpl-cms-api/dpl-cms-api.example.json
```diff
@@ -0,0 +1,83 @@
+{
+  "swagger": "2.0",
+  "schemes": ["http"],
+  "info": {
+    "description": "The REST API provide by the core REST module.",
+    "title": "DPL CMS - REST API",
+    "version": "Versioning not supported"
```
kasperg, line 7: Consider if this should be included in VCS.

We use the external url from the DPL CMS repository in our Orval configuration.

src/core/utils/helpers/general.ts
```diff
@@ -102,7 +102,7 @@ export const usePrevious = <Type>(value: Type) => {
 };
 
 export const convertPostIdToFaustId = (postId: Pid) => {
-  const matches = postId.match(/^[0-9]+-[a-z]+:([0-9]+)$/);
+  const matches = postId.match(/^[0-9]+-[a-z]+:([a-zA-Z0-9]+)$/);
```
kasperg, line 105: Consider adding a TODO here explaining that there is thoughtwork here in progress.

src/core/dpl-cms-api/useProxyUrlGET.tsx
```diff
@@ -0,0 +1,49 @@
...
+  UseQueryOptions,
+  QueryFunction,
+  UseQueryResult,
+  QueryKey
+} from "react-query";
+import { getProxyUrlGETQueryKey, proxyUrlGET } from "./dpl";
+import type { ProxyUrlGETParams } from "./model";
+import { ErrorType } from "./mutator/fetcher";
+
+const useProxyUrlGET = <
```
kasperg, line 12: Consider if this warrants some documentation.

So this is our own alternative implementation of `useProxyUrlGET`. The reason why we need it here (support for `enabled`?) is not clear and others may assume that they can just use the generated hooks out of the box.

It would be nice with some explanation of why it is needed. If we cannot use the React Query hooks at all for DPL CMS then we should consider if/how it can be disabled for this API in Orval.

kasperg (approved):
:+1: A few things for you to look into. Nothing blocking.

orval.config.ts
```diff
@@ -98,5 +98,31 @@ export default defineConfig({
...
+        query: {
+          useQuery: true
+        }
+      },
+      prettier: true
+    },
+    input: {
+      // TODO: Needs to be set to HEAD version when ready.
+      target:
+        "https://raw.githubusercontent.com/danskernesdigitalebibliotek/dpl-cms/e7a65cef84ab1fcfed53f03e039a5605b91941b5/openapi.json",
```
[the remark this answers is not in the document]
spaceo: 👍

orval.config.ts
spaceo: 👍

src/core/dpl-cms-api/dpl-cms-api.example.json
spaceo: 👍

src/core/utils/helpers/general.ts
spaceo: 👍

src/core/dpl-cms-api/useProxyUrlGET.tsx
spaceo: After long investigation I found a way to avoid replicating the hook.
The `requestOptions: false` setting in the orval configuration prevents the `enabled` option to be overwritten.

sinejespersen (approved):

PR #186, opened 2022-10-06
Feature/5379 create reservation type

sinejespersen:
#### Link to issue

#### Description

#### Screenshot of the result

#### Checklist

### DIFF
--- a/src/apps/loan-list/list/loan-list.dev.tsx
+++ b/src/apps/loan-list/list/loan-list.dev.tsx
@@ -261,6 +261,26 @@ export default {
     loanListDigitalPhysicalLoansEmptyListText: {
       defaultValue: "Du har i øjeblikket 0 lån",
       control: { type: "text" }
+    },
+    publizonEbookText: {
+      defaultValue: "E-bog",
+      control: { type: "text" }
+    },
+    publizonAudioBookText: {
+      defaultValue: "Lydbog",
+      control: { type: "text" }
+    },
+    publizonPodcastText: {
+      defaultValue: "Podcast",
+      control: { type: "text" }
+    },
+    materialByAuthorText: {
+      defaultValue: "Af",
+      control: { type: "text" }
+    },
+    materialAndAuthorText: {
+      defaultValue: "og",
+      control: { type: "text" }
     }
   },
   decorators: [withQuery]
--- a/src/apps/loan-list/list/loan-list.entry.tsx
+++ b/src/apps/loan-list/list/loan-list.entry.tsx
@@ -67,6 +67,11 @@ export interface LoanListEntryProps {
   loanListDigitalLoansEmptyListText: string;
   loanListPhysicalLoansEmptyListText: string;
   loanListDigitalPhysicalLoansEmptyListText: string;
+  materialAndAuthorText: string;
+  materialByAuthorText: string;
+  publizonAudioBookText: string;
+  publizonPodcastText: string;
+  publizonEbookText: string;
 }
 
 const LoanListEntry: FC<LoanListEntryProps> = () => <LoanList />;
--- a/src/apps/loan-list/materials/utils/checkbox.tsx
+++ b/src/apps/loan-list/materials/utils/checkbox.tsx
@@ -8,7 +8,7 @@ interface CheckBoxProps {
   hideLabel?: boolean;
   selected?: boolean;
   disabled?: boolean;
-  onChecked: (faust: FaustId) => void;
+  onChecked?: (faust: FaustId) => void;
 }
 
 const CheckBox: FC<CheckBoxProps> = ({
@@ -24,7 +24,7 @@ const CheckBox: FC<CheckBoxProps> = ({
       <input
         id={id}
         className="checkbox__input"
-        onChange={() => onChecked(id as FaustId)}
+        onChange={() => onChecked && onChecked(id as FaustId)}
         checked={selected}
         type="checkbox"
         disabled={disabled}
--- a/src/apps/loan-list/materials/utils/material-fetch-hoc.tsx
+++ b/src/apps/loan-list/materials/utils/material-fetch-hoc.tsx
@@ -1,82 +1,68 @@
-import React, { useEffect, useState, ComponentType } from "react";
-import {
-  GetMaterialManifestationQuery,
-  useGetMaterialManifestationQuery
-} from "../../../../core/dbc-gateway/generated/graphql";
-import { LoanMetaDataType } from "../../../../core/utils/types/loan-meta-data-type";
+import React, { useEffect, useState, ComponentType, FC } from "react";
+import { useGetMaterialManifestationQuery } from "../../../../core/dbc-gateway/generated/graphql";
+import { Product } from "../../../../core/publizon/model";
 import { FaustId } from "../../../../core/utils/types/ids";
-
-export interface MaterialDetailsProps {
-  loanMetaData: LoanMetaDataType;
-}
-
-export interface SelectableMaterialProps {
-  loanMetaData: LoanMetaDataType;
-  disabled?: boolean;
-  materialsToRenew?: number[];
-  onChecked?: (faust: FaustId) => void;
-}
-
-export interface StackableMaterialProps {
-  stack?: LoanMetaDataType[];
-  loanMetaData: LoanMetaDataType;
-  amountOfMaterialsWithDueDate?: number;
-  dueDateLabel?: string;
-  openModal?: boolean;
-  selectMaterial?: ({
-    material,
-    loanMetaData
-  }: {
-    material: GetMaterialManifestationQuery | undefined | null;
-    loanMetaData: LoanMetaDataType;
-  }) => void;
-}
+import { BasicDetailsType } from "../../../../core/utils/types/basic-details-type";
+import { mapManifestationToBasicDetailsType } from "../../../../core/utils/helpers/list-mapper";
 
 export interface MaterialProps {
-  material: GetMaterialManifestationQuery;
+  material?: BasicDetailsType | null;
 }
 
-type InputProps = { loanMetaData: LoanMetaDataType } & (
-  | StackableMaterialProps
-  | SelectableMaterialProps
-  | MaterialDetailsProps
-);
-export type WithMaterialProps = MaterialProps &
-  (SelectableMaterialProps | StackableMaterialProps | MaterialDetailsProps);
+type InputProps = {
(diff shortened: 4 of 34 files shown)


src/apps/loan-list/materials/utils/checkbox.tsx
```diff
@@ -24,7 +24,7 @@ const CheckBox: FC<CheckBoxProps> = ({
       <input
         id={id}
         className="checkbox__input"
-        onChange={() => onChecked(id as FaustId)}
+        onChange={() => onChecked && onChecked(id as FaustId)}
```
spaceo, line 27: Is it not possible to declare the id type in the props as a `FaustId`?

Side note:
If you choose to move the checkbox out as a global component, you would have to make use a generic for the id type, since the id would vary between several types.
Generics and components: https://javascript.plainenglish.io/react-components-as-typescript-generic-functions-8aa83afff597

src/apps/loan-list/materials/utils/material-fetch-hoc.tsx
```diff
@@ -1,82 +1,68 @@
...
-type InputProps = { loanMetaData: LoanMetaDataType } & (
-  | StackableMaterialProps
-  | SelectableMaterialProps
-  | MaterialDetailsProps
-);
-export type WithMaterialProps = MaterialProps &
-  (SelectableMaterialProps | StackableMaterialProps | MaterialDetailsProps);
+      // Todo error handling
+      const { isSuccess: isSuccessManifestation, data } =
+        useGetMaterialManifestationQuery({
```
spaceo, line 40: This hook name is too generic. We could easily have other cases where we would need to fetch a Material Manifestation but with other fields. I think it should be renamed by changing the query name.

src/core/utils/helpers/list-mapper.ts
```diff
@@ -0,0 +1,161 @@
...
+import { FaustId } from "../types/ids";
+import { store } from "../../store";
+import { GetMaterialManifestationQuery } from "../../dbc-gateway/generated/graphql";
+import { BasicDetailsType } from "../types/basic-details-type";
+import { Product, Loan } from "../../publizon/model";
+import { LoanType } from "../types/loan-type";
+
+// Creates a "by author, author and author"-string
+// String interpolation todo?
+export const getContributors = (creators: string[] | undefined) => {
```
spaceo, line 11: Why can creators be undefined? It is maybe a matter of taste but I think it is better to keep it strictly to `creators: string[]` and make sure in the code that references `getContributors(creators)` check if `creators` is not undefined.

spaceo: Another thing: We have a LOT of places where we are showing a string with authors/contributors. Can this be used in all those places?

src/core/utils/helpers/list-mapper.ts
```diff
@@ -0,0 +1,161 @@
...
+      returnContentString = `${texts.materialByAuthorText} ${creators
+        .slice(0, -1)
+        .join(", ")} ${texts.materialAndAuthorText} ${creators.slice(-1)}`;
+    }
+  }
+  return returnContentString;
+};
+
+//
+function getYearFromDataString(date?: string) {
```
spaceo, line 32: This is the same case as above why making date optional when the whole functionality depends on the argument?

src/core/utils/helpers/list-mapper.ts
```diff
@@ -0,0 +1,161 @@
...
+    publicationDate,
+    title,
+    description,
+    productType,
+    contributors,
+    externalProductId
+  } = material;
+
+  const {
+    text: { data: texts }
```
spaceo, line 113: I think we should avoid using the store directly if we can. We have one system for handling text strings: useText. In similar functions to this we have simply added a parameter to the function called: `t`. See: https://github.com/reload/dpl-react/blob/storybook-debug/src/core/utils/helpers/general.ts#L50:L50

src/core/utils/types/basic-details-type.ts
```diff
@@ -0,0 +1,11 @@
+import { Pid } from "./ids";
+
+export interface BasicDetailsType {
```
spaceo, line 3: If all the props can be optional (undefined) I think you should wrap the whole interface in `Partial`:
https://www.typescriptlang.org/docs/handbook/utility-types.html#partialtype

src/core/utils/types/loan-type.ts
```diff
@@ -1,7 +1,8 @@
-export interface LoanMetaDataType {
+import { ListType } from "./list-type";
+
+export interface LoanType extends ListType {
   dueDate: string | null | undefined;
   loanDate: string | null | undefined;
-  id: string;
   isRenewable: boolean;
   materialItemNumber: string | null | undefined;
```
spaceo, line 7: All the propeties that can be undefined you can instead of this:
`dueDate: string | null | undefined;`
specify it like this:
`dueDate?: string | null;`

src/core/utils/types/reservation-type.ts
```diff
@@ -0,0 +1,11 @@
+import { ListType } from "./list-type";
+
+export interface ReservationType extends ListType {
+  dateOfReservation: string | undefined;
+  expiryDate: string | null | undefined;
+  pickupDeadline: string | null | undefined;
+  numberInQueue?: number | undefined;
```
spaceo, line 7: See above comment

spaceo (changes requested):
A lot of code, therefore several remarks.

src/apps/loan-list/list/loan-list.entry.tsx
```diff
@@ -67,6 +67,11 @@ export interface LoanListEntryProps {
   loanListDigitalLoansEmptyListText: string;
   loanListPhysicalLoansEmptyListText: string;
   loanListDigitalPhysicalLoansEmptyListText: string;
+  materialAndAuthorText: string;
+  materialByAuthorText: string;
+  publizonAudioBookText: string;
+  publizonPodcastText: string;
+  publizonEbookText: string;
```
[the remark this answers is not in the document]
sinejespersen: I guess I can change it. But this is a temporary solution until the string interpolation stuff works.

src/apps/loan-list/materials/utils/material-fetch-hoc.tsx
sinejespersen: I am adding some fields to it in a later pr, so perhaps I can rename it in that?

src/core/utils/helpers/list-mapper.ts
sinejespersen: I don't know. Probably if the input fits, but there are some strings that it is using ("and","or"). But I think maybe this will be changed when the string interpolation stuff is created?

src/core/utils/helpers/list-mapper.ts
sinejespersen: Imma change both

src/core/utils/helpers/list-mapper.ts
sinejespersen: I've changed it. I used the store like that, because I saw it in a pr and assumed it was fine. But sending the hook with is fine as well, I just wasnt aware.

src/core/utils/types/basic-details-type.ts
sinejespersen: I am not sure I get this.....

src/core/utils/types/loan-type.ts
sinejespersen: Changed

src/apps/loan-list/materials/utils/material-fetch-hoc.tsx
spaceo: 👍

src/core/utils/helpers/list-mapper.ts
spaceo: 👍

src/core/utils/types/basic-details-type.ts
spaceo: Either like [this](https://www.typescriptlang.org/play?ssl=10&ssc=1&pln=11&pc=1#code/PTAECgFMA8AcHsBOAXUyCetKgGIEMBXAZ2QEkATUAXlAAMASAbwDsCBbAI0kQF9aBuKHCSoMWUAAUAlpRoMW7LrwC0TEoinMA5jwBcTfMTLk+g8GOwA5AgBsbeDjcgAeACoA+aqEagA2gGlQTVAAa0h0eAAzUFcAXV0YgNjQAB9QVjtQHjNNZG5IvABjbAAhPCIpQoARSGQ8KRsiV0xIMqJsRnBQUEJkAAskIgT1TS1BbtgZBOlycdAYPMRmPBsJRHhyAkLjYeQNbTm2PEWpFeasXf2xrtBySCJCjVhkKXhmS9G59Eg8RA+Dm4vZBOf7XHjgcwtUBtSo1OoNJpQmjWOwOJzOCS-F4rZww6q1eqNc6tcqQdzuQRAA) or like [this](https://www.typescriptlang.org/play?#code/PTAECgFMA8AcHsBOAXUyCetKgGIEMBXAZ2QEkATUAXlAAMASAbwDsCBbAI0kQF9aBuKHCSoMWUAAUAlpRoMW7LrwC0TEoinMA5jwBcTfMTLk+g8GOwB5WMinxmeADYBBZuQByBR47wdHkAB4AFQA+alBGUABtAGlQTVAAa0h0eAAzUCCAXQB+XUzYrNAAH1BWb1AeM01kbjS8AGNsACE8IikGgBFIZDwpRyIgzEhWomxGcFBQQmQACyQifPVNLUEp2Bl86XI10BhaxAdHCUR4cgIG4yXkDW1dtjwDqSchrGvb1cnQckgiBo0bHZmO8Vrt0JA8IgQXcvrZkP5oZ8eOBzMNQKMOt1ev1BmiaNZbPYnK4PF4fH5Ahiuj0+gNXiM2pAQoIgA)

src/apps/loan-list/list/loan-list.dev.tsx
```diff
@@ -117,31 +117,6 @@ export default {
...
-      control: { type: "text" }
-    },
-    loanModalMaterialAndAuthorText: {
-      defaultValue: "og",
-      control: { type: "text" }
-    },
-    materialDetailsAndAuthorText: {
-      defaultValue: "og",
-      control: { type: "text" }
-    },
```
spaceo, line 144: Why are you deleting arguments?

PR #188, opened 2022-10-07
Blacklist Pickup Branches

kasperbirch1:
#### Remove blacklist branches from pickup select in the reservation modal 

- We no longer fetch branches with useGetBranches, instead, we get them as config props from storybook / CMS
- new function excludeBlacklistedBranches, which sorts out the blacklisted IDs
- UseConfigFunction has been changed to be able to turn a comma-separated string into an array or JSON.parse a string

#### Screenshot of the result
Skærmbillede 2022-10-10 kl  11 45 15

#### Checklist


src/apps/material/material.dev.tsx
```diff
@@ -26,6 +26,16 @@ export default {
...
+    },
```
kasperg, line 38: Consider whether this control should have a defined type as we do for other controls.

src/components/reservation/ReservationModalBody.tsx
```diff
@@ -58,6 +58,14 @@ const ReservationModalBody = ({
   parallelManifestations,
   selectedPeriodical
 }: ReservationModalProps) => {
+  const t = useText();
+  const config = useConfig();
+  const branches = config<AgencyBranch>("branchesConfig", {
+    jsonParse: true
+  });
+
+  const userListItemsBranches = Array.isArray(branches) ? branches : [];
```
kasperg, line 67: Consider if you can provide a more explanatory name.

I do not understand why it is called `userListItemsBranches`.

src/core/utils/config.tsx
```diff
@@ -2,11 +2,26 @@ import { RootState, useSelector } from "../store";
 import { addConfigEntries } from "../config.slice";
 import withSuffix from "./withSuffix";
 
-export type UseConfigFunction = (key: string) => unknown;
+export type UseConfigFunction = <T>(
+  key: string,
+  options?: { jsonParse?: boolean; stringToArray?: boolean }
+) => string | T[] | undefined;
```
kasperg, line 8: Consider if we can provide a more descriptive type for the return value.

It seems as if the method will return a string if options are not defined. It will always return an array of strings if `stringToArray` is set. It will return an array of `T` if `jsonParse` is defined. Also `jsonParse` and `stringToArray` cannot be set at the same time.

src/apps/material/material.dev.tsx
```diff
@@ -26,6 +26,16 @@ export default {
       defaultValue: "1",
       control: { type: "text" }
     },
+    blacklistedBranchesConfig: {
+      name: "Blacklisted branches",
+      defaultValue: "FBS-751032,FBS-751031,FBS-751009,FBS-751027,FBS-751024",
+      control: { type: "text" }
+    },
```
kasperg, line 33: Consider if it is worth specifying that this blacklist is for reservation pickup locations.

There will be other blacklists in future PRs. If future PRs rename the entry you can disregard this.

kasperg (approved):
:+1: I have some considerations for you to consider before you merge.

sinejespersen (approved):

PR #190, opened 2022-10-10
Exclude blacklistBranches in AvailabilityLabel + FindOnShelfModal

kasperbirch1:
#### This PR Exclude blacklistBranches in  AvailabilityLabel + FindOnShelfModal
Because blacklisted branches must not be included in the AvailabilityLabel or FindOnShelfModal accessibility check

#### Checklist

### DIFF
--- a/src/apps/material/material.dev.tsx
+++ b/src/apps/material/material.dev.tsx
@@ -32,7 +32,12 @@ export default {
       control: { type: "text" }
     },
     blacklistedPickupBranchesConfig: {
-      name: "Blacklisted branches",
+      name: "Blacklisted Pickup branches",
+      defaultValue: "FBS-751032,FBS-751031,FBS-751009,FBS-751027,FBS-751024",
+      control: { type: "text" }
+    },
+    blacklistedAvailabilityBranchesConfig: {
+      name: "Blacklisted Availability branches",
       defaultValue: "FBS-751032,FBS-751031,FBS-751009,FBS-751027,FBS-751024",
       control: { type: "text" }
     },
--- a/src/components/availability-label/availability-label.tsx
+++ b/src/components/availability-label/availability-label.tsx
@@ -24,12 +24,9 @@ export const AvailabilityLabel: React.FC<AvailabilityLabelProps> = ({
   cursorPointer = false
 }) => {
   const config = useConfig();
-  const blacklistBranches = config<string>(
-    "blacklistedAvailabilityBranchesConfig",
-    {
-      transformer: "stringToArray"
-    }
-  );
+  const blacklistBranches = config("blacklistedAvailabilityBranchesConfig", {
+    transformer: "stringToArray"
+  });
 
   const t = useText();
   const { data, isLoading, isError } = useGetAvailabilityV3({
--- a/src/components/availability-label/availability-label.tsx
+++ b/src/components/availability-label/availability-label.tsx
@@ -24,9 +24,12 @@ export const AvailabilityLabel: React.FC<AvailabilityLabelProps> = ({
   cursorPointer = false
 }) => {
   const config = useConfig();
-  const blacklistBranches = config<string>("blacklistedPickupBranchesConfig", {
-    transformer: "stringToArray"
-  });
+  const blacklistBranches = config<string>(
+    "blacklistedAvailabilityBranchesConfig",
+    {
+      transformer: "stringToArray"
+    }
+  );
 
   const t = useText();
   const { data, isLoading, isError } = useGetAvailabilityV3({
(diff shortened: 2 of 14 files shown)


src/components/reservation/UserListItems.tsx
```diff
@@ -47,8 +48,19 @@ const UserListItems: FC<UserListItemsProps> = ({
...
+    stringToArray: true
+  });
+
+  let pickupModalBranches = branches;
+  if (Array.isArray(blacklistBranches)) {
+    pickupModalBranches = excludeBlacklistedBranches(
+      branches,
+      blacklistBranches
+    );
+  }
```
spaceo, line 62: Could this go into a function in the helper called: `getPickupModalBranches()`?

src/components/find-on-shelf/FindOnShelfModal.tsx
```diff
@@ -41,13 +42,18 @@ const FindOnShelfModal: FC<FindOnShelfModalProps> = ({
...
+  });
   const t = useText();
   const pidArray = getManifestationsPids(manifestations);
   const faustIdArray = pidArray.map((manifestationPid) =>
     convertPostIdToFaustId(manifestationPid)
   );
   const { data, isError, isLoading } = useGetHoldingsV3({
-    recordid: faustIdArray
+    recordid: faustIdArray,
+    ...(Array.isArray(blacklistBranches) ? { exclude: blacklistBranches } : {})
```
spaceo, line 56: simple and concise 👍

spaceo (commented):
Looks good!
Have one sugesstion...

src/apps/material/material.dev.tsx
```diff
@@ -26,6 +26,16 @@ export default {
       defaultValue: "1",
       control: { type: "text" }
     },
+    blacklistedBranchesConfig: {
+      name: "Blacklisted branches",
+      defaultValue: "FBS-751032,FBS-751031,FBS-751009,FBS-751027,FBS-751024",
+      control: { type: "text" }
+    },
```
kasperg, line 33: Please add seperate configuration for pickup and availability branches.

There are actually 3 different configurations for the different parts of the system that needs to be blacklisted.

kasperg (changes requested):

src/components/availability-label/availability-label.tsx
```diff
@@ -13,79 +14,90 @@ export interface AvailabilityLabelProps {
...
-  const { data, isLoading, isError } = useGetAvailabilityV3({
-    recordid: faustIds
-  });
+    const t = useText();
+    const { data, isLoading, isError } = useGetAvailabilityV3({
+      recordid: faustIds,
+      ...(Array.isArray(blacklistBranches)
+        ? { exclude: blacklistBranches }
+        : {})
+    });
```
kasperg, line 37: With the latest changes to `useConfig()`, `blacklistBranches` should always return an array of strings so either there is no need for this `Array.isArray()` check or the check should be different - e.g. checking for the length if we only want to pass the `exclude` parameter if there are any blacklisted branches configured.

src/components/availability-label/availability-label.tsx
```diff
@@ -13,79 +14,90 @@ export interface AvailabilityLabelProps {
   handleSelectManifestation?: () => void | undefined;
   cursorPointer?: boolean;
 }
+export const AvailabilityLabel: React.FC<AvailabilityLabelProps> = withConfig(
```
kasperg, line 17: 🤔 I am still no sure whether this is the right place to add this. Have you tried updating the story to `component: withConfig(AvailabilityLabel)` instead?

If we cannot figure this out then perhaps we could leave this as is, merge it and talk to Mikkel about it on monday.

src/components/material/periodical/MaterialPeriodical.tsx
```diff
@@ -18,8 +19,13 @@ const MaterialPeriodical: FC<MaterialPeriodicalProps> = ({
...
   selectPeriodicalHandler
 }) => {
+  const config = useConfig();
+  const blacklistBranches = config("blacklistedPickupBranchesConfig", {
+    transformer: "stringToArray"
+  });
   const { data, isLoading, isError } = useGetHoldingsV3({
-    recordid: [String(faustId)]
+    recordid: [String(faustId)],
+    ...(Array.isArray(blacklistBranches) ? { exclude: blacklistBranches } : {})
```
kasperg, line 28: Note the superfluous `isArray()` check here.

kasperg (changes requested):
There are a few unnecessary array checks and then we have the continuous `withConfig()` saga for the tests.

Otherwise LGTM.

sinejespersen (approved):

PR #191, opened 2022-10-11
Blacklist search result

kasperbirch1:
#### This PR makes it possible to only get results from whitelisted branches in search results. 

This whitelist is created with the following steps
- get branches and blacklisted search Branches from useConfig
- use getWhitelistBranches to get a whitelist
- use cleanBranchesId to extract only the numbers from id's 

The whitelist is sent with the search query as filters { branchId: **whitelist**} 

#### Checklist


src/core/utils/config.tsx
```diff
@@ -2,11 +2,30 @@ import { RootState, useSelector } from "../store";
...
 
-export type UseConfigFunction = (key: string) => unknown;
+type ReturnValue<T> = T extends string ? string : T[];
+
+type UseConfigFunction = <T>(
+  key: string,
+  options?: {
+    transformer?: "jsonParse" | "stringToArray";
+  }
+) => ReturnValue<T>;
```
spaceo, line 12: I think you should just define the conditional return type directly since  it is not used anywhere else:
```suggestion
) =>  T extends string ? string : T[];
```

src/components/reservation/helper.ts
```diff
@@ -152,4 +152,19 @@ export const excludeBlacklistedBranches = (
...
 
+export const cleanBranchesId = (branches: AgencyBranch[]): string[] => {
+  return (
+    branches
+      .map((branch) => {
+        // Filtering on branchId, only uses agency number for example "775100" and not ISIL "DK-775100"
+        // So we need to filter on the digits after the -
+        const pattern = /-(\d*)/g;
+        const matches = pattern.exec(branch.branchId);
+        return matches ? matches[1] : "";
```
kasperg, line 163: :+1: for commenting on why this is needed.

kasperg (approved):
:+1: LGTM

sinejespersen (approved):

PR #193, opened 2022-10-12
Feature/5379 reservation materials component

sinejespersen:
#### Link to issue

#### Description

- extracted StatusCircleIcon into own file, the round thing/counter in the materials
- added getColors and getThresholds (gets config through the configuration loader)
- added a bunch of interface texts (which will be changed later)
- added reservation-info component, which displays the info only related to reservations
- added a reservation helper

#### Screenshot of the result

image
image

#### Checklist


src/apps/reservation-list/reservation-material/reservation-material.tsx
```diff
@@ -0,0 +1,60 @@
...
+    document
+      .querySelector(".list-reservation a")
+      ?.addEventListener("click", stopPropagationFunction, true);
+
+    return () => {
+      document
+        .querySelector(".list-reservation a")
+        ?.removeEventListener("click", stopPropagationFunction, true);
+    };
+  }, []);
```
kasperg, line 32: Could you please elaborate what is going here?

A component reaching outside its scope through `document.querySelector()` does not seem like a good idea. As I see it we will end up with many event listeners on each element. I wonder what has lead you here.

If there are links within the child component that we want to override could be pass an onClick function to the child components instead and let them handle things?

src/apps/reservation-list/reservation-material/reservation-material.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+  const openDetailsModal = useCallback((e: MouseEvent) => {
+    e.stopPropagation();
+    // Todo
+  }, []);
+
+  return (
+    <div>
```
kasperg, line 40: Do we really need the empty wrapper `<div>` here?

src/core/utils/types/reservation-type.ts
```diff
@@ -8,4 +8,5 @@ export interface ReservationType extends ListType {
   state: string;
   pickupBranch?: string;
   expectedRedeemDateUtc?: string;
+  pickupNumber?: string;
```
kasperg, line 11: Could we make `pickupNumber` a number instead of a string? Based on the naming that seems like the right type for the value.

src/core/utils/helpers/list-mapper.ts
```diff
@@ -169,4 +166,72 @@ export const mapManifestationToBasicDetailsType = (
...
+        4: "cancelled", // in publizon Cancelled
+        5: "expired" // in publizon Expired
+      };
+
+      return {
+        identifier,
+        faust: null,
+        dateOfReservation: createdDateUtc,
+        expiryDate: expireDateUtc,
+        state: status ? publizonReservationState[status] : "",
```
kasperg, line 197: Consider if we can make the state type more explicit instead of just allowing a string.

It seems to use a fixed set of specified strings anyway.

src/core/utils/helpers/general.ts
```diff
@@ -93,6 +94,21 @@ export const getCoverTint = (index: number) => {
...
+
+export const getThresholds = () => {
+  return getConf("thresholds", configuration);
+};
+
+export const daysBetweenTodayAndDate = (date: string) => {
+  const inputDate = dayjs(date);
+  const today = dayjs();
+
+  return Math.ceil(inputDate.diff(today, "day", true));
```
kasperg, line 109: Why do we tell `dayjs` to return a floating point number which is then rounded when [the library will itself will return an integer by default](https://day.js.org/docs/en/display/difference)?

src/core/utils/helpers/list-mapper.ts
```diff
@@ -169,4 +166,72 @@ export const mapManifestationToBasicDetailsType = (
...
+        3: "redeemed", // in publizon Redeemed
+        4: "cancelled", // in publizon Cancelled
+        5: "expired" // in publizon Expired
+      };
+
+      return {
+        identifier,
+        faust: null,
+        dateOfReservation: createdDateUtc,
+        expiryDate: expireDateUtc,
```
kasperg, line 196: Consider converting the mentioned here to proper date/dayjs objects during mapping. This means we have much better type safety throughout the rest of the system.

kasperg (changes requested):
:+1: nice stuff here.

There are a couple of places where I would like to see some changes. Primarily regarding the global query selector and HTML elements.

I also have some suggestions for improved type safety for you to consider.

src/apps/loan-list/materials/utils/status-circle-icon.tsx
```diff
@@ -0,0 +1,36 @@
...
+}) => {
+  const t = useText();
+
+  return (
+    <div
+      className="list-reservation__counter"
+      aria-label={t("loanListStatusCircleIconAriaLabelText")}
+    >
+      <div
+        role="progressbar"
```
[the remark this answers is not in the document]
sinejespersen: Yeah I think I might have copied it from the design system without being critical, I agree with your sentiment and will change it in both places. I'll make a pr in the design system as well - its on my todo :)

src/core/utils/types/reservation-type.ts
sinejespersen: It just isn't a number, it is a string, like "Reserveringshylde 74". I use the name from the [ReservationDetailsV2](https://github.com/danskernesdigitalebibliotek/dpl-react/blob/release/4/src/core/fbs/model/reservationDetailsV2.ts) to be consistent... But I agree it is a misleading name... Should I change it?

src/core/utils/helpers/general.ts
sinejespersen: I'll change it

src/apps/loan-list/materials/utils/status-circle-icon.tsx
```diff
@@ -0,0 +1,36 @@
...
+const StatusCircleIcon: FC<StatusCircleIconProps> = ({
+  color,
+  percent,
+  children
+}) => {
+  const t = useText();
+
+  return (
+    <div
+      className="list-reservation__counter"
```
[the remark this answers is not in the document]
sinejespersen: Fair point. I'll change

src/apps/loan-list/materials/utils/status-circle-icon.tsx
```diff
@@ -0,0 +1,36 @@
...
+      aria-label={t("loanListStatusCircleIconAriaLabelText")}
```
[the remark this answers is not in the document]
sinejespersen: I have on my endless todo to fix all the strings in the apps when the string interpolation works.... I'll add a todo instead, because it'll change in a future pr anyway.

src/apps/reservation-list/reservation-material/reservation-material.tsx
sinejespersen: no

src/apps/loan-list/modal/due-date-loans-modal.tsx
```diff
@@ -18,7 +18,7 @@ const DueDateLoansModal: FC<DueDateLoansModalProps> = ({
   loansModal
 }) => {
   const t = useText();
-  const aMonthAgo = dayjs().subtract(7, "month").format("YYYY-MM-DD");
+  const aMonthAgo = dayjs().subtract(1, "month").format("YYYY-MM-DD");
```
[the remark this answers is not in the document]
sinejespersen: 😶

src/apps/reservation-list/list/list.tsx
```diff
@@ -0,0 +1,32 @@
...
+const List: FC<ListProps> = ({ list, header }) => {
+  return (
+    <>
+      <div className="dpl-list-buttons m-32">
+        <h2 className="dpl-list-buttons__header">
+          {header}
+          <div className="dpl-list-buttons__power">{list.length}</div>
+        </h2>
+      </div>
+      <div className="list-reservation-container m-32">
```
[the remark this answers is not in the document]
sinejespersen: https://github.com/danskernesdigitalebibliotek/dpl-design-system/blob/release/4/src/stories/Library/list-buttons/list-buttons.scss :)

src/core/utils/types/reservation-type.ts
kasperg: Good question. I think it is fair to keep it around for consistency but adding a comment explaining why it is a string, perhaps with a example of what it might contain, would be helpful.

src/apps/loan-list/materials/stackable-material/additional-materials-button.tsx
```diff
@@ -28,18 +29,16 @@ const AdditionalMaterialsButton: FC<AdditionalMaterialsButtonProps> = ({
     <>
       <div
         className="list-reservation__hidden-explanation"
-        id={`materials-modal-${isMobile ? "mobile" : "desktop"}-text`}
+        id={`materials-modal-${showOn}-text`}
```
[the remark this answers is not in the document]
kasperg: Lets just leave it as is. You know more about the implementation than me here :smile:

kasperg (approved):
:+1: LGTM.

I have a few followup comments for you to consider before you mere.

PR #194, opened 2022-10-12
remove use of t as argument in helper

sinejespersen:
#### Description

It seems the t as argument makes my components re-render and re-render and re-render and re-render and re-render and ...

### DIFF
--- a/src/core/utils/helpers/list-mapper.ts
+++ b/src/core/utils/helpers/list-mapper.ts
@@ -11,6 +11,8 @@ import { store } from "../../store";
 export const getContributors = (creators: string[]) => {
   let returnContentString = "";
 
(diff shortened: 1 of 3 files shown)


src/core/utils/helpers/list-mapper.ts
```diff
@@ -108,10 +110,14 @@ export const mapProductToBasicDetailsType = (
     externalProductId
   } = material;
 
+  const {
```
spaceo, line 113: Sucks that you need to reintroduce this. But I understand that things need to move along.
I think you should put a TODO here about getting rid of it again in the future and the reason why it must be solved like this for now.

spaceo (changes requested):
Bleh. Too bad - but please see comment.

PR #196, opened 2022-10-19
Show only unique periodical editions in periodical dropdowns

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFSOEG-263

#### Description
This PR fixes an issue with periodical dropdowns. The data is set up in a way where if a library has 5 specimens of e.g. volume 2022 - 29 of Alt for damerne, there would be 5 records with the value 29. This resulted in overpopulating the dropdown with duplicate values. We now filter these out.

#### Screenshot of the result
image

#### Checklist

n/a


Adamik10:
Follow as a part of #197

PR #197, opened 2022-10-19
Fix/only unique periodicals on material page

Adamik10:
#### Link to issue

#### Description
This PR fixes an issue with periodical dropdowns on the material page and in the find on shelf modal, where the data for editions (e.g. VOGUE "2020 - week 32" edition) was shown multiple times if a library has more than one copy of it. 
Moreover, the following dropdown functionality was fixed: 
- chronologically first edition is selected on page load
- chronologically first edition is selected within a given year on year change

To reflect these changes, the material test was also adjusted.

#### Checklist

The PR also introduces addition of the full lodash library.
There is a general need for utility functions that work with objects, as the current out of the box object value filtering possibilities are not ideal, the produced code can be quite confusing and the need for them spans around the whole project. As such, we decided to add the full lodash package to the project. We previously agreed on only importing specific parts of the library, but now found out that lodash themselves don't recommend that (link for reference: https://lodash.com/per-method-packages) and this method of working with lodash would increase the bundle size of the project. Instead, we limit the bundle size by using the lodash-recommended package - babel-plugin-lodash that only imports specific methods used in components. Please refer to the link above for further explanation.


src/components/material/periodical/helper.ts
```diff
@@ -28,4 +40,71 @@ export function makePeriodicalEditionsFromHoldings(holdings: HoldingsV3[]) {
...
+  year: string,
+  selectPeriodicalHandler: (selectedPeriodical: PeriodicalEdition) => void
+) {
+  const changedFullPeriodicalEdition = groupList[year].find((edition) => {
+    return edition.volumeNumber === event.target.value;
+  });
+  if (changedFullPeriodicalEdition) {
+    selectPeriodicalHandler(changedFullPeriodicalEdition);
+  }
+}
```
kasperg, line 108: Consider updating `handleSelectEdition()` to accept a volumeNumber instead of an event.

That would mean that  we could call it `handleSelectYear()` instead of having the code duplicated. It would also abstract our helpers from React events tied closely to the UI. 

handleSelectEdition

src/components/material/periodical/helper.ts
```diff
@@ -28,4 +40,71 @@ export function makePeriodicalEditionsFromHoldings(holdings: HoldingsV3[]) {
...
+  const filteredPeriodicalEditionsObj = allYears.reduce((acc, curr, index) => {
+    // Sort editions array
+    // eslint-disable-next-line no-param-reassign
+    acc[curr] = filteredEditions[index].sort((a, b) => {
+      return a.localeCompare(b, "da-DK", { numeric: true });
+    });
+    return acc;
+  }, {} as { [key: string]: string[] });
+  return filteredPeriodicalEditionsObj;
+}
```
kasperg, line 71: I do not understand what is going on here. We should talk this through face to face.

Adamik10: Summary of the talk: 
There is a general need for utility functions that work with objects, as the current out of the box object value filtering possibilities are not ideal, the produced code can be quite confusing and the need for them spans around the whole project. As such, we decided to add the full lodash package to the project. We previously agreed on only importing specific parts of the library, but now found out that lodash themselves don't recommend that (link for reference: https://lodash.com/per-method-packages) and this method of working with lodash would increase the bundle size of the project. Instead, we limit the bundle size by using the lodash-recommended package - babel-plugin-lodash that only imports specific methods used in components. Please refer to the link above for further explanation.

Adamik10: fixed 😉

src/components/material/periodical/helper.ts
Adamik10: Good point! I didn't see that one! Thanks 😉

kasperg (approved):
:+1:

sinejespersen (approved):

sinejespersen (approved):
There seem to be an exported type in a helper (GroupList), why is this? (out of curiosity - therefore approved)

PR #198, opened 2022-10-21
Don't show "online material reserve button" if no manifestation have online access

Adamik10:
#### Link to issue
[DDFSOEG-264](https://reload.atlassian.net/browse/DDFSOEG-264)

#### Description
In case none of the material manifestations have online access, we shouldn't show the "online reservation" button on the material page. 

#### Screenshot of the result
n/a

#### Checklist

n/a

### DIFF
--- a/src/components/material/material-buttons/MaterialButtons.tsx
+++ b/src/components/material/material-buttons/MaterialButtons.tsx
@@ -4,6 +4,7 @@ import { AccessTypeCode } from "../../../core/dbc-gateway/generated/graphql";
 import { convertPostIdToFaustId } from "../../../core/utils/helpers/general";
 import { ButtonSize } from "../../../core/utils/types/button";
 import { Manifestation } from "../../../core/utils/types/entities";
+import { hasCorrectAccessType } from "./helper";
 import MaterialButtonsOnline from "./online/MaterialButtonsOnline";
 import MaterialButtonsFindOnShelf from "./physical/MaterialButtonsFindOnShelf";
 import MaterialButtonsPhysical from "./physical/MaterialButtonsPhysical";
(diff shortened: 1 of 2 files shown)


src/components/material/material-buttons/MaterialButtons.tsx
```diff
@@ -18,21 +18,28 @@ const MaterialButtons: FC<MaterialButtonsProps> = ({
...
-    (type) => type.code === AccessTypeCode.Physical
-  );
+  const hasCorrectAccessType = (
+    desiredAccessType: AccessTypeCode,
+    manifest: Manifestation
+  ) => {
+    return !!manifest.accessTypes.some(
+      (type) => type.code === desiredAccessType
+    );
+  };
```
kasperg, line 28: Consider whether this is something that is more widely usable.

Checking for an access type seems like something we would do a lot of places. If we have a good implementation for this then it should be in a helper and used where we can.

kasperg (approved):
:+1: but please consider my comments.

src/components/material/material-buttons/MaterialButtons.tsx
Adamik10: I see your point, and even though we're currently not using it any other places, I think it could be a helpful function in the future (in case we ever need it).

src/components/material/material-buttons/MaterialButtons.tsx
```diff
@@ -18,21 +18,28 @@ const MaterialButtons: FC<MaterialButtonsProps> = ({
...
   size
 }) => {
-  const hasPhysicalAccess = manifestation.accessTypes.some(
-    (type) => type.code === AccessTypeCode.Physical
-  );
+  const hasCorrectAccessType = (
+    desiredAccessType: AccessTypeCode,
+    manifest: Manifestation
+  ) => {
+    return !!manifest.accessTypes.some(
```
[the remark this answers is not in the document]
Adamik10: That is absolutely correct. I was being too smart for my own good haha.

sinejespersen (approved):

PR #200, opened 2022-10-21
Add facet browser

kasperbirch1:
#### This PR adds FacetBrowser modal
A button (+ MORE FILTERS) has been added to the search results page. This opens the facet browser modal which retrieves facets from the query that has also been searched for.

The search results change in the background when terms are added or removed

We have added a new type of disclosure because our current one uses html tag details and we couldn't get that to work with our logic for when FacetDisclosure should be open

We have also added a resetPager function that we use when we change the filters

And then we have changed the queries to be able to use filters

#### Screenshot of the result
Skærmbillede 2022-10-21 kl  15 56 35
Skærmbillede 2022-10-21 kl  15 56 54

#### Checklist


src/apps/search-result/search-result.dev.tsx
```diff
@@ -98,6 +98,76 @@ export default {
...
+      name: "facet browser screen reader modal description text",
+      defaultValue: "Modal for facet browser",
+      control: { type: "text" }
+    },
+    facetBrowserModalCloseModalAriaLabelText: {
+      name: "facet browser close modal aria label text",
+      defaultValue: "Close facet browser modal",
+      control: { type: "text" }
+    },
+    accessTypesText: {
```
spaceo, line 112: Prefix m. facet or similar?

src/core/utils/modal.tsx
```diff
@@ -46,46 +46,62 @@ function Modal({
...
-        zIndex: modalIds.indexOf(modalId) + 10
+      className="modal__backdrop"
+      role="button"
+      tabIndex={0}
+      onClick={(e) => {
+        if (e.target === e.currentTarget) dispatch(closeModal({ modalId }));
+      }}
+      // TODO is this the right way to handle this?
+      onKeyPress={(e) => {
+        if (e.key === "Escape") dispatch(closeModal({ modalId }));
```
spaceo, line 57: When has the demand for an escape key option been introduced?

src/components/tag/Tag.tsx
```diff
@@ -0,0 +1,29 @@
...
+  children
+}) => {
+  return (
+    <button
+      type="button"
+      className={clsx(
+        "tag tag--outlined",
+        selected && "tag--outlined-selected"
+      )}
+      onClick={onClick ?? undefined}
```
spaceo, line 22: Is it necessary to specify undefined here? The prop type already indicates that it can be undefined.

src/components/facet-browser/FacetBrowserModal.tsx
```diff
@@ -0,0 +1,58 @@
...
+    <Modal
+      classNames="modal-right modal--no-padding"
+      modalId={FacetBrowserModalId}
+      screenReaderModalDescriptionText={t(
+        "facetBrowserModalScreenReaderModalDescriptionText"
+      )}
+      closeModalAriaLabelText={t("facetBrowserModalCloseModalAriaLabelText")}
+    >
+      {/* TODO Is it necessary to have "as FacetResult[]" ? */}
+      <FacetBrowserModalBody facets={data.search.facets as FacetResult[]} />
```
spaceo, line 53: See how we have done this in the reservation modal. So we make sure to define the data type at its source:
https://github.com/danskernesdigitalebibliotek/dpl-react/blob/release/4/src/components/reservation/ReservationModalBody.tsx#L94:L94

src/components/search-bar/search-result-header/SearchResultHeader.tsx
```diff
@@ -11,10 +15,19 @@ const SearchResultHeader: React.FC<SearchResultHeaderProps> = ({
...
-      {`${t("showingResultsForText")} “${q}” (${hitcount})`}
-    </h1>
+    <>
+      <h1 className="text-header-h2 mb-16 search-result-title">
+        {`${t("showingResultsForText")} “${q}” (${hitcount})`}
+      </h1>
+      <Tag
+        onClick={() => dispatch(openModal({ modalId: FacetBrowserModalId }))}
+      >
+        + FLERE FILTRE
```
spaceo, line 28: Translatable text...

spaceo (commented):
Looks good. A few comments...

src/apps/search-result/search-result.tsx
```diff
@@ -21,6 +23,36 @@ const SearchResult: React.FC<SearchResultProps> = ({ q, pageSize }) => {
...
+          ...filters,
+          [facet]: { [term]: score }
+        });
+      }
+    }
+
+    if (action === "remove") {
+      if (Object.keys(filters).includes(facet)) {
+        const copy = { ...filters };
+        delete copy[facet][term];
```
spaceo, line 51: Yes, of course, we need this so we are not manipulating filters but a clone of filters :)

src/apps/search-result/facet-browser.test.ts
```diff
@@ -0,0 +1,247 @@
+const coverUrlPattern = /^https:\/\/res\.cloudinary\.com\/.*\.(jpg|jpeg|png)$/;
+
+describe("Facet Browser", () => {
+  // I have used it.only because it is best practice to have self contained tests
+  // https://docs.cypress.io/guides/references/best-practices#Having-tests-rely-on-the-state-of-previous-tests
+  it.only("Open facet browser and check if all the facets are there", () => {
```
kasperg, line 6: Pleas remove `only`.

I appreciate the intention here but you have misunderstood [the purpose of `.only()`](https://docs.cypress.io/guides/references/best-practices#Having-tests-rely-on-the-state-of-previous-tests).

> Usually you wouldn't check code using test.only into source control - you would use it for debugging, and remove it once you have fixed the broken tests.

src/components/facet-browser/FacetBrowserDisclosure.tsx
```diff
@@ -0,0 +1,87 @@
...
+      <div
+        className={clsx(
+          "disclosure__headline text-body-large",
+          removeHeadlinePadding && "disclosure__headline--no-padding"
+        )}
+        onClick={onClickHandler}
+        onKeyDown={onClickHandler}
+        role="button"
+        tabIndex={0}
+      >
```
kasperg, line 57: Please rework this structure.

It renders two nested elements each with `role="button"`. That does not seem right. I would expect the headline to be the button and then it controls the element with the children.

src/apps/search-result/facet-browser.test.ts
```diff
@@ -0,0 +1,247 @@
...
+    cy.contains(".disclosure__text", "Main languages");
+    cy.contains(".disclosure__text", "Access types");
+    cy.contains(".disclosure__text", "Children or adults");
+    cy.contains(".disclosure__text", "Creators");
+    cy.contains(".disclosure__text", "Fiction or nonfiction");
+    cy.contains(".disclosure__text", "Genre and form");
+    cy.contains(".disclosure__text", "Material types");
+    cy.contains(".disclosure__text", "Subjects");
+    cy.contains(".disclosure__text", "Work types");
```
kasperg, line 15: Please [do not select against CSS classes as they are likely to change](https://docs.cypress.io/guides/references/best-practices#Selecting-Elements).

I think it would be OK to select based on the text content alone. If we want to be more specific then [selecting based on accessibility roles is also a good idea](https://testing-library.com/docs/queries/about/#priority).

This goes for the entire test.

src/apps/search-result/facet-browser.test.ts
```diff
@@ -0,0 +1,247 @@
...
+    cy.contains(".facet-browser__facet-group button", "Dansk");
+    cy.contains(
+      ".facet-browser__facet-group button",
+      "Sproget kan ikke bestemmes"
+    );
+    cy.contains(".facet-browser__facet-group button", "Fransk");
+    cy.contains(".facet-browser__facet-group button", "Latin");
+    cy.contains(".disclosure__text", "Main languages").click();
+    cy.get(`[aria-label="Close facet browser modal"]`).click();
+  });
```
kasperg, line 33: Please consider reducing this.

I see a lot of work went into writing all this but I do not think we have any reason to believe that one facet would behave differently than another. I would consider it sufficient to test that all facets are rendered and that all terms within one facet are rendered.

src/apps/search-result/search-result.dev.tsx
```diff
@@ -101,7 +101,83 @@ export default {
...
+      control: { type: "text" }
+    },
+    clearAllText: {
+      name: "Clear all text",
+      defaultValue: "Clear all",
+      control: { type: "text" }
+    },
+    addMoreFiltersText: {
+      name: "Add more filters text",
+      defaultValue: "+ MORE FILTERS",
```
kasperg, line 179: Please do not capitalize text if not needed.

If it is a visual effect then casing should be managed with CSS.

src/apps/search-result/facet-browser.test.ts
```diff
@@ -0,0 +1,247 @@
...
+    cy.contains("button", "+ MORE FILTERS").click();
+    cy.get(`[aria-controls="facet-creators"]`).should(
+      "have.attr",
+      "aria-expanded",
+      "true"
+    );
+    cy.contains("button", "Joanne K. Rowling").should(
+      "have.class",
+      "tag--outlined-selected"
+    );
```
kasperg, line 177: Please use aria attributes instead of CSS classes to check state.

This goes for the entire test.

src/apps/search-result/facet-browser.test.ts
```diff
@@ -0,0 +1,247 @@
...
+
+    cy.log("Close modal and check for new results");
+    cy.get(`[aria-label="Close facet browser modal"]`).click();
+    cy.contains("h1", "“harry” (36)");
+
+    cy.log(
+      "Open modal and check if creators are opened and Joanne K. Rowling is selected"
+    );
+    cy.contains("button", "+ MORE FILTERS").click();
+    cy.get(`[aria-controls="facet-creators"]`).should(
```
kasperg, line 169: Consider selecting the element using the same approach as you used on L148.

Consistency is good.

src/apps/search-result/facet-browser.test.ts
kasperg: Consider naming and the [describe/it pattern](https://prowe214.medium.com/unit-tests-more-readable-describe-it-statements-in-protractor-jasmine-3810b07a6ac5)

Example:

To describe... (the) facet browser:
 - It... renders all facets
 - It ... renders all terms in a facet when clicked
 - It ... updates the search result when a term is selected
 - It ... updates the search result when a term is deselected
 - It ... renders all facets as closed by default
 - It ... renders a facet as open when it has a selected term 
 - It ... renders a facet as closed when it has no more selected terms

src/apps/search-result/search-result.tsx
```diff
@@ -36,57 +39,57 @@ const SearchResult: React.FC<SearchResultProps> = ({ q, pageSize }) => {
...
-  );
+
+    const {
+      search: { works: resultWorks, hitcount: resultCount }
+    } = data as {
+      search: {
+        works: Work[];
+        hitcount: SearchWithPaginationQuery["search"]["hitcount"];
+      };
+    };
```
kasperg, line 88: Consider if we can get rid of using `as` here.

If you have already worked through this with Mikkel then you can disregard this.

src/components/facet-browser/FacetBrowserModal.tsx
```diff
@@ -0,0 +1,94 @@
...
+      classNames="modal-right modal--no-padding"
+      modalId={FacetBrowserModalId}
+      screenReaderModalDescriptionText={t(
+        "facetBrowserModalScreenReaderModalDescriptionText"
+      )}
+      closeModalAriaLabelText={t("facetBrowserModalCloseModalAriaLabelText")}
+    >
+      {isLoading || !data ? null : (
+        <FacetBrowserModalBody
+          facets={data.search.facets as FacetResult[]}
```
kasperg, line 83: Consider if we can get rid of using `as` here.

If you have already worked through this with Mikkel then you can disregard this.

kasperg (changes requested):
:+1: Lots of code, good stuff, some comments for you to look into.

src/apps/search-result/facet-browser.test.ts
kasperbirch1: It is not the same element as on L148. It is the parent element that shows if Disclosure is open

src/components/facet-browser/FacetBrowserModalBody.tsx
```diff
@@ -73,21 +73,23 @@ const FacetBrowserModalBody: React.FunctionComponent<
                 }
 
                 return (
-                  <Tag
+                  <ButtonTag
                     key={term}
-                    onClick={() =>
+                    onClick={(e) => {
+                      // This to prevent the disclosure from closing when clicking on a tag because event bobbling
+                      e.stopPropagation();
```
kasperg, line 80: :+1: nice commenting.

kasperg (approved):
:+1: LGTM

sinejespersen (approved):

PR #205, opened 2022-10-24
OLD: Feature/show campaign + test campaign functionality

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFSOEG-242

#### Description
This PR adds a new feature: showing administrator-created campaign on the search result page. The premise for the functionality is that if a given search (e.g. search for "Harry Potter") matches with any of the administrator-specified criteria, a campaign banner show up above the search results. This way a library can show seasonal content/current happenings, etc. An example in the "HP" search case could be that the library has J.K.Rowling coming and reading some parts of a Harry Potter book, so they want to advertize it. 
PR also includes a cypress test for the campaigns.

#### Screenshot of the result
xyz

#### Checklist

n/a


Adamik10:
Follow -> #209

PR #206, opened 2022-10-26
Feature/5390 clean up renew modal

sinejespersen:
#### Link to issue

https://platform.dandigbib.org/issues/5390

#### Description

- add pagesize as props to RenewLoansModal
- add some missing texts
- split up LoanListProps (I've noticed you guys doing that, and I think its grrreat
- remove renewalStatus-code which we will not be using (this was there because we had some unanswered questions, that are now answered questions, and it will not be needed)
- created ModalDetailsHeader -> which is the header for details modals, the name, cover and such
- moved the renewbutton into separate component (RenewButton)
- renamed some of the very long names...
- 
#### Screenshot of the result

#### Checklist

### DIFF
--- a/src/apps/loan-list/modal/renew-button.tsx
+++ b/src/apps/loan-list/modal/renew-button.tsx
@@ -3,9 +3,10 @@ import { RenewedLoanV2 } from "../../../core/fbs/model";
 import { useText } from "../../../core/utils/text";
 import { useRenewLoansV2 } from "../../../core/fbs/fbs";
 import IconCheckmark from "../../../components/icon-checkmark/icon-checkmark";
+import { FaustId } from "../../../core/utils/types/ids";
 
 interface RenewButtonProps {
-  faust: string;
+  faust: FaustId;
   setDueDate: (date: string) => void;
 }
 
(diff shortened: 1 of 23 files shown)


src/apps/loan-list/modal/renew-button.tsx
```diff
@@ -0,0 +1,104 @@
...
+          onError: () => {}
+        }
+      );
+    },
+    [determineStatusText, mutate, setDueDate]
+  );
+
+  return (
+    <div className="modal-details__buttons">
+      {renewable && faust && (
```
spaceo, line 67: According to the `RenewButtonProps["faust"]` it will always be string. And in that case a check here for the faust (why is it not a `FaustId` btw?) is not necessary.

src/apps/loan-list/modal/renew-button.tsx
```diff
@@ -0,0 +1,104 @@
...
+        <button
+          type="button"
+          onClick={() => renew(parseInt(faust, 10))}
```
spaceo, line 70: Better with a callback outside of the rendering

spaceo (commented):
Approved 👍  with some comments

PR #207, opened 2022-10-26
Feature/5404 add periodical and series to basic details type

sinejespersen:
#### Link to issue

[- ](https://platform.dandigbib.org/issues/5404

#### Description

- add periodical and series to basic details type
- rename graphql from feedback on previous pr

#### Screenshot of the result

**The crown**

image

**2020 nr 10**

image

#### Checklist


src/core/utils/helpers/list-mapper.ts
```diff
@@ -36,6 +36,24 @@ function getYearFromDataString(date: string) {
...
+    title: string;
+    numberInSeries?: {
+      number?: Array<number> | null;
+    } | null;
+  }[]
+) {
+  return series
+    .map(({ title, numberInSeries }) => {
+      if (numberInSeries && numberInSeries.number) {
+        return `${title} ${numberInSeries?.number?.[0]}`;
```
spaceo, line 50: since you already checked for `numberInSeries.number` it should only be necessary to know if key 0 exists.

src/core/utils/helpers/list-mapper.ts
```diff
@@ -36,6 +36,24 @@ function getYearFromDataString(date: string) {
   return new Date(date).getFullYear();
 }
 
+function getSeriesString(
```
spaceo, line 39: We also use series in our part of the codebase. Could you please double check if we do it in the same manner?

spaceo (changes requested):
Looks good. I have one thing I would like you to investigate though...

PR #209, opened 2022-10-28
Feature/show campaign + test campaign functionality

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFSOEG-242

#### Description
This PR adds a new feature: showing an administrator-created campaign on the search result page. The premise for the functionality is that if a given search (e.g. search for "Harry Potter") matches with any of the administrator-specified criteria (set up in the cms admin), a campaign banner show up above the search results. This way a library can show seasonal content/current happenings, etc. An example in the "HP" search case could be that the library has J.K.Rowling coming and reading some parts of a Harry Potter book, so they want to advertize it. 
The main task was to fetch dynamic facets based on a given search and then use them to fetch a campaign (if there is one matches currently created campaigns).
PR also includes a cypress test for the campaigns.

#### Screenshot of the result
Full campaign:
image

Image only:
image

Text only:
image

#### Checklist

In order to be able to run this locally and have it working - go to src/core/dpl-cms/mutator/fetcher.ts and change defaultBaseUrl to be: https://varnish.pr-156.dpl-cms.dplplat01.dpl.reload.dk

Please only go through commits that have my face _only_ on them:  [0ef703b](https://github.com/danskernesdigitalebibliotek/dpl-react/pull/209/commits/0ef703b084d2b2e5f10faa16e25ed4ca918232b9) - [8616ac8](https://github.com/danskernesdigitalebibliotek/dpl-react/pull/209/commits/8616ac82bc31bb570df1daac8f87509fef2529bd) and [c38c29c](https://github.com/danskernesdigitalebibliotek/dpl-react/pull/209/commits/c38c29ce889126a2cc91a886cd72fc8b82068ba6) - [4e83c12](https://github.com/danskernesdigitalebibliotek/dpl-react/pull/209/commits/4e83c126d0f541905508d6b7851b107cf01b97e0)


src/components/atoms/link-no-style.tsx
```diff
@@ -18,7 +18,7 @@ export const LinkNoStyle: React.FC<LinkNoStyleProps> = ({
     <Link
       href={url}
       isNewTab={isNewTab}
-      className={`"hide-linkstyle" ${className}`}
+      className={`hide-linkstyle${` ${className}` || ""}`}
```
kasperg, line 21: Consider using [`clsx`](https://www.npmjs.com/package/clsx) to manage these classes.

This double nesting of string literals is pretty hard to read - for me at least. I know it is outside the functional scope of this PR but you started editing here 😅 .

src/components/campaign/campaign.test.ts
```diff
@@ -0,0 +1,87 @@
+const coverUrlPattern = /^https:\/\/res\.cloudinary\.com\/.*\.(jpg|jpeg|png)$/;
+
+describe("Find on shelf modal - default", () => {
```
kasperg, line 3:
```suggestion
describe("Campaign", () => {
```

src/components/campaign/campaign-body.tsx
```diff
@@ -0,0 +1,28 @@
...
+          className={`campaign__image ${
+            !campaignData.text ? "campaign__image--full-width" : ""
+          }`}
+          src={campaignData.image.url}
+          alt={campaignData.image.alt}
+        />
+      )}
+      <h4 className="campaign__title campaign__title--ellipsis">
+        {campaignData.text}
+      </h4>
```
kasperg: Please make the title conditional of the `text` property.

It seems like this will add an empty title element even if is an image-only campaign. That does not seem right.

src/components/campaign/campaign.test.ts
```diff
@@ -0,0 +1,87 @@
...
+
+  it("Shows an image-only campaign without text", () => {
+    // Intercept campaign data call - full campaign.
+    cy.fixture("search-result/campaign-image-only.json")
+      .then((result) => {
+        cy.intercept("POST", "**/dpl_campaign/match", result);
+      })
+      .as("Campaign service");
+
+    cy.get("section").find("Lorem ipsum Harry Potter").should("not.exist");
```
kasperg, line 83: Please change this to test for the availability of a campaign image with an alt text and potentially the lack of a campaign text.

The fact that a specific title is not shown does not really prove the existence of a image-only campaign.

src/components/campaign/campaign.test.ts
```diff
@@ -0,0 +1,87 @@
...
+        cy.intercept("POST", "**/dpl_campaign/match", result);
+      })
+      .as("Campaign service");
+
+    cy.get("section").contains("Lorem ipsum Harry Potter");
+  });
+
+  it("Shows a text-only campaign without an image", () => {
+    // Intercept campaign data call - full campaign.
+    cy.fixture("search-result/campaign-text-only.json")
```
kasperg, line 66: Please remove or update this comment.

1. It is wrong. We do not work with a full campaign here.
2. It duplicates what is in my opinion a pretty trivial `cy.fixture` setup.

src/components/campaign/campaign.test.ts
```diff
@@ -0,0 +1,87 @@
...
+      .then((result) => {
+        cy.intercept("POST", "**/dpl_campaign/match", result);
+      })
+      .as("Campaign service");
+
+    cy.get("section").find("img").should("not.exist");
+  });
+
+  it("Shows an image-only campaign without text", () => {
+    // Intercept campaign data call - full campaign.
```
kasperg, line 76: As above.

src/components/campaign/campaign.test.ts
```diff
@@ -0,0 +1,87 @@
...
+
+  it("Shows a full campaign", () => {
+    // Intercept campaign data call - full campaign.
+    cy.fixture("search-result/campaign.json")
+      .then((result) => {
+        cy.intercept("POST", "**/dpl_campaign/match", result);
+      })
+      .as("Campaign service");
+
+    cy.get("section").contains("Lorem ipsum Harry Potter");
```
kasperg, line 61: Consider also testing for the image.

If this is the test of a full campagin then I would expect that it also tested for the image as a full campaign is both image and text.

Also consider adding a test for url handling.

src/components/search-result-list/SearchResultList.tsx
```diff
@@ -0,0 +1,87 @@
...
+    ...(isObjectEmpty(filters)
+      ? {}
+      : { filters: { ...formatFilters(filters) } })
+  });
+
+  useEffect(() => {
+    if (data) {
+      mutate(
+        {
+          data: data.search.facets as CampaignMatchPOSTBodyItem[]
```
kasperg, line 59: I tried to define the arguments for the facets required by the campaign service to match the facets returned by the facet service. Having to use `as` here indicate there is a difference. I wonder what it is?

src/components/search-result-list/SearchResultList.tsx
```diff
@@ -0,0 +1,87 @@
...
+    }
+  }, [mutate, data]);
+
+  return (
+    <ul className="search-result-page__list my-32">
+      {campaignData && campaignData.data && (
+        <Campaign campaignData={campaignData.data} />
+      )}
+      {resultItems.map((item, i) => (
+        <li key={item.workId}>
```
kasperg, line 79: Please do not return the campaign as a part of the search result list (`<ul>`).

It is not a list item.

Also consider making the search result list a `<ol>` instead of an `<ul>` since the results are in fact ordered. Perhaps that is outside the scope of this PR.

src/components/search-result-list/SearchResultList.tsx
```diff
@@ -0,0 +1,87 @@
...
+    null
+  );
+  const { data } = useSearchFacetQuery({
+    q: { all: q },
+    facets: allFacetFields,
+    facetLimit: 10,
+    ...(isObjectEmpty(filters)
+      ? {}
+      : { filters: { ...formatFilters(filters) } })
+  });
```
kasperg, line 53: Please reconsider where in the stack facets are retrieved and whether it can be pushed up.

As far as I can tell this implies that we issue another facet query from the search result list to get the facets to use for campaign fetching. This is now done twice: Here and in `FacetBrowserModal.tsx`.

As I see it is would be preferable to do it once in `search-result.tsx` and then passed down to subcomponents.

kasperg (changes requested):
:+1: I have reviewed the files mentioned. I think the implementation has some minor and major problems that I would like you to look into. Perhaps this needs to be coordinated with the other work in the facet browser.

src/components/atoms/link-no-style.tsx
Adamik10: done :)

src/components/campaign/campaign.test.ts
Adamik10: Of course! Good eye! Thanks 😉

src/components/campaign/campaign.test.ts
Adamik10: 👍

src/components/campaign/campaign.test.ts
Adamik10: Done :)

src/components/campaign/campaign.test.ts
Adamik10: That is a very fair point! We now test for all the aspects 😉

src/components/search-result-list/SearchResultList.tsx
Adamik10: Types of property 'score' are incompatible -> Type 'number | **null** | undefined' is not assignable to type 'number | undefined'.

src/components/campaign/campaign-body.tsx
Adamik10: Of course! How did I not see this myself.. 
Thanks 😉

src/components/search-result-list/SearchResultList.tsx
Adamik10: Roger! Makes sense 😉 
I dare not change it to an ordered list as a part of this PR because of the looming shadow of failing tests.

src/components/search-result-list/SearchResultList.tsx
Adamik10: Yeah that's what I thought too, so I moved it up, but then @spaceo told me that the call is cached (which I think is very cool), so the second query just uses the cached data.

src/components/atoms/link-no-style.tsx
Adamik10: Aaaaand reverted - clsx just renders the prop name instead of taking the value, so there were some problems with the styling where the covers on search result page were assigned class "className".

src/components/search-result-list/SearchResultList.tsx
kasperg: We disagree here.

I do not think we should rely on cross-component query caching unless we have to. With both react-query and Redux it means we have multiple state managers which does not sound like a good idea. With components being almost adjacent in the tree there is little work in passing the facets down and on the other hand we risk the two queries moving out of sync and thus accidentally making the cache unusable.

If we wanted to render the facet browser on its own in Storybook it would be useful but that is not the case at the moment.

Oh well.  I said my piece. We leave it as is.

kasperg (approved):
All campaign changes are good to me.

src/components/campaign/campaign.test.ts
```diff
@@ -50,36 +55,51 @@ describe("Find on shelf modal - default", () => {
...
+        cy.intercept("**/dpl_campaign/match", result);
       })
-      .as("Campaign service");
+      .as("Campaign service - full campaign");
 
-    cy.get("section").contains("Lorem ipsum Harry Potter");
+    cy.get("img")
+      .should("be.visible")
+      .and("have.prop", "naturalWidth")
+      .should("be.greaterThan", 0);
```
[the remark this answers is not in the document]
Adamik10: I made it into a helper function called isImageLoaded(), let's see if it can be reused other places though. Because it just takes the first image. Maybe it could be refactored in the future depending on the use cases.

src/apps/demo-modal/demo-modal.test.ts
```diff
@@ -17,7 +17,10 @@ describe("Opens demo modal", () => {
       .should("contain", "demo-modal-two");
     cy.get(".modal").find("button.modal-btn-close").should("exist");
     // Close the second modal again
-    cy.get(".modal").find("button.modal-btn-close").first().click();
+    cy.get(".modal")
+      .find("button.modal-btn-close")
+      .first()
+      .click({ force: true });
```
sinejespersen, line 23: why force:true?

Adamik10: @sinejespersen I'm not sure, actually. That's a part of @kasperbirch1 's changes from PR #200 (My PR won't be merged in before his is approved, so he will answer your question if you ask in #200 ).

sinejespersen (approved):

PR #211, opened 2022-11-02
Feature/5401 pause reservations

sinejespersen:
#### Link to issue

https://platform.dandigbib.org/issues/5401

#### Description

- add ${json-unit.any-string} to wiremock file so input can be anything
- remove unnecessary date check from function materialIsOverdue
- add pauseReservationStartDateConfig -> the start date of the pause is configurable
- rename something from "line" to "queue"
- Add a bunch of texts
- add test of reservation list 
- add user to reservation list (as the pause has to do with the user and not the reservations)
- add separate functions for getting reservations that are physical/digital (due to sorting, this is also why the title has been added to the reservation type)
- refetch user on close modal
- add pause reservation "toggle button" 
- rename delete reservation test
- add pause reservation test
- add date input file
- add pause reservation modal
- add "you are in the front of the line!" to reservation item, which was forgotten

#### Screenshot of the result

image

image

#### Checklist


src/apps/loan-list/utils/helpers.ts
```diff
@@ -31,10 +31,7 @@ export const removeLoansWithIds = (list: LoanType[], ids: string[]) => {
 };
 
 export const materialIsOverdue = (date: string | undefined) => {
```
spaceo, line 33: While you are at it with reducing the function you could also write:
```
export const materialIsOverdue = (date: string | undefined) =>
  dayjs().isAfter(dayjs(date), "day");

File: src/apps/reservation-list/list/reservation-pause-toggler.tsx [hunk 2]
```diff
@@ -0,0 +1,69 @@
...
+  const { pauseReservation } = getModalIds();
+  const [onHoldDates, setOnHoldDates] = useState<string>();
+
+  const openPauseReservationModal = () => {
+    open(pauseReservation as string);
+  };
+
+  useEffect(() => {
+    if (user && user.onHold) {
+      setOnHoldDates(`${user.onHold.from} - ${user.onHold.to}`);
```
spaceo, line 27: Don't you need `onHoldDates` in the dependency array if `useEffect`?
Could be that you have an infinite loop now because of it. Also I would set the initial state of `onHoldDates ` to `null` and add a check for it in:
```
if (!onHoldDates  && user && user.onHold) {

File: src/apps/reservation-list/list/reservation-pause-toggler.tsx [hunk 3]
```diff
@@ -0,0 +1,69 @@
...
+            >
+              {onHoldDates}
+            </span>
+          )}
+          <div className="dpl-pause-reservation-component__flex__button">
+            <button
+              aria-label={t("reservationListPauseReservationAriaModalText")}
+              type="button"
+              onClick={openPauseReservationModal}
+              className={`dpl-toggle-button ${
```
spaceo, line 55: We normally use `clsx` for this

src/apps/reservation-list/modal/pause-reservation/date-input.tsx
```diff
@@ -0,0 +1,34 @@
+import React, { FC } from "react";
+
+interface DateInputProps {
+  label: string;
+  id: string;
+  value: string;
+  onChange: (date: string) => void;
+}
+
+const DateInput: FC<DateInputProps> = ({ label, id, value, onChange }) => {
```
spaceo, line 10: This is very generic, and that is cool. But I think we should move it to a global place then.

src/apps/reservation-list/modal/pause-reservation/pause-reservation.tsx
```diff
@@ -0,0 +1,123 @@
...
+}
+
+const PauseReservation: FC<PauseReservationProps> = ({ id, user }) => {
+  const t = useText();
+  const { mutate } = useUpdateV5();
+  const { close } = useModalButtonHandler();
+  const { pauseReservation } = getModalIds();
+  const config = useConfig();
+  const [startDate, setStartDate] = useState<string>(
+    (config("pauseReservationStartDateConfig") as string) || ""
```
spaceo, line 23: It should be possible to write:
```
config<string>(....

Comment (spaceo) [reply src/apps/reservation-list/modal/pause-reservation/pause-reservation.tsx:23]: Instead of 
```
...as string

src/apps/reservation-list/modal/pause-reservation/pause-reservation.tsx
```diff
@@ -0,0 +1,123 @@
...
+        },
+        {
+          onSuccess: () => {
+            close(pauseReservation as string);
+          },
+          // todo error handling, missing in figma
+          onError: () => {}
+        }
+      );
+    }
```
spaceo, line 49: We do this in the reservation modal as well. I think we should use the same function:
https://github.com/reload/dpl-react/blob/eeffaae91f848f662bcffa4563169b2f78cd014c/src/components/reservation/forms/ModalReservationFormText.tsx#L61-L87
So we could make the unSubmit function in the link I wrote and turn it into a global patron data submit handler.
One important thing that is happening is that the QueryClient is feeded with the updated user data. And that results in two things:
* We get updated user data next time we request it in the application (otherwise we would get old user data from the cache)
* Since the QueryClient gets the updated user data, the next time we request the user data, it will be served from the cache and an actual request is not being performed

src/apps/reservation-list/modal/pause-reservation/pause-reservation.tsx
```diff
@@ -0,0 +1,123 @@
...
+  }, [close, endDate, mutate, pauseReservation, startDate, user]);
+
+  useEffect(() => {
+    if (user?.onHold?.from) {
+      setStartDate(user.onHold.from);
+    }
+    if (user?.onHold?.to) {
+      setEndDate(user.onHold.to);
+    }
+  }, [user?.onHold]);
```
spaceo, line 59: seems weird to me with optional chaining in the dependency array. But I guess it works? :)

src/apps/reservation-list/utils/helpers.ts
```diff
@@ -12,8 +12,23 @@ export const sortByOldestPickupDeadline = (list: ReservationType[]) => {
...
-export const getReserved = (list: ReservationType[]) => {
-  return [...list].filter(({ state }) => state === "reserved");
+export const getReservedDigital = (list: ReservationType[]) => {
+  return [...list]
+    .filter(({ state }) => state === "reserved")
+    .sort(
+      (objA, objB) =>
+        new Date(objA.pickupDeadline || new Date()).getTime() -
+          new Date(objB.pickupDeadline || new Date()).getTime() ||
+        (objA.title || "").localeCompare(objB.title || "")
```
spaceo, line 22: I don't understand this bit

spaceo (changes requested):
I have some comments

src/apps/reservation-list/list/reservation-pause-toggler.tsx
sinejespersen: I dont think I understand, why do I need onHoldDates in the dependency array?

src/apps/reservation-list/modal/pause-reservation/date-input.tsx
sinejespersen: moved it

src/apps/loan-list/utils/helpers.ts
sinejespersen: changed

src/apps/reservation-list/modal/pause-reservation/pause-reservation.tsx
spaceo: It is because of how this [function overloading](https://github.com/danskernesdigitalebibliotek/dpl-react/blob/a9f8d508e91ac7fb8ffa6ade98b784210a7f2a01/src/core/utils/config.tsx#L5-L22) has been constructed.
I guess you can leave it as is. It would be nice with a todo before the line wit `as string` mentioneing that it would be nice to do it otherwise but then the overloading of the `config`function would need to be changed.

spaceo (approved):

PR #213, opened 2022-11-04
Add facet line

kasperbirch1:
#### Link to issue

#### Description

#### Screenshot of the result

#### Checklist


src/components/search-bar/search-result-header/SearchResultHeader.tsx
```diff
@@ -1,32 +1,67 @@
...
+    },
+    {
+      keepPreviousData: true,
+      placeholderData: {
+        search: {
+          facets: getPlaceHolderFacets(lineFacets)
+        }
+      }
+    }
+  );
```
kasperg, line 50: This is misunderstood.

1. The facet line is based on the `intelligentFacets` query - not the standard one.
2. The query should not use a limited/different sent of facets compared to the standard one
3. It is the number of terms for each facet that determines whether the facet should be shown as a dropdown (multiple terms) or as a button (single term).

src/components/facet-browser/FacetBrowserModalBody.tsx
```diff
@@ -50,15 +40,16 @@ const FacetBrowserModalBody: React.FunctionComponent<
         // Remove facets disclosures with no tags
         if (values.length === 0) return null;
 
+        const hasChosenTerms = Boolean(filters[name]);
```
kasperg, line 43: Consider using the word "selected" instead of "chosen". It is also what we use for the individual terms on L58.

src/components/facet-line/Dropdown.tsx
```diff
@@ -0,0 +1,100 @@
...
+
+export type DropdownItem = {
+  title: string;
+  href?: string;
+  disabled?: boolean;
+  value?: unknown;
+  selected?: boolean;
+};
+
+export const DropdownIcon = ({
```
kasperg, line 15: Please refactor this dropdown element into a general reusable component.

It seems as if it is getting duplicated throughout the project. It would be preferable to have it separately like we have for checkboxes and date-input.

src/components/facet-line/FacetLine.tsx
```diff
@@ -0,0 +1,127 @@
...
+      });
+    }
+  };
+
+  return (
+    <ul className="facet-line mt-48">
+      {facets.map(({ name, values }) => {
+        if (showFacetAsSelect.includes(name)) {
+          return (
+            <li className="facet-line__item">
```
kasperg, line 69: :+1: for using a HTML lists to render these elements.

src/components/facet-line/FacetLineSelected.tsx
```diff
@@ -0,0 +1,50 @@
...
+  filterHandler
+}) => {
+  return (
+    <ul className="facet-line-selected">
+      {Object.entries(filters).map(([facet, value]) => {
+        return (
+          <>
+            {Object.entries(value).map(([label, term]) => {
+              return (
+                <li className="facet-line-selected-terms">
```
kasperg, line 24: It is outside the scope of this PR but I would have expected the terms to be BEM elements within the selected block e.g. `facet-line-selected__term`.

src/components/search-bar/search-result-header/SearchResultHeader.tsx
```diff
@@ -1,32 +1,67 @@
...
-        {t("addMoreFiltersText")}
-      </ButtonTag>
+      {data && (
+        <FacetLine
+          filters={filters}
+          facets={data.search.facets as FacetResult[]}
+          filterHandler={filterHandler}
+        />
+      )}
+      <FacetLineSelected filters={filters} filterHandler={filterHandler} />
```
kasperg, line 64: Consider moving all facet components (the selectable terms and the selected terms) into a single element and then using this element as a part of the search result directly instead of passing it through the header.

This is the same approach as we use for the facet browser and the header is not really related to the facetline either.

kasperg (changes requested):

src/components/facet-line/Dropdown.tsx
kasperbirch1: I have made a reusable component ready but I made a separate PR to use it in the places we use select now
https://github.com/danskernesdigitalebibliotek/dpl-react/pull/225

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,72 @@
...
+      <select
+        className={classes.select}
+        aria-label={ariaLabel}
+        onChange={handleOnChange ? (e) => handleOnChange(e) : undefined}
+      >
+        {placeholder && (
+          <option hidden selected>
+            {placeholder}
+          </option>
+        )}
```
kasperg, line 49: Please reconsider the `hidden` and `selected` attributes.

1. As far as I can tell [`hidden` is not a valid attribute for `<option>` elements](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/option). Do we intend it to be `disabled`?
2. The placeholder element should only be `selected` if no other element is selected. Only a single element may be selected if the select element does use `multiple`.

kasperg, line 49: This is likely to fail accessibility tests because [the element does not have an accessible name but instead relies on a value to communicate purpose](https://accessibilityinsights.io/info-examples/web/select-name/).

I do not have a solution at hand at the moment but I think this will become a problem. At least we now have a generic component to solve it in.

kasperg (changes requested):
Good stuff here but there are also some major issues we need to look into - maybe together 😄 .

src/components/facet-line/FacetLineFilters.tsx
```diff
@@ -0,0 +1,125 @@
...
+  const t = useText();
+  const dispatch = useDispatch();
+
+  const formatValuesToDropdown = (facet: string, values: FacetValue[]) => {
+    return values.map((value) => {
+      return {
+        label: value.term,
+        value: value.term,
+        disabled: Boolean(filters?.[facet]?.[value.term]),
+        selected: Boolean(filters?.[facet]?.[value.term])
```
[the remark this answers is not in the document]
kasperbirch1: I don't know how I can achieve the desired appearance and functionality without them being both disabled and selected? 

if only uses the select
Skærmbillede 2022-11-17 kl  17 47 48

if I use both
Skærmbillede 2022-11-17 kl  17 43 57

And if I don't have both of them it won't switch back to placeholder if a term is removed

src/components/facet-browser/intelligent-facets-search.graphql
```diff
@@ -0,0 +1,12 @@
+query intelligentFacets ($q: SearchQuery!, $facetsLimit: Int!, $valuesLimit: Int!) {
```
kasperg, line 1: Please move this file so it is co-located with there it is used.

This file is placed in the `/src/components/facet-browser/` directory but it is used by code in the `
src/components/facet-line/` directory.

src/components/facet-line/FacetLine.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+const FacetLine: React.FunctionComponent<FacetLineProps> = ({
+  q,
+  filters,
+  filterHandler
+}) => {
+  const { data } = useIntelligentFacetsQuery({
+    q: { all: q },
+    facetsLimit: 5,
+    valuesLimit: 5
```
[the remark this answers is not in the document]
kasperg: Please update `facetsLimit` to 6 as requested.

src/apps/search-result/facet-browser.test.ts
```diff
@@ -1,39 +1,118 @@
...
-    cy.contains("Main languages");
-    cy.contains("Access types");
-    cy.contains("Children or adults");
-    cy.contains("Creators");
-    cy.contains("Fiction or nonfiction");
-    cy.contains("Genre and form");
-    cy.contains("Material types");
-    cy.contains("Subjects");
-    cy.contains("Work types");
+    cy.get('[aria-controls="facet-mainLanguages"]')
```
kasperg, line 54: Consider if this should use `cy` data attributes instead of `aria-controls` like we do elsewhere.

cypress/fixtures/search-result/facet-browser/intelligentFacets.json
```diff
@@ -0,0 +1,59 @@
+{
```
kasperg, line 1: Please move this file so it is co-located with there it is used.

This file is placed in the /cypress/fixtures/facet-browser/ directory but it is used by code in the `facet-line` tests.

kasperg (commented):
:+1: I think this is looking good but there are some inconsistencies about how elements are selected with/without `cy` data attributes.

To show that our tests run and pass without rebasing/merging the entirely of #237 I suggest cherry picking https://github.com/danskernesdigitalebibliotek/dpl-react/pull/237/commits/e0a31f1d7ef35ef5855291695401463dd176fb3a which adds the necessary commands.

jeppekroghitk (approved):

PR #214, opened 2022-11-07
Feature/5379 reservation list pagination and empty list

sinejespersen:
#### Link to issue

#### Description

#### Screenshot of the result

#### Checklist


src/components/result-pager/result-pager.tsx
```diff
@@ -21,7 +32,7 @@ function ResultPager({
       </p>
       <button
         type="button"
-        className="btn-primary btn-outline btn-medium arrow__hover--right-small"
+        className="test-result-pager btn-primary btn-outline btn-medium arrow__hover--right-small"
```
spaceo, line 35: Test?

src/components/result-pager/result-pager.tsx
```diff
@@ -12,6 +12,17 @@ function ResultPager({
...
+  if (itemsShown >= hitcount) {
+    return (
+      <div className="result-pager">
+        <p className="text-small-caption result-pager__title">
+          {t("showingText")} {hitcount} {t("outOfText")} {hitcount}
+        </p>
+      </div>
+    );
+  }
+
```
spaceo, line 25: I dont understand this part. How can you show more than there is in total? Originally, when I made the first version, I simply hid the pager if there was less items on the page than the pageSize.

spaceo (commented):
Some comments

src/components/result-pager/result-pager.tsx
sinejespersen: So Imma change this in another pr, and just... fix it everywhere (https://docs.cypress.io/guides/references/best-practices)

src/components/result-pager/result-pager.tsx
sinejespersen: I've added a comment, I've kept the part that removes the button in the pagination when it is not necessary, and I am setting ```itemsShown``` differently

sinejespersen: I've fixed this, it still doesnt show the button if there is no reason to click it/no more list items, I dont remember if you had any strong opinions on this matter

PR #215, opened 2022-11-07
Feature/ general functionality for statistics implementation

Adamik10:
#### Link to issue
[DDFSOEG-268](https://reload.atlassian.net/browse/DDFSOEG-268)

#### Description
This PR Introduces useStatistics() custom hook that can be used throughout the project to contact Mapp web tracking. The tracking works in two different ways depending on the environment. 
- In production, the app is mounted into drupal CMS system where a script is loaded creating a globally accessible wts object - used for sending tracking information.
- In development, the app creates its own custom wts object (since the "real" one from Mapp doesn't exist) that logs tracked data into the console instead of sending it to Mapp. 

#### Screenshot of the result
This screenshot is taken from the Google Chrome developer tools -> network tab showcasing our tracking event that is being sent with desired data.
image

#### Checklist

Please ignore the 3 failing integration tests:
1 of them doesn't exist on this branch at all and the other two show a green result locally - which means the fault is in release/4 branch.


src/core/statistics/useStatistics.ts
```diff
@@ -0,0 +1,24 @@
+export function useStatistics() {
+  // If the global wts object doesn't exist, it means we are in dev environment.
+  // Here instead of actually tracking we just log the data to the console.
+  if (!window.wts) {
+    window.wts = {
+      push(trackingProps: [string, string, { [key: string]: string }]) {
+        // eslint-disable-next-line no-console
+        console.log(
+          `Tracking: ${trackingProps[0]}, ${trackingProps[1]}, ${JSON.stringify(
+            trackingProps[2]
```
kasperg, line 10: :+1:

src/types/index.d.ts
```diff
@@ -0,0 +1,8 @@
+declare global {
+  interface Window {
+    wts: {
+      push(props: [string, string, { [key: string]: string }]): void;
```
kasperg, line 4: Please improving the documentation and specification for this.

Examples of better specification:
- The initial string should always be "send"
- The second string should be either "page", "link", "click" or "pageupdate"
- The third string is not simply an object with string keys and values. Values may be objects in itself.

Linking to sources like the following would be helpful:

- https://documentation.mapp.com/1.0/en/manual-track-request-25105181.html
- https://documentation.mapp.com/1.0/en/how-to-send-manual-tracking-requests-page-updates-7240681.html

src/core/statistics/useStatistics.ts
```diff
@@ -0,0 +1,24 @@
...
+          )}`
+        );
+      }
+    };
+  }
+
+  return {
+    track: (trackedData: { [key: string]: string }) => {
+      window.wts.push(["send", "page", trackedData]);
+    }
```
kasperg, line 20: Consider returning multiple functions with different signatures supporting different behavior.

There may be different behavior and specifications for page/update and click/link tracking.

src/apps/search-result/search-result.tsx
```diff
@@ -49,11 +48,19 @@ const SearchResult: React.FC<SearchResultProps> = ({ q, pageSize }) => {
...
     setResultItems([]);
   }, [q, pageSize, filters]);
 
+  // Tracking whenever the search query changes.
+  const { track } = useStatistics();
+  useEffect(() => {
+    track({ internalSearch: q });
+    // We actaully just want to track if the query changes.
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [q]);
```
kasperg, line 62: Based on our discussion today this may not be the actual implementation of search tracking that we want to use.

I do not know whether we should let it stay in the current form or await more clarification.

src/types/index.d.ts
Adamik10: Very good thinking. Thanks! 🧠 💡

src/apps/search-result/search-result.tsx
Adamik10: I adjusted it, but I think we should keep it for testing 😉

src/core/statistics/useStatistics.ts
Adamik10: I adjusted this, but not into returning multiple functions, because I believe we can achieve all we want with the setup. Let's see if it if it works.

src/core/statistics/useStatistics.ts
```diff
@@ -0,0 +1,52 @@
...
+export type EventData = {
+  linkId: string;
+  customClickParameter: CustomClickParameter;
+};
+
+// 1. page - when a page request is simulated
+// 2. click - for measuring actions that don't cause page load;
+// 3. link - clicking a link that triggers a new page load
+// 4. pageupdate - information on the page changes without a new page load
+type EventType = "page" | "click" | "link" | "pageupdate";
```
kasperg, line 18: :+1: Excellent

src/core/statistics/useStatistics.ts
```diff
@@ -0,0 +1,52 @@
...
+// https://documentation.mapp.com/1.0/en/how-to-send-manual-tracking-requests-page-updates-7240681.html
+
+export type CustomClickParameter = {
+  [key: number]: number | { [key: number]: string };
+};
+
+export type EventData = {
+  linkId: string;
+  customClickParameter: CustomClickParameter;
+};
```
kasperg, line 12: Please rework this structure for event data. Based on [the examples provided by Mapp](https://documentation.mapp.com/1.0/en/manual-track-request-25105181.html) I see the following:

1. Neither `linkId` or `customClickParameter` are required as event data properties.
2. Many other properties are used.

[The documentation lists to so many different parameters](https://documentation.mapp.com/1.0/en/which-parameters-can-be-sent-to-mapp-intelligence-7211946.html) that I do not think it is relevant for us to limit them. To me an object with properties would be sufficiently specific.

src/core/statistics/useStatistics.ts
```diff
@@ -0,0 +1,52 @@
...
+      id: number,
+      name: string,
+      trackedData: string | number | string[]
+    ) => {
+      const eventData: EventData = {
+        linkId: name,
+        customClickParameter: {}
+      };
+      eventData.customClickParameter[id] = trackedData;
+      window.wts.push(["send", eventType, eventData]);
```
kasperg, line 47: Without specific hooks per event I do not think this signature works well.

As mentioned previous `linkId` and `customClickParameter` are not required across event types.

I do not think it is helpful to specify the event type when invoking the hook either. It might as well be sent when invoking the `track()` method.

I suggest something along the following signature for the track method:

```typescript
export type Parameters = {
  [key: string]: number | string | { [key: number]: string | number };
};

function track(eventType: EventType, parameters: Parameters) {}
```

(Note that this is totally untested)

kasperg (changes requested):
:+1: on improving documentation but I think your typing/restructuring of the code needs more work.

src/core/statistics/useStatistics.ts
Adamik10: Good point! I generalized it to recognize customClickParameter object, but also not limit the EventData type to it.

src/core/statistics/useStatistics.ts
Adamik10: I moved specifying the event type down to the track() function, but specified the parameters more strictly than you describe. With the current setup we assume that we will be needing specifically the id, name, and trackedData, which also needs to be sent to Mapp in a very specific format.

src/core/statistics/useStatistics.ts
```diff
@@ -0,0 +1,60 @@
+// Useful resources for Mapp tracking:
+// https://documentation.mapp.com/1.0/en/manual-track-request-25105181.html
+// https://documentation.mapp.com/1.0/en/how-to-send-manual-tracking-requests-page-updates-7240681.html
+
+export type CustomClickParameter = {
+  [key: number]:
+    | number
+    | { [key: number]: string }
+    | { [key: string]: string[] };
+};
```
spaceo, line 10: Note that is also possible to use the TS utility: `Record`:
```
 type CustomClickParameter = Record<
    string,
    number | Record<number, string> | Record<string, string[]>
  >;

File: src/core/statistics/useStatistics.ts [hunk 9]
```diff
@@ -0,0 +1,60 @@
...
+  name: string;
+  trackedData: string | number | string[];
+};
+
+export function useStatistics() {
+  // If the global wts object doesn't exist, it means we are in dev environment.
+  // Here instead of actually tracking we just log the data to the console.
+  if (!window.wts) {
+    window.wts = {
+      push(trackingProps: ["send", EventType, EventData]) {
```
spaceo, line 36: Why not also creating a type for the first parameter, like the second and third?
Could for instance be called: `Action`

src/apps/search-result/search-result.tsx
```diff
@@ -58,11 +59,18 @@ const SearchResult: React.FC<SearchResultProps> = ({ q, pageSize }) => {
...
   useEffect(() => {
     setResultItems([]);
   }, [q, pageSize, filters]);
 
+  const { track } = useStatistics();
+  useEffect(() => {
+    track("page", { id: 10, name: "On site search string", trackedData: q });
+    // We actaully just want to track if the query changes.
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [q]);
```
spaceo, line 72: The only other dependency here would be the track function. Why not include that? Is it varying?

Adamik10: Yes - first it's not there and then it is which would call it twice instead of once.

src/core/statistics/useStatistics.ts
Adamik10: Good point, why not! :) We now have a type for it 😉

src/core/statistics/useStatistics.ts
Adamik10: Yeah is that new? I keep seeing it since yesterday or two days ago. Or maybe I'm just late to the party 😝
Thanks though, will use it ;)

kasperg (approved):
:+1: LGTM

sinejespersen (approved):

PR #216, opened 2022-11-07
MaterialType change to "tidsskrift" instead of "periodikum". + Test split

kasperbirch1:
#### FBI API has been changed so that materialType is now called "tidsskrift" instead of "periodikum"
 
**Other than that**
in an attempt to find out why the periodical test fails. have we discovered that in cypress you can use wait(alias) to be sure that it does not continue until the fixed data is loaded

in addition, to making it easier to read and debug. The Material test is now divided so that there are separate tests for infomedia and periodicals.

#### Checklist

### DIFF
--- a/cypress/fixtures/material/availability.json
+++ b/cypress/fixtures/material/availability.json
@@ -1,6 +1,6 @@
 [
   {
-    "available": false,
+    "available": true,
     "recordId": "52557240",
     "reservable": true,
     "reservations": 2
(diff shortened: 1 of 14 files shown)


kasperg (changes requested):

kasperbirch1:
It has run 5 times without failing the periodical test, but the 4 times it failed the campaign test

Skærmbillede 2022-11-08 kl  13 30 52
Skærmbillede 2022-11-08 kl  13 32 27

kasperg:
Continued in #228.

PR #218, opened 2022-11-08
Feature/statistics - rest

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFSOEG-281

#### Description
This PR uses previously implemented useStatistics() custom hook to track 16 desired instances:
- On site search query 
- Search result count
- Material type
- Material genre
- Material language
- Material source
- Material target audience
- Material topic number (DK5-number)
- Material fiction/nonfiction
- Material status
- Search result click
- Autosuggest click
- Reservation of a physical material
- Search facet values
- Campaign click
- Campaign shown

#### Screenshot of the result
n/a

#### Checklist

n/a


src/apps/search-header/search-header.tsx
```diff
@@ -233,6 +240,11 @@ const SearchHeader: React.FC = () => {
       const highlightedCategoryIndex =
         highlightedIndexAfterClick - (textData.length + materialData.length);
       const selectedItemString = determineSuggestionTerm(changes.selectedItem);
+      track("click", {
+        id: 54,
+        name: "Autosuggest - klik",
+        trackedData: selectedItemString
+      });
       redirectTo(
```
kasperg, line 248: Be aware of race conditions between tracking and redirects.

Tracking an event is an asynchronous action and we cannot be certain that it will complete before the redirect is issued. I can think of a few ways to handle this:

1. Checking if the analytics provide supports some kind of callback handling which will fire when tracking has completed (I do not think this is the case for Mapp)
2. Checking if the analytics provider uses [JavaScript APIs which bypass the problem](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/sendBeacon)
3. Registering the event on the target page. This is typically done by checking for parameters in the fragment/hash part of the URL when loading the target page. This adds some complexity to the code.
4. Waiting a bit before issuing the redirect to allow the tracking request to complete. This has the downside of making the UI appear slower than it need be.

There may be other approaches for this too. Consider asking around for more input here. 

I tried searching for this but the best I could find was [a 7 year old Stack Overflow post](https://stackoverflow.com/a/32897826).

src/apps/search-result/search-result.tsx
```diff
@@ -58,11 +60,18 @@ const SearchResult: React.FC<SearchResultProps> = ({ q, pageSize }) => {
...
   useEffect(() => {
     setResultItems([]);
   }, [q, pageSize, filters]);
 
+  const { track } = useStatistics();
+  useEffect(() => {
+    track("click", { id: 10, name: "On site search string", trackedData: q });
+    // We actaully just want to track if the query changes.
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [q]);
```
kasperg, line 73:
```
    // We actaully just want to track if the query changes.
    // eslint-disable-next-line react-hooks/exhaustive-deps
  }, [q]);
```

This gets a bit repetitive to look at.

I wonder if it would be a problem to just include `track` in the list of dependencies. Another option might be to move `const { track } = useStatistics();` into the `useEffect()` hook?

src/apps/search-result/search-result.tsx
```diff
@@ -124,6 +133,32 @@ const SearchResult: React.FC<SearchResultProps> = ({ q, pageSize }) => {
 
   const worksAreLoaded = Boolean(resultItems.length);
 
+  useEffect(() => {
+    if (isFirstLoad) {
+      setIsFirstLoad(false);
+      return;
+    }
```
kasperg, line 140: Please document why this is needed.

Do we not want to track the number of results for the initial search result?

src/components/atoms/link.tsx
```diff
@@ -1,27 +1,46 @@
...
+
   return (
+    // If we want to track this click we have to first NOT set the href attribute
+    // & instead redirect through an onclick handler
+    // eslint-disable-next-line jsx-a11y/anchor-is-valid
     <a
       id={id}
-      href={String(href)}
+      // eslint-disable-next-line no-script-url
+      href={trackClick ? "javascript:void(0)" : String(href)}
```
kasperg, line 32: Please reconsider this.

[The `jsx-a11y` documentation for this linting rule](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/docs/rules/anchor-is-valid.md) suggests alternative solutions which improve accessibility.

src/components/availability-label/availability-label.tsx
```diff
@@ -39,6 +40,16 @@ export const AvailabilityLabel: React.FC<AvailabilityLabelProps> = ({
...
+
+  // Track material availability (status) if the button is active - also meaning
+  // it is displayed on the material page
+  if (selected) {
+    track("click", {
+      id: 38,
+      name: "Materiale Status",
+      trackedData: availabilityText
+    });
+  }
```
kasperg, line 52: Please consider how this handles rerenders.

A - sometimes involuntary - rerender of a selected availability label will cause a new tracking request. That does not seem like a good idea.

This issue may also be related to our currently high number of availability labels per work.

src/components/campaign/Campaign.tsx
```diff
@@ -1,16 +1,31 @@
...
 import { LinkNoStyle } from "../atoms/link-no-style";
 import CampaignBody from "./CampaignBody";
 
 export interface CampaignProps {
   campaignData: CampaignMatchPOST200Data;
 }
+
+// In order to see campaigns in development mode, you will most likely need a
+// browser plugin such as Chrome's "Allow CORS: Access-Control-Allow-Origin"
+// in order to bypass CORS policy for dpl-cms data calls
```
kasperg, line 14: I understand what you are getting at here but I do not think this is the right place to document this.

I would suggest adding a markdown file describing how to go about working with DPL CMS if you do not have it running locally in [a separate markdown file as a part of the documentation](https://github.com/danskernesdigitalebibliotek/dpl-react/tree/develop/docs).

If you need more inspiration we also have [a lot of similar docs in the CMS project](https://github.com/danskernesdigitalebibliotek/dpl-cms/tree/develop/documentation).

src/components/facet-browser/FacetBrowserModalBody.tsx
```diff
@@ -30,6 +31,25 @@ const FacetBrowserModalBody: React.FunctionComponent<
...
+      return;
+    }
+    track("click", {
+      id: 20,
+      name: "Søgning Facet",
+      trackedData: JSON.stringify(
+        mapValues(filters, (filter) => {
+          return Object.keys(filter);
+        })
+      )
```
kasperg, line 48: Consider if using `JSON.stringify` is the right way to track facets.

When we have multiple values in other situations we join array values with `, `. I wonder why we do not use the same approach here.

src/components/facet-browser/FacetBrowserModalBody.tsx
```diff
@@ -30,6 +31,25 @@ const FacetBrowserModalBody: React.FunctionComponent<
       setOpenFacets([...openFacets, facet]);
     }
   };
+  const { track } = useStatistics();
+
+  useEffect(() => {
```
kasperg, line 36: Consider if this should use `useDeepCompareEffect`.

This seems like a `useEffect` which depends on a non-primitive data structure which may point to different values but having equal values.

src/components/material/MaterialHeader.tsx
```diff
@@ -78,6 +80,34 @@ const MaterialHeader: React.FC<MaterialHeaderProps> = ({
...
+    string[]
+  >([]);
+
+  useEffect(() => {
+    const newManifestationMaterialTypes = manifestation.materialTypes.map(
+      (item) => item.specific
+    );
+    if (isEqual(manifestationMaterialTypes, newManifestationMaterialTypes)) {
+      return;
+    }
```
kasperg, line 96: Consider creating the list of material types outside of the hook and then using the list as a dependency of the hook.

This should make the code less complex.

Also note that you may need `useDeepCompareEffect()` here.

kasperg (changes requested):
:+1: This really gets around throughout our components.

I think there a some issues that you need to look into before this can be passed along.

src/apps/material/material.tsx
```diff
@@ -35,23 +35,56 @@ import {
...
-
   const { data, isLoading } = useGetMaterialQuery({
     wid
   });
+  const { track } = useStatistics();
+  useEffect(() => {
+    if (data?.work?.genreAndForm) {
+      track("click", {
+        id: 25,
+        name: "Materiale Genre",
```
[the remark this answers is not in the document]
Adamik10: We could have an array of objects in a separate file that stores all this info. Something like:
```
[
     {
     id: 20
     name: "Materiale Genre"
     }, {}, {}, ...
]```

File: src/apps/search-result/search-result.tsx (hunk 2 above)
Comment (Adamik10) [reply src/apps/search-result/search-result.tsx:73]: This is what it looks like with `track` included as a dependency 😅 :
image

File: src/apps/search-result/search-result.tsx (hunk 3 above)
Comment (Adamik10) [reply src/apps/search-result/search-result.tsx:140]: well no, because the initial search result is always 0. It is set on line 48 with ` const [hitcount, setHitCount] = useState<number>(0);`

File: src/apps/material/material.tsx (hunk 10 above)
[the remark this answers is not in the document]
Comment (Adamik10) [reply src/apps/material/material.tsx]: ... and that's also how I implemented it now :) we use this general object all around the apps

File: src/components/availability-label/availability-label.tsx (hunk 5 above)
Comment (Adamik10) [reply src/components/availability-label/availability-label.tsx]: This should be fixed now - I introduced a useEffect() here.

File: src/components/campaign/Campaign.tsx (hunk 6 above)
Comment (Adamik10) [reply src/components/campaign/Campaign.tsx]: Thanks, I made a separate .md file and a separate PR for it 😉

File: src/components/facet-browser/FacetBrowserModalBody.tsx (hunk 7 above)
Comment (Adamik10) [reply src/components/facet-browser/FacetBrowserModalBody.tsx]: It is because there are multiple facet categories that can be tracked here AND they each can have multiple values. So in order for them to make sense of it it needs to be structured data we send. Well.. or at least in this case stringified structured data.
But I could totally join the separate values into one string! Instead of an array. DIdn't think of that 🤓

File: src/components/facet-browser/FacetBrowserModalBody.tsx (hunk 8 above)
Comment (Adamik10) [reply src/components/facet-browser/FacetBrowserModalBody.tsx:37]: I went through the tracking around the apps and we now officially use useDeepCompareEffect 😉 
Had to merge develop branch in first though.

File: src/components/material/MaterialHeader.tsx (hunk 9 above)
Comment (Adamik10) [reply src/components/material/MaterialHeader.tsx]: That makes total sense. Thanks! :)

File: src/apps/search-header/search-header.tsx (hunk 1 above)
Comment (Adamik10) [reply src/apps/search-header/search-header.tsx]: Good to know that the request is asynchronous at any rate. But this was a difficult one for me to take a stand. There isn't any promise we can listen for in order to wait for the Mapp request to be done. 
But we do have a solution that DDF can consider now using timeout.

File: src/apps/search-result/search-result.tsx (hunk 3 above)
Comment (kasperg) [reply src/apps/search-result/search-result.tsx:144]: OK. Have you considering making the default value for the `hitcount` `null` instead of `0` and then checking for that instead? To me it seems preferable to adding another state.

File: src/components/atoms/link.tsx (hunk 4 above)
Comment (kasperg) [reply src/components/atoms/link.tsx]: Is there a reason this is not a proper `button` element instead of `<span role="button">`?

It seems as if [this would be preferable to the custom onClick + onKeyUp handling](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/button_role#sect1).

Comment (kasperg) [reply src/components/atoms/link.tsx]: Also remember to pass the `className` attribute to the button.

Comment (kasperg) [reply src/components/atoms/link.tsx]: Also remember to support the `target` attribute. This may require you to use [`Window.open()`](https://developer.mozilla.org/en-US/docs/Web/API/Window/open).

File: src/apps/search-header/search-header.tsx (hunk 1 above)
Comment (kasperg) [reply src/apps/search-header/search-header.tsx]: It would be preferable if our own `track()` function could return a promise which resolved after a specific time.

This would mean that we could track like so:

```
track({"some": "data"}).then(() => {
  redirectTo();
})
```

It is a more elegant solution and would mean less duplication.

src/components/atoms/link.tsx
kasperg: Also remember that the tracking function should take the waiting into account so tracking will occur before the browser is redirected.

kasperg (commented):
:+1: There are a some outstanding issues regarding link tracking and outstanding which still needs to be investigated.

src/apps/search-result/search-result.tsx
Adamik10: That sounds very reasonable! Haven't thought of that, but now that's the way we do it :)

src/apps/search-header/search-header.tsx
Adamik10: I agree and we now walk down this path.

src/components/atoms/link.tsx
Adamik10: @kasperg I tried to make it a button, and then when I did I saw the reason why I didn't make it a button in the first place. It's because it was never meant to be one (see picture) so the styling is all over the place.
image
So if we do make it a button then we first have to go back to the design system and accommodate for that. It's not a problem at all, it will just consume extra time. And considering we want to present this to DDF on Thursday at noon, I'd say we rather utilize a custom role="button" span.
I'm open to your suggestion if you feel strongly about it, pls let me know back ;)

kasperg (approved):
:+1: LGTM

jeppekroghitk (approved):

PR #221, opened 2022-11-11
Fix publication year

kasperbirch1:
#### The FBI API has changed the location of publicationYear
it has moved from being a child of manifestation to being in manifestation.edition.publicationYear. this means that I have had to correct the search query, fixtures, and rewrites in all components that touch publicationYear.

#### Checklist

It is especially this commit where I made logical corrections that you should pay attention to
https://github.com/danskernesdigitalebibliotek/dpl-react/pull/221/commits/30062e7d7b1ff3f07c4a72c585470e1bfb11540f

### DIFF
--- a/src/core/utils/helpers/general.ts
+++ b/src/core/utils/helpers/general.ts
@@ -12,13 +12,19 @@ import { FaustId, Pid } from "../types/ids";
 import { getUrlQueryParam } from "./url";
 import { LoanType } from "../types/loan-type";
 
+export const getManifestationPublicationYear = (
+  manifestation: Manifestation
+): string | null => {
+  return manifestation.edition?.publicationYear?.display || null;
+};
+
 export const orderManifestationsByYear = (
   manifestations: Manifestation[],
   order: "asc" | "desc" = "desc"
 ) => {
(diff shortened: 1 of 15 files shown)


src/core/utils/helpers/general.ts
```diff
@@ -75,7 +75,8 @@ export const getFirstPublishedManifestation = (
 
 export const getFirstPublishedYear = (manifestations: Manifestation[]) => {
   return String(
-    getFirstPublishedManifestation(manifestations)?.publicationYear.display
+    getFirstPublishedManifestation(manifestations)?.edition?.publicationYear
```
spaceo, line 78: I think we could make a function called getManifestationPublicationYear that receives a manifestation and returns: `string | null`
So: 

```ts
getManifestationPublicationYear(manifestation: Manifestation): string | null
```
I am not fan og the long optional chaining and by having it done in one place, we only need to change that function if/when it changes again.

spaceo (changes requested):
I think we need to centralize the way of resolving the publication date of a manifestation

spaceo (approved):

sinejespersen (approved):

PR #222, opened 2022-11-11
Usetext vitest unit test

spaceo:
#### Description

The code base is growing and so does the number of functions and custom hooks.
I missed unit testing while developing the extended text functionality and expansion of the useText hook.

So this is a suggestion for introducing unit tests by using [Vitest](https://vitest.dev/) and [testing-library/react-hooks](https://github.com/testing-library/react-hooks-testing-library) for hook testing.

I started out with Jest but had problems with having both Jest and Cypress in the same codebase. They have colliding test function names and Typescript could not figure it out.

I got Vitest recommended and it seems very fast and just as capable as Jest.

#### Screenshot of the result

image

#### Checklist

This is first iteration of bringing unit tests into our setup.
Next step is to bring it into our CI workflow.


tests/unit/text.test.tsx
```diff
@@ -0,0 +1,60 @@
...
+      }
+    }
+  }
+});
+
+const Wrapper = ({ children }: { children: ReactNode }) => (
+  <Provider store={store}> {children} </Provider>
+);
+
+test("Should render expected text strings with placeholders and plural handling", () => {
```
Adamik10, line 28: This test needs to be ran manually, right? because we don't have any setup to run it with Github actions.

Adamik10 (approved):
Overall I think the solution is very elegant and I can't wait to use it myself! :)
I left a couple questions and comments, but nothing I could see that would be blocking. Good job!

docs/ui_text_handling.md
```diff
@@ -61,3 +61,84 @@ const HelloWorld: React.FC = () => {
...
+```
+
+### Plural example
+
+Some times you want two versions of a text be shown
+depending on if you have one or multiple items being referenced in the text.
+
+That can be accommodated by using the plural text definition.
+
+Let's say that the authenticated user has a list of unread message in an inbox.
```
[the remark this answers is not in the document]
spaceo: Thanks for the corrections! Just a friendly note: You can also mark these language corrections as suggestions, then it is even easier to correct @Adamik10

src/core/utils/text.tsx
```diff
@@ -92,23 +92,43 @@ const constructTextDefinitionFromRawTextTextEntry = (
...
 
-  return (key: string, { placeholders, count } = {}) => {
+  return (key: string, { placeholders, count } = { count: 0 }) => {
     const textDefinition = constructTextDefinitionFromRawTextTextEntry(
       data?.[key] ?? key
     );
-    const processedTexts = placeholders
-      ? processTexts(textDefinition.text, placeholders)
+
+    const textPlaceholders = { ...(placeholders ?? {}) };
```
sinejespersen, line 115: Isnt this already in develop?

tests/unit/text.test.tsx
```diff
@@ -0,0 +1,74 @@
...
+      }
+    }
+  }
+});
+
+const Wrapper = ({ children }: { children: ReactNode }) => (
+  <Provider store={store}> {children} </Provider>
+);
+
+test("Should strings with placeholders", () => {
```
sinejespersen, line 28:
```suggestion
test("Should render strings with placeholders", () => {
```

package.json
```diff
@@ -69,6 +70,7 @@
     "@storybook/manager-webpack5": "^6.5.0-beta.7",
     "@storybook/preset-typescript": "^3.0.0",
     "@storybook/react": "^6.5.0-beta.7",
+    "@testing-library/react-hooks": "^8.0.1",
```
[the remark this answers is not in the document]
spaceo: Yes. Well, it is a library for testing hooks in cojcuntion with Vitest.

tests/unit/text.test.tsx
spaceo: Check

src/core/utils/text.tsx
spaceo: Should be

package.json
```diff
@@ -34,6 +34,7 @@
     "start:storybook:test": "NODE_ENV=test start-storybook --port 57021 --quiet --ci",
     "start:storybook:dev": "NODE_ENV=development  start-storybook --port 80",
     "dev": "concurrently --raw \"yarn start:storybook:dev\" \"yarn watch\"",
+    "test:unit": "vitest",
     "test": "cypress run",
```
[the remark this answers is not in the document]
spaceo: I'll keep it like this for now. Since it has been OK'ed that we introduce Vitest for unit testing I would like to get it into our setup and start to write more tests. But soonish we should have configured our CI setup to make use of it also.

PR #223, opened 2022-11-13
Use deep compare effect

spaceo:
#### Link to issue

https://reload.atlassian.net/browse/DDFSOEG-316

#### Description
This is the actual fix for the problem described in this PR:
https://github.com/danskernesdigitalebibliotek/dpl-react/pull/219

#### Checklist


.eslintrc
```diff
@@ -68,7 +68,13 @@
     // We like to use arrow function syntax also for functional components.
     "react/function-component-definition": "off",
     // No complaints about missing trailing comma
-    "@typescript-eslint/comma-dangle": "off"
+    "@typescript-eslint/comma-dangle": "off",
+    "react-hooks/exhaustive-deps": [
+      "warn",
+      {
+        "additionalHooks": "useDeepCompareEffect"
```
kasperg, line 75: :+1: Good to get this in for future reference.

src/apps/search-result/search-result.tsx
```diff
@@ -59,11 +60,11 @@ const SearchResult: React.FC<SearchResultProps> = ({ q, pageSize }) => {
 
   // If q changes (eg. in Storybook context)
   //  then make sure that we reset the entire result set.
-  useEffect(() => {
+  useDeepCompareEffect(() => {
     setResultItems([]);
   }, [q, pageSize, filters]);
 
-  useEffect(() => {
+  useDeepCompareEffect(() => {
```
kasperg, line 67: 😮 - so there are actually already two places where this is needed.

Even better to get a fix for this in. Perhaps React Query has saved us unnecessary requests - but actually not unnecessary rerenders in other situations.

src/apps/loan-list/modal/renew-loans-modal-content.tsx
```diff
@@ -34,6 +34,10 @@ const RenewLoansModalContent: FC<RenewLoansModalContentProps> = ({
   const { mutate } = useRenewLoansV2();
   const { close } = useModalButtonHandler();
   const { itemsShown, PagerComponent } = usePager(loansModal.length, pageSize);
+  // TODO: Investigate if we can use useIntersection instead of useInView
+  // and then we can remove the react-hook-inview dependency.
+  // useIntersection documentation:
+  // https://github.com/streamich/react-use/blob/master/docs/useIntersection.md#useintersection
   const [ref, isVisible] = useInView({ threshold: 0 });
```
kasperg, line 41: To me this is almost a requirement to add `react/use`.

I see `react/use` as a potentially single and well-maintained source of hooks for us going forward. If it cannot replace the other location were we use a third-party hook then the value already seems limited and we might as well go with more specialized implementations with a smaller surface according to our needs.

kasperg (commented):
Please add more commentary about your (or our) thoughts about adding `react/use` going forward.

The reviewers at ITK do not have the same context as we do and in my opinion adding `react/use` instead of a more specialized solution also sets a direction for future use of third-party hooks.

To me we are on the verge of an ADR here.

src/apps/loan-list/modal/renew-loans-modal-content.tsx
spaceo: Ok changed it...

sinejespersen (approved):

PR #224, opened 2022-11-14
Feature/5377 loaner status menu

jeppekroghitk:
#### Link to issue

https://platform.dandigbib.org/issues/5377

#### Description

- add menu app: entry, dev, mount
- add menu navigation list
- add menu notification list
- add placeholder links
- add method for filtering loans already expired and loans soon to expire based on threshold defined in config
- cypress test of above

#### Screenshot of the result

Screenshot 2022-11-22 at 09 56 53

#### Checklist

### DIFF
--- a/src/apps/menu/menu-navigation-list/menu-navigation-list.tsx
+++ b/src/apps/menu/menu-navigation-list/menu-navigation-list.tsx
@@ -33,8 +33,7 @@ const MenuNavigationList: FC<MenuNavigationListProps> = ({
   };
   return (
     <>
-      {menuNavigationData.map(({ name, link, dataId }: MenuNavigationData) => {
-        return (
+      {menuNavigationData.map(({ name, link, dataId }: MenuNavigationData) => (
           <div className="link-filters__tag-wrapper">
             <a href={link} className="link-tag link-tag link-filters__tag">
               {name}
--- a/src/apps/menu/menu-navigation-list/menu-navigation-list.tsx
+++ b/src/apps/menu/menu-navigation-list/menu-navigation-list.tsx
@@ -1,0 +1,45 @@
+import React, { FC } from "react";
+
+interface MenuNavigationData {
+  name: string;
+  link: string;
+  dataId: string;
+}
+
+interface MenuNavigationListProps {
+  menuNavigationData: MenuNavigationData[];
+  loansCount: number;
+  reservationCount: number;
+  feeCount: number;
+}
+
+const MenuNavigationList: FC<MenuNavigationListProps> = ({
+  menuNavigationData,
+  loansCount,
+  reservationCount,
+  feeCount
+}) => {
+  const dataMap: { [key: string]: number } = {
+    "1": loansCount, // Loans
+    "2": reservationCount, // Reservations
+    "3": 0, // TODO: Get count of "Huskeliste"
+    "4": feeCount // Fees & Replacement costs
+  };
+  return (
+    <>
+      {menuNavigationData.map(({ name, link, dataId }: MenuNavigationData) => {
+        return (
+          <div className="link-filters__tag-wrapper">
+            <a href={link} className="link-tag link-tag link-filters__tag">
+              {name}
+            </a>
+            {dataMap[dataId] !== 0 && (
+              <span className="link-filters__counter">{dataMap[dataId]}</span>
+            )}
+          </div>
+        );
+      })}
+    </>
+  );
+};
+export default MenuNavigationList;
--- a/src/apps/menu/menu-navigation-list/menu-navigation-list.tsx
+++ b/src/apps/menu/menu-navigation-list/menu-navigation-list.tsx
@@ -19,10 +19,16 @@ const MenuNavigationList: FC<MenuNavigationListProps> = ({
   reservationCount,
   feeCount
 }) => {
+  /*
+  TODO: elaborate mapping of menu content and data-sources.
+  As the menu content from drupal and the data-sources
+  providing the number of items from diverse data endpoints are noncoherent, we need a way to
+  map x menu-content to y data-content.
+  */
   const dataMap: { [key: string]: number } = {
     "1": loansCount, // Loans
     "2": reservationCount, // Reservations
-    "3": 0, // TODO: Get count of "Huskeliste"
+    "3": 0, // TODO: Get count of "My List"
     "4": feeCount // Fees & Replacement costs
   };
   return (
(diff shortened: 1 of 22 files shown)


src/apps/menu/menu.test.ts
```diff
@@ -0,0 +1,341 @@
...
+    cy.visit("/iframe.html?path=/story/apps-menu--menu-entry");
+    cy.wait(["@reservations", "@physical_loans", "@digital_loans", "@fees"]);
+  });
+
+  it("Menu", () => {
+    // 2. Systemet viser Lånerstatusmenuen med
+    // 2.a. Brugerens navn
+    cy.get(".modal-header__name")
+      .should("exist")
+      .should("have.text", "Testkort ITK CMS Merkur");
```
sinejespersen, line 254:
```suggestion
      .and("have.text", "Testkort ITK CMS Merkur");
```

src/apps/menu/menu-navigation-list/menu-navigation-list.tsx
```diff
@@ -0,0 +1,45 @@
...
+  const dataMap: { [key: string]: number } = {
+    "1": loansCount, // Loans
+    "2": reservationCount, // Reservations
+    "3": 0, // TODO: Get count of "Huskeliste"
+    "4": feeCount // Fees & Replacement costs
+  };
+  return (
+    <>
+      {menuNavigationData.map(({ name, link, dataId }: MenuNavigationData) => {
+        return (
```
jeppekroghitk, line 31:
```suggestion
      {menuNavigationData.map(({ name, link, dataId }: MenuNavigationData) => (
```

spaceo (changes requested):
Looks good. No real blockers.
Except that I do not get the full concept of the menu rendering. Would be nice to have that clarified before I approve

src/apps/menu/menu.tsx
```diff
@@ -0,0 +1,207 @@
...
+                {t("menuLogOutText")}
+                <div className="ml-16">
+                  <ArrowIcon />
+                </div>
+              </button>
+            </Link>
+          </div>
+        </div>
+      </div>
+    </div>
```
spaceo, line 203: Where is it determined if a user is logged in or not?

src/apps/menu/menu-navigation-list/menu-navigation-list.tsx
```diff
@@ -0,0 +1,49 @@
...
+    "4": feeCount // Fees & Replacement costs
+  };
+  return (
+    <>
+      {menuNavigationData.map(({ name, link, dataId }: MenuNavigationData) => (
+        <div className="link-filters__tag-wrapper">
+          <a href={link} className="link-tag link-tag link-filters__tag">
+            {name}
+          </a>
+          {dataMap[dataId] !== 0 && (
```
spaceo, line 41: Could `dataMap[dataId]` be undefined?
Maybe you should check for falsey instead of being that strict?:
```typescript
 {dataMap[dataId] && (

Comment (jeppekroghitk) [reply src/apps/menu/menu-navigation-list/menu-navigation-list.tsx:41]: You're right, checking for falsey is the right thing to do here.
I did it because the data was fixed, it worked and i knew it has to be reworked at some point anyways.

File: src/apps/menu/menu.tsx [hunk 5]
```diff
@@ -0,0 +1,207 @@
...
+  }, [fbsFees]);
+
+  /*
+  TODO: Find a proper way to handle tabindex when this menu is active.
+  All content behind the overlay cannot be targetable
+  while the menu is open.
+  */
+
+  return (
+    <div className="modal-backdrop">
```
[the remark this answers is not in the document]
spaceo: We need to investigate if the markup can replaced with the `Modal` component.
You can look at the `FacetBrowserModal` for inspiration

src/apps/menu/menu.tsx
jeppekroghitk: It is not. I added the GuardedApp to the .entry file. Seems like the way to do it.
Altho, we are still awaiting answers to how the login is going to flow in this context, so it will be revisited nonetheless. 
TODO added.

src/apps/menu/menu.tsx
```diff
@@ -0,0 +1,207 @@
...
+}
+
+const Menu: FC = () => {
+  const t = useText();
+  const config = useConfig();
+  const { data: patronData } = useGetPatronInformationByPatronIdV2();
+  const { data: patronReservations } = useGetReservationsV2();
+  const { data: publizonData } = useGetV1UserLoans();
+  const { data: fbsData } = useGetLoansV2();
+  const { data: fbsFees } = useGetFeesV2();
```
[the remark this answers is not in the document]
jeppekroghitk: I think the cache will save us in this case. No data is updated from this view, so invalidation is not relevant.
I added loading state as a TODO, as we talked about this might be a global decision we have to make at some point.

src/apps/menu/menu.tsx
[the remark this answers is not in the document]
jeppekroghitk: Keeping the TODO here. I don't think that the overlay here is going to act as a regular Modal, so it will have to be revisitied at a later stage.

src/apps/menu/menu.test.ts
```diff
@@ -0,0 +1,341 @@
+import { TOKEN_LIBRARY_KEY } from "../../core/token";
+
```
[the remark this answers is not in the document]
jeppekroghitk: Yes, added TODO!

src/apps/menu/menu.tsx
spaceo: The GuardedApp hoc is used in conjuction with a [guardedRequest](https://github.com/reload/dpl-react/blob/09bb7f547fcb520cd293c526151eddce8eab8d92/src/components/material/MaterialHeader.tsx#L51-L59).

The flow is:
* An anonymous tries to do something that triggers a request that requires the user to be authenticated
* The request is persisted by using redux persist
* The user is redirected to Adgangsplatformen for login
* After successful authentication the user is redirected back to where we started
* Because the app is wrapped in the guardedApp hoc the app is being held back until the persisted request has been fired
* When the request have been executed the app is loaded and the user can see the effects in the UI of the action that originally was taken

So that is a very long way of saying:
I do not think you need the guarded app hoc here, unless I don't get the full picture.

spaceo (changes requested):
LGTM. Except that I do not think you need the guarded app. Would also be nice with some TODO comments of how the future authentication/authorization thoughts are - if possible.

src/apps/menu/menu.tsx
jeppekroghitk: Alright. Removed the GuardedApp. 
As this menu is accessible without having to reload the page, we need to check if the user i authorized.
I will visit this at a later stage when i know what this key will look like. Should be a pretty simple check to add at that point.

spaceo (approved):
LGTM

sinejespersen (approved):

PR #227, opened 2022-11-16
5411: simplify pagination

sinejespersen:
#### Link to issue

https://platform.dandigbib.org/issues/5411

#### Description

#### Screenshot of the result

#### Checklist


src/components/result-pager/use-pager.tsx
```diff
@@ -6,21 +6,21 @@ const usePager = (
...
-    setitemsShown(pageSize);
+    setItemsShown(pageSize);
   };
 
   const pagehandler = () => {
     const currentPage = page + 1;
     const itemsOnPage = (currentPage + 1) * pageSize;
+    // the "itemsOnPage > hitcount"-check is to
+    // To avoid the "showing 10 out of 8"-situation
+    setItemsShown(itemsOnPage > hitcount ? hitcount : itemsOnPage);
```
spaceo, line 22: I am still confused here. I have some questions:

Why do we need both `itemsOnPage`and `itemsShown`?
How do `itemsOnPage`and `itemsShown` differ?
Why is `setItmesShown` removed from the pageHandler.

I could probably find out the meaning behind this by debugging it.
Although I know that you easily can get errors in the realm of "off-by-one" working with this, I think it is possible to make a structure that is more simple and thereby easier to read.

sinejespersen: its mostly from here: https://github.com/danskernesdigitalebibliotek/dpl-react/blob/235f086fb638138414e87426e1500ea6b5ee2d84/src/apps/search-result/search-result.tsx#L24

items on page is to figure out if they are larger than hitcount. 
is setItemsShown removed? I renamed it, is that what you mean?

PR #232, opened 2022-11-17
Fix the "view online" button

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFSOEG-301

#### Description
This PR fixes an issue with the material page action. button. For online materials it should be showing:
- go to filmstriben (when the source is filmstriben)
- go to ereolen (when the source is ereol)
- listen online (when the material type is lydbog (net))
- see online (for the rest)

#### Screenshot of the result
image
image
image
image

#### Checklist

n/a

### DIFF
--- a/src/components/material/material-buttons/online/MaterialButtonOnlineExternal.tsx
+++ b/src/components/material/material-buttons/online/MaterialButtonOnlineExternal.tsx
@@ -1,4 +1,5 @@
 import React, { useState, FC, useEffect } from "react";
+import { MaterialType } from "../../../../core/dbc-gateway/generated/graphql";
 import { useProxyUrlGET } from "../../../../core/dpl-cms/dpl-cms";
 import { useText } from "../../../../core/utils/text";
 import { ButtonSize } from "../../../../core/utils/types/button";
@@ -47,7 +48,10 @@ const MaterialButtonOnlineExternal: FC<MaterialButtonOnlineExternalProps> = ({
     }
   }, [data, error, translatedUrl, urlWasTranslated]);
 
-  const label = (sourceName: string, materialTypes: string[]) => {
+  const label = (
+    sourceName: string,
+    materialTypes: MaterialType["specific"][]
+  ) => {
     if (sourceName.includes("ereol")) {
       return `${t("goToText")} ereolen`;
     }
--- a/src/components/material/material-buttons/online/MaterialButtonOnlineExternal.tsx
+++ b/src/components/material/material-buttons/online/MaterialButtonOnlineExternal.tsx
@@ -53,10 +53,10 @@ const MaterialButtonOnlineExternal: FC<MaterialButtonOnlineExternalProps> = ({
     materialTypes: MaterialType["specific"][]
   ) => {
     if (sourceName.includes("ereol")) {
-      return `${t("goToText")} ereolen`;
+      return t("goToText", { placeholders: { "@source": "ereolen" } });
     }
     if (sourceName.includes("filmstriben")) {
-      return `${t("goToText")} filmstriben`;
+      return t("goToText", { placeholders: { "@source": "filmstriben" } });
     }
     if (materialTypes.find((element) => element.includes("lydbog"))) {
       return t("listenOnlineText");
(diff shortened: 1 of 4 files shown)


src/components/material/material-buttons/online/MaterialButtonOnlineExternal.tsx
```diff
@@ -44,10 +47,27 @@ const MaterialButtonOnlineExternal: FC<MaterialButtonOnlineExternalProps> = ({
     }
   }, [data, error, translatedUrl, urlWasTranslated]);
 
+  const label = (sourceName: string, materialTypes: string[]) => {
+    if (sourceName.includes("ereol")) {
+      return `${t("goToText")} ereolen`;
+    }
+    if (sourceName.includes("filmstriben")) {
+      return `${t("goToText")} filmstriben`;
+    }
```
spaceo, line 56: You should make use of the placeholder functionality in the new text functionality. See readme ;)

src/components/material/material-buttons/online/MaterialButtonOnlineExternal.tsx
```diff
@@ -44,10 +47,27 @@ const MaterialButtonOnlineExternal: FC<MaterialButtonOnlineExternalProps> = ({
     }
   }, [data, error, translatedUrl, urlWasTranslated]);
 
+  const label = (sourceName: string, materialTypes: string[]) => {
```
spaceo, line 50: Is there a type from the generated graphql.tsx file we can use for material types?

spaceo (changes requested):
Great! Please note my comment regarding placeholders in texts.

src/components/material/material-buttons/online/MaterialButtonOnlineExternal.tsx
Adamik10: Yes indeed, good sir! I am adjusting it now :)

src/components/material/material-buttons/online/MaterialButtonOnlineExternal.tsx
Adamik10: Okay I shall :3 but those changes you introduced are not a part of this PR 🙈 But I guess that's alright, right? I'll just use it "as if" it already existed here.

Adamik10: Oh, I found it now... I think maybe you should mention the document in the readme file as well. That's where I looked 🙈

spaceo: Don't get anything of what you wrote but glad that you have fixed it.

spaceo:
Cool. It is accepted.
One thing you could consider is to introduce a function for deciding the online type (because we might need the concept of deciding the online type elsewhere in the future:
```typescript

export const getOnlineType = (
  sourceName: string,
  materialTypes: MaterialType["specific"][]
) => {
  if (sourceName.includes("ereol")) {
    return "ebook";
  }
  if (sourceName.includes("filmstriben")) {
    return "emovie";
  }
  if (materialTypes.find((element) => element.includes("lydbog"))) {
    return "audiobook";
  }
  return "unknown";
};
```
And then use that in your label function:
```
const label = (
  sourceName: string,
) => {
  const onlineType = getOnlineType(sourceName);
  switch (onlineType) {
    case "ebook":
    case "emovie":
      return t("goToText", { placeholders: { "@source": onlineType } });
    case "audiobook":
      return t("listenOnlineText");
    default:
      return t("seeOnlineText");
  }
};

```typescript
const label = (
  sourceName: string,
) => {
  const onlineType = getOnlineType(sourceName);
  switch (onlineType) {
    case "ebook":
    case "emovie":
      return t("goToText", { placeholders: { "@source": onlineType } });
    case "audiobook":
      return t("listenOnlineText");
    default:
      return t("seeOnlineText");
  }
};

spaceo (commented):
LGTM 👍

spaceo (approved):
Greatnezzzz!

sinejespersen (approved):

PR #233, opened 2022-11-17
Add close logic to facet browser showResults button

kasperbirch1:
### Add close logic to facet browser showResults button
it is now possible to press the showResults button to close the modal instead of the cross in the corner.

#### Checklist


kasperg (approved):
:+1: LGTM

jeppekroghitk (approved):
LGTM

PR #239, opened 2022-11-21
Document campaigns

Adamik10:
#### Description

This PR introduces documentation on campaigns - explaining the concept, usage and (in my opinion most importantly) how to get to show them in local development environmet. 

#### Checklist

n/a


docs/campaigns.md
```diff
@@ -0,0 +1,37 @@
...
+criteria for each campaign they create. The visibility criteria is based on
+search filter facets. Does that sound familiar? Yes, we use anoter API to get
+that very data in THIS project - in the search result app.
+The facets differ based on the search string the user uses for their search.
+
+As an example, the dpl-cms admin could wish to show a Harry Potter related
+campaign to all the users whose search string retreives search facets which
+have "Harry Potter" as one of the most relevant subjects.
+Campaigns in dpl-cms can use triggers such as subject, main language, etc.
+
```
spaceo, line 28: Would it be an idea to show an example of the hook being used, or are you afraid of that the documentation gets deprecated when/if the code changes?
This is an ancient problem that I do not have the golden answer to.

docs/campaigns.md
```diff
@@ -0,0 +1,37 @@
...
+## How campaign setup works in dpl-cms
+
+Dpl-cms is a cms system based on Drupal, where the system administrators can set
+up campaigns they want to show to their users. Drupal also allows the cms system
+to act as an API endpoint that we then can contact from our apps.
+The cms administratots can specify the content (image, text) and the visibility
+criteria for each campaign they create. The visibility criteria is based on
+search filter facets. Does that sound familiar? Yes, we use anoter API to get
+that very data in THIS project - in the search result app.
+The facets differ based on the search string the user uses for their search.
```
spaceo, line 22:
```suggestion
Dpl-cms is a cms system based on Drupal, where the system administrators can set
up campaigns they want to show to their users. Drupal also allows the cms system
to act as an API endpoint that we then can contact from our apps.

The cms administrators can specify the content (image, text) and the visibility
criteria for each campaign they create. The visibility criteria is based on
search filter facets.
Does that sound familiar? Yes, we use another API to get that very data
in THIS project - in the search result app.
The facets differ based on the search string the user uses for their search.
```

spaceo (changes requested):
Some typos needs fixing. Also consider if there should be a code example in  the documentation. I could imagine that it is a bit abstract for a person that does not know about the systems.

docs/campaigns.md
Adamik10: You know what, that's not a bad idea :) I included a snippet 😉

spaceo (approved):
LGTM

jeppekroghitk (approved):

PR #242, opened 2022-11-22
Add missing campaign interception in Cypress tests

spaceo:
#### Description

Since the campaign endpoint was introduced in search results
we need to have the requests intercepted in all tests that shows search results

#### Checklist


kasperg (approved):
:+1: LGTM.

Note that the tests are failing. Now that we have made considerable work to make them pass it would nice to have this rebased againsta the latest changes to show that things still work as expected.

sinejespersen (approved):

PR #244, opened 2022-11-23
Order digital copy

kasperbirch1:
#### Link to issue

#### Description

#### Screenshot of the result

#### Checklist


src/components/material/material-buttons/helper.ts
```diff
@@ -2,10 +2,16 @@ import { intersection } from "lodash";
...
+export const isDigitalArticleService = (manifest: Manifestation) =>
+  manifest.access.some(
+    ({ __typename }) => __typename === "DigitalArticleService"
+  );
+
 export const hasCorrectAccessType = (
   desiredAccessType: AccessTypeCode,
   manifest: Manifestation
 ) => {
+  if (isDigitalArticleService(manifest)) return true;
```
Adamik10, line 14: Please make this into a separate function that checks for correct access - access and accessType are two different things. 😉

Adamik10 (commented):
I found something you could look into :) See the comment, please.

src/components/material/digital-modal/DigitalModal.tsx
```diff
@@ -0,0 +1,69 @@
...
+        pid,
+        email
+      }
+    });
+  };
+
+  // Pre fill the email field with the patron's email
+  const { data: patronData } = useGetPatronInformationByPatronIdV2();
+  useEffect(() => {
+    if (patronData?.patron?.emailAddress) {
```
spaceo, line 40: What happens if the service is slow and the user starts typing?

src/components/material/digital-modal/DigitalModal.tsx
```diff
@@ -0,0 +1,69 @@
...
+      screenReaderModalDescriptionText={t(
+        "orderDigitalCopyModalScreenReaderModalDescriptionText"
+      )}
+      closeModalAriaLabelText={t(
+        "orderDigitalCopyModalCloseModalAriaLabelText"
+      )}
+    >
+      {isSuccess && <DigitalModalSuccess modalId={modalId} />}
+      {isError && <DigitalModalError modalId={modalId} />}
+      {!isSuccess && !isError && (
```
spaceo, line 57: This is a minor. But since we have two requests happening here you could consider, in order to make it clear to future devs, to alias the properties from the digital article hook with: "articleOrder*". Eg.: "articleOrderSuccess" or "articleOrderError" etc.

src/components/material/digital-modal/DigitalModalBody.tsx
```diff
@@ -0,0 +1,44 @@
...
+type DigitalModalBodyProps = {
+  email: string;
+  handleOnChange: Dispatch<SetStateAction<string>>;
+  handleSubmit: () => void;
+  isLoading: boolean;
+};
+
+const DigitalModalBody: React.FunctionComponent<DigitalModalBodyProps> = ({
+  email,
+  handleOnChange,
```
spaceo, line 15: Why is the handleOnchange used as a callback here? I think it would be prettier to have a local useState for textinput changed in the same component that renders the textinput.

src/components/material/digital-modal/DigitalModalError.tsx
```diff
@@ -0,0 +1,26 @@
+import React from "react";
```
spaceo, line 1: The only difference between the DigitalModalSuccess and the DigitalModalError is the description. I think you could just make one component that could be called DigitalModalFeedback or similar?

src/components/material/digital-modal/helper.ts
```diff
@@ -0,0 +1,15 @@
+import { DigitalArticleService } from "../../../core/dbc-gateway/generated/graphql";
+import { Manifestation } from "../../../core/utils/types/entities";
+
+export const createDigitalModalId = (digitalArticleIssn: string) =>
+  `digital-modal-${digitalArticleIssn}`;
```
spaceo, line 5: I think we should make a string literal type for the issn id. Like we for instance have for FaustId.

spaceo (commented):
LGTM. Woud like you to look at my comments though

src/components/material/digital-modal/DigitalModalBody.tsx
kasperbirch1: I have an email state in DigitalModal because I need to send the email with the mutate function from useGetPatronInformationByPatronIdV2. and I use useGetPatronInformationByPatronIdV2 to provide the state I use to decide whether it is DigitalModalFeedback or DigitalModalBody that is rendered. so I can't move it into DigitalModalBody

src/components/material/digital-modal/DigitalModal.tsx
kasperbirch1: That's a good question 😅 I would think that it will overwrite what the user is currently typing..

I think this will be a rare experience. I've tried throttling my network to slow 3g and it doesn't cause problems there

But let me know if you have a suggestion to improve it?

kasperbirch1: That's a good question 😅 I would think that it will overwrite what the user is currently typing..

I think this will be a rare experience. I've tried throttling my network to slow 3g and it doesn't cause problems there

But let me know if you have a suggestion to improve it?

src/components/material/digital-modal/DigitalModal.tsx
```diff
@@ -52,16 +58,19 @@ const DigitalModal: React.FunctionComponent<DigitalModalProps> = ({
...
+
+      {!articleOrderSuccess &&
+        !articleOrderError &&
+        typeof userEmail === "string" && (
+          <DigitalModalBody
+            userEmail={userEmail}
+            handleSubmit={orderDigitalCopy}
+            isLoading={articleOrderLoading}
+          />
+        )}
```
spaceo, line 73:
```suggestion
      {!articleOrderSuccess &&
        !articleOrderError &&
        userEmail !== null && (
          <DigitalModalBody
            userEmail={userEmail}
            handleSubmit={orderDigitalCopy}
            isLoading={articleOrderLoading}
          />
        )}
```

spaceo (approved):
Approved. But please look at:
https://github.com/danskernesdigitalebibliotek/dpl-react/pull/244#discussion_r1036155109

src/components/material/digital-modal/DigitalModalBody.tsx
```diff
@@ -0,0 +1,47 @@
...
+      buttonLabel={
+        isLoading
+          ? t("orderDigitalCopyButtonLoadingText")
+          : t("orderDigitalCopyButtonText")
+      }
+      disabledButton={isLoading}
+    >
+      <TextInput
+        type="email"
+        label="Email"
```
sinejespersen, line 38: t("email") ?

PR #245, opened 2022-11-24
Check if logged in to view informedia article

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFSOEG-325

#### Description
This PR introduces conditional login to view an infomedia article on material page. If the user is logged in, the modal will open, otherwise the user will get redirected to the login platform with a return url pointing to the same page the user left, but with the modal open. In storybook context, we redirect to the login story without the redirect functionality. 

#### Screenshot of the result
https://user-images.githubusercontent.com/28546954/204552189-e6748e56-3f23-4a22-a0b9-120ee4d9f7b3.mov

#### Checklist

As you can see from the screen recording attached, the button as of now doesn't switch to loading state upon click, so the interface may seem a little unresponsive. We could add it in a separate PR, or as an addition to this one.


src/core/utils/helpers/url.ts
```diff
@@ -123,3 +123,32 @@ export const removeQueryParametersFromUrl = (url: URL, parameter: string) => {
...
+    const storybookRedirect = `${origin}/?path=/story/sb-utilities-adgangsplatformen--sign-in`;
+    // We don't use redirectTo() because that would redirect inside the storybook iframe.
+    window.top.location.href = storybookRedirect;
+    return;
+  }
+  const { pathname, search } = returnUrl;
+  const localPathToReturnTo = `${pathname}${search}`;
+  const redirectUrl = appendQueryParametersToUrl(authUrl, {
+    "current-path": localPathToReturnTo
+  });
```
spaceo, line 149: We could create a global url function called eg.: `createAuthenticationRedirectUrl`

Adamik10: makes total sense! Roger 😉

spaceo (approved):
LGTM

sinejespersen (approved):

PR #247, opened 2022-11-25
Rewrite code to use the reusable dropdown component (NEW)

kasperbirch1:
#### Introduce the reusable Dropdown in places where we used select before

The changes are made to ensure that the Dropdowns remain consistent and make the code more readable

#### Checklist

### DIFF
--- a/src/components/Dropdown/Dropdown.tsx
+++ b/src/components/Dropdown/Dropdown.tsx
@@ -4,10 +4,8 @@ import DropdownIcon from "./DropdownIcon";
 
 export type Option = {
   label: string;
-  href?: string;
   disabled?: boolean;
   value: string;
-  selected?: boolean;
 };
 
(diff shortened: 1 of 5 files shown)


src/components/Dropdown/Dropdown.tsx
```diff
@@ -51,15 +51,25 @@ const Dropdown: React.FunctionComponent<DropdownProps> = ({
...
+
+        {options.map(({ label, value, disabled }) => {
           return (
             <option
               key={label}
               value={value}
               className={classes.option}
               disabled={disabled}
-              selected={selected}
+              selected={value === defaultValue}
```
kasperg, line 72: :+1: Now that you have `options` as an array of `Options` and `placeholder` as an optional `Option` you can clean up the code even further.

sinejespersen (approved):

PR #249, opened 2022-11-28
Remove goToEReolenText as a material app prop.

Adamik10:
#### Description
goToEReolenText is depricated since the introduction of the @source placeholder in the goToText. We no longer use it in any of the apps. 

#### Checklist


kasperg (approved):
:+1: LGTM

sinejespersen (approved):

PR #255, opened 2022-12-01
Add promo bar

kasperbirch1:
#### Add logic to find a reservation that is available before the selected

A new useAlternativeAvailableManifestation hook has been added. It receives a work and returns an alternative manifest with few or 0 reservations. 

if needed it can also just return a boolean isOtherManifestationPreferred.

it is used in the reservation modal under edition.  

#### PromoBar is a new reusable component 
PromoBar accepts the following props:
- type: decide which icon to use
- text: is the text that is displayed
- theme: decide which text and background color
- sticky: sets a negative margin that matches the gap used in .reservation-modal-list

The first place it should be used is together with reservation modal items to show information about faster reservation

#### Screenshot of the result
Skærmbillede 2022-12-02 kl  13 10 59 (3)
Skærmbillede 2022-12-02 kl  13 11 13 (3)

#### Checklist


src/apps/material/material.dev.tsx
```diff
@@ -536,6 +536,17 @@ export default {
...
       control: { type: "text" }
+    },
+    skipQueueMaterialIsAvailableInAnotherEditionText: {
+      name: "Skip queue material is available in another edition",
+      defaultValue:
+        "Skip the queue - The material is available in another edition",
+      control: { type: "text" }
+    },
+    skipQueueMaterialIsAvailableInAnotherEditionReservationsText: {
+      name: "Skip queue reservations cont",
```
spaceo, line 547: Do not understand name. What does it mean?

kasperbirch1: I have changed it to materialIsAvailableInAnotherEditionReservationsText 😊

src/components/reservation/ReservationModalBody.tsx
```diff
@@ -196,7 +207,20 @@ const ReservationModalBody = ({
...
+                  classNames="px-35"
+                  sticky
+                  type="info"
+                  text={`${t("materialIsAvailableInAnotherEditionText")}
+                    - ${otherManifestationPreferred.titles.main}
+                    - ${getAuthorLine(otherManifestationPreferred, t)}
+                    - ${t(
+                      "materialIsAvailableInAnotherEditionReservationsText"
+                    )}:
+                    ${otherManifestationPreferred.reservations}`}
```
kasperg, line 221: Please update this to use placeholders and plural handling.

spaceo:
LGTM 👍

PR #256, opened 2022-12-01
Login to view infomedia review

Adamik10:
#### Link to issue

DDFSOEG-312
DDFSOEG-327

#### Description
This PR introduces a "redirect lo login" functionality for infomedia review views - if a user is NOT logged in they get presented with a button to login. If they successfully login, they get redirected back to the material page with an open Reviews-disclosure and are also additionally scrolled down to the precise review they wanted to view before the redirect took place.

This PR also fixes a bug where the material page would break for works that have reviews where the `review.origin` cannot be made into a link for review metadata line. If that's the case, we now just don't show this metadata with a link.

#### Screenshot of the result
https://user-images.githubusercontent.com/28546954/205275461-f5011a8f-14e2-44cb-8f4b-69e0c55a2740.mov

#### Checklist

n/a


src/apps/material/material.tsx
```diff
@@ -151,6 +152,10 @@ const Material: React.FC<MaterialProps> = ({ wid }) => {
   const parallelManifestations = materialIsFiction(work) ? manifestations : [];
   const infomediaId = getInfomediaId(currentManifestation);
 
+  // Get disclosure URL parameter from the current URL to see if it should be open
+  const shouldOpenReviewDisclosure =
```
spaceo, line 156: we have created a function for this (getUrlQueryParam) that we use in other places. I think you should use that.

src/core/utils/helpers/url.ts
```diff
@@ -134,8 +134,8 @@ export function redirectToLoginAndBack({
   returnUrl,
   trackingFunction
 }: RedirectToLoginAndBackParams) {
-  const { pathname, search } = returnUrl;
-  const localPathToReturnTo = `${pathname}${search}`;
+  const { pathname, search, hash } = returnUrl;
+  const localPathToReturnTo = `${pathname}${search}${hash || ""}`;
```
spaceo, line 138: My ts server in vs code says that `hash` is a string. Is it necessary to write: `hash || ""`. Isn't `hash` sufficient?

src/components/material/ReviewInfomedia.tsx
```diff
@@ -61,7 +98,8 @@ const ReviewInfomedia: React.FC<ReviewInfomediaProps> = ({ review }) => {
         />
       )}
       {/* eslint-enable react/no-danger */}
-      {review.origin && (
+      {/* We want to make sure a link can be made out of the review origin (includes a ".") */}
+      {review.origin && review.origin.includes(".") && (
```
spaceo, line 102: I don't think this is strict enough.
We need this in other places, eg. in the proxy url functionality. So we could create a function that checks if the string is a valid url.
Something like:
```typescript 
const isValidHttpUrl = (text: string) => {
  try {
    const url = new URL(text);
    return url.protocol === 'http:' || url.protocol === 'https:';
  } catch (err) {
    return false;
  }
}

**Review by spaceo — CHANGES_REQUESTED** [review]
Looks good. I have some comment I would like you to look at though.

File: src/core/utils/helpers/url.ts (hunk 2 above)
Comment (Adamik10) [reply src/core/utils/helpers/url.ts:138]: True. We now output just hash :)

File: src/apps/material/material.tsx (hunk 1 above)
Comment (Adamik10) [reply src/apps/material/material.tsx:156]: Perfect addition! Thanks for pointing me the right direction :)

File: src/apps/material/material.tsx [hunk 4]
```diff
@@ -39,6 +39,7 @@ import { PeriodicalEdition } from "../../components/material/periodical/helper";
 import InfomediaModal from "../../components/material/infomedia/InfomediaModal";
 import { useStatistics } from "../../core/statistics/useStatistics";
 import { statistics } from "../../core/statistics/statistics";
+import DisclosureCustomOpening from "../../components/material/disclosures/DisclosureCustomOpening";
```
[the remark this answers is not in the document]
Adamik10: Fair enough :) It is now called DisclosureControllable and resides outside the material folder inside the components level.

spaceo (approved):
LGTM 👍

src/components/Disclosures/DisclosureControllable.tsx
```diff
@@ -40,17 +42,22 @@ const FacetBrowserDisclosure: FC<FacetBrowserDisclosureProps> = ({
...
     >
       <div
         className={clsx(
           "disclosure__headline text-body-large",
           removeHeadlinePadding && "disclosure__headline--no-padding"
         )}
+        onClick={toggleOpen}
+        onKeyDown={toggleOpen}
+        role="button"
+        tabIndex={0}
```
sinejespersen, line 54: can this be opened via keyboard?

Adamik10: yup, indeed it can! :)

https://user-images.githubusercontent.com/28546954/206406926-775e0b92-641b-427b-bc87-3a944dd2f42d.mov

PR #258, opened 2022-12-02
Add label to search header search field

Adamik10:
#### Link to issue
DDFSOEG-330

#### Description
To improve accessibility, this PR adds a label to the search header search field that screen readers can pick up on. This label uses a translatable system text. 

#### Screenshot of the result
n/a

#### Checklist

n/a


src/apps/search-header/search-header.dev.tsx
```diff
@@ -9,11 +9,16 @@ export default {
...
   argTypes: {
-    altText: {
+    searchHeaderIconAltText: {
       name: "Alt text for search button image",
       defaultValue: "search icon",
       control: { type: "text" }
     },
+    searchHeaderInputLabel: {
+      name: "Search header input label",
+      defaultValue: "The main search field used to search for literature.",
```
kasperg, line 19: Consider removing the specific reference to literature. The municipal libraries provide many other types of materials 😅 .

src/components/search-bar/search-bar.tsx
```diff
@@ -11,21 +11,27 @@ const SearchBar: React.FC<SearchBarProps> = ({ getInputProps }) => {
   const t = useText();
   return (
     <>
+      <label htmlFor="search-header-field" className="hide-visually">
+        {t("searchHeaderInputLabel")}
+      </label>
```
kasperg, line 16: Please look into using [`getLabelProps()` provided by Downshift](https://www.downshift-js.com/downshift/#props-used-in-examples).

I would expect this as a minimum provided the necessary link between label and input element instead of hardcoding ids.

kasperg (changes requested):

src/apps/search-header/search-header.entry.tsx
```diff
@@ -4,7 +4,8 @@ import { withUrls } from "../../core/utils/url";
 import SearchHeader from "./search-header";
 
 export interface SearchHeaderTextProps {
-  altText?: string;
+  searchHeaderIconAltText?: string;
```
kasperg, line 7: :+1: for using more specific text prop names.

src/apps/search-header/search-header.dev.tsx
Adamik10: Very fair :) I haven't thought of that.

src/components/search-bar/search-bar.tsx
Adamik10: That's a great suggestion, thanks! We now use downshift's getLabelProps functionality instead :)

kasperg (approved):
:+1: LGTM

src/components/search-bar/search-bar.tsx
```diff
@@ -5,27 +5,37 @@ import { useText } from "../../core/utils/text";
...
+const SearchBar: React.FC<SearchBarProps> = ({
+  getInputProps,
+  getLabelProps
+}) => {
   const t = useText();
   return (
     <>
-      {/* The downshift combobox uses prop spreading by design */}
+      {/* The downshift combobox uses prop spreading by design & associated control is desctructured too */}
+      {/* eslint-disable-next-line jsx-a11y/label-has-associated-control, react/jsx-props-no-spreading */}
```
sinejespersen, line 19: jsx-a11y/label-has-associated-control <- think you can avoid that by wrapping the label around the input <label><input></l...

src/components/search-bar/search-bar.tsx
```diff
@@ -5,27 +5,37 @@ import { useText } from "../../core/utils/text";
...
+        aria-label={t("inputPlaceholderText")}
         {...getInputProps()}
       />
       {/* eslint-enable react/jsx-props-no-spreading */}
       <input
         type="image"
         src={searchIcon}
-        alt={t("altText")}
+        alt={t("searchHeaderIconAltText")}
+        aria-label={t("searchHeaderIconAltText")}
```
sinejespersen, line 38: why both alt and aria-label?

src/components/search-bar/search-bar.tsx
Adamik10: Mmmm @sinejespersen it doesn't seem like it. Check out the screenshot where I wrap it around the input but still get the error.
image

src/components/search-bar/search-bar.tsx
Adamik10: Because the system text should describe what the search icon does/is, which I think is precise enough to use for screen readers too. 
But I'm open to suggestions of course :)

src/components/search-bar/search-bar.tsx
sinejespersen: oh sorry, add the htmlFor instead... I think i've made it work with some sort of combination of the two... https://developer.mozilla.org/en-US/docs/Web/API/HTMLLabelElement/htmlFor

src/components/search-bar/search-bar.tsx
sinejespersen: alt texts are picked up by the screen reader, so if the texts are the same both shouldnt be necessary

src/components/search-bar/search-bar.tsx
Adamik10: @sinejespersen the thing is that Downshift (the library we use for the autosuggest + search input field) gives the input the htmlFor attribute on runtime (see the `{...getLabelProps()}`). Eslint is not happy with it because it cannot see the attribute hardcoded (which is understandable), but we decided (there was indeed an internal discussion about this) to utilize the library instead of hard coding it. Thus the eslint-disable.

src/components/search-bar/search-bar.tsx
Adamik10: Aaaah perfect! Didn't know that! Thanks @sinejespersen 💯 
I'm correcting it now.

src/components/search-bar/search-bar.tsx
sinejespersen: ah I see

sinejespersen (approved):

PR #260, opened 2022-12-02
Fix falling material test

kasperbirch1:
#### Fix falling material test

An unnecessary check has been introduced by mistake

in the MaterialPeriodical was there a check for if selectedPeriodical has a value

but selectedPeriodical is null to start with and will not be set until later in the flow.

The check in MaterialPeriodical ensures that MaterialPeriodicalSelect will never be rendered. which we do not want. Therefore I have removed it.

This is also why material tests fail.

#### Checklist


kasperg (approved):
:+1: LGTM

This error was introduced in #241 where we apparently got so used to failing tests that we accepted this change without checking whether it had introduced additional errors 😞 .

PR #262, opened 2022-12-05
Fix bug: Change <main> to <section>

kasperbirch1:
#### Because dpl-cms provides a main tag where react is rendered, we cannot also use main in the code
Skærmbillede 2022-12-05 kl  15 12 42

#### Checklist


kasperg (approved):
:+1: LGTM

PR #263, opened 2022-12-06
Fix unwanted redirect to material page on tab

kasperbirch1:
#### Link to issue

#### Fix unwanted redirect to material page on tab
Because we use onKeyUp we also need to specify which key we are listening to. Therefore we check if e.key === "Enter" otherwise the tab button and all other buttons will also activate the function.

#### Checklist

### DIFF
--- a/src/components/search-result-list/search-result-list-item/search-result-list-item.tsx
+++ b/src/components/search-result-list/search-result-list-item/search-result-list-item.tsx
@@ -98,7 +98,7 @@ const SearchResultListItem: React.FC<SearchResultListItemProps> = ({
     <article
       className="search-result-item arrow arrow__hover--right-small"
       onClick={handleClick}
-      onKeyUp={handleClick}
+      onKeyUp={(e) => e.key === "Enter" && handleClick}
     >
       <div className="search-result-item__cover">
         <SearchResultListItemCover


src/components/search-result-list/search-result-list-item/search-result-list-item.tsx
```diff
@@ -98,7 +98,7 @@ const SearchResultListItem: React.FC<SearchResultListItemProps> = ({
     <article
       className="search-result-item arrow arrow__hover--right-small"
       onClick={handleClick}
-      onKeyUp={handleClick}
+      onKeyUp={(e) => e.key === "Enter" && handleClick}
```
kasperg, line 101: Consider allowing both Enter and Space as would be the case if the element was a `<button>`.

kasperg (approved):
I considered whether the element should be converted to a`<button>` element because it represent something that triggers a response when clicked. This would also mean that we would get focus, click/key behavior etc. out of the box.

However when reading up on the definition I no longer think that is appropriate. Making the entire element a button [marks all the content as presentational in the accesibility tree](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/button_role#all_descendants_are_presentational) despite the inner markup. This would prevent screen readers etc. from understanding the title, series info, authors, favourite buttons etc.

Consequently I think making the entire element clickable a UI tweak which is best implemented as is.

sinejespersen (approved):

PR #264, opened 2022-12-07
Correcting syntax error in MaterialButtons

kasperbirch1:
#### Correcting syntax error in MaterialButtons
Brackets were missing to check if either hasCorrectAccessType or hasCorrectAccess was true

#### Change isPeriodical to check that it exactly contains "tidsskrift"
Instead of using includes as returned true because "tidsskrift" is part of "tidsskriftsartikel" which is the type of material that can be ordered as a digital copy

#### Checklist


kasperg (approved):
:+1: LGTM

PR #265, opened 2022-12-07
Add test for order digital copy + Move infomedia & periodical into its own separate files

kasperbirch1:
#### add a test for order digital copy + Move infomedia & periodical into its own separate files

There is now a test that checks the following things
- render a material that can be ordered as a digital copy
- render modal to order digital copy
- order digital copy with fail
- order digital copy with succes

in addition, I have split up the material test so that Periodical + Infomedia has its own test files

#### Checklist

### DIFF
--- a/src/apps/material/order-digital-copy.test.ts
+++ b/src/apps/material/order-digital-copy.test.ts
@@ -54,33 +54,27 @@ describe("Material - Order digital copy", () => {
       .should("be.visible")
       .and("contain", "Faglig formidling er ikke kun skriveteknik");
 
-    cy.getBySel("material-header-buttons-find-on-shelf-digital-article").should(
+    cy.getBySel("material-header-buttons-online-digital-article").should(
       "be.visible"
     );
   });
 
   it("render modal to order digital copy", () => {
-    cy.getBySel("material-header-buttons-find-on-shelf-digital-article")
+    cy.getBySel("material-header-buttons-online-digital-article")
       .should("be.visible")
       .click();
 
-    cy.getBySel("email-order-digital-copy")
-      .should("be.visible")
-      .and("have.value", "test@test.com");
+    cy.getBySel("email-order-digital-copy").should("be.visible");
 
-    cy.getBySel("button")
+    cy.getBySel("order-digital-button")
       .should("be.visible")
       .and("contain", "Order digital copy");
   });
 
-  it("order digital copy with fail", () => {
-    cy.getBySel("material-header-buttons-find-on-shelf-digital-article")
+  it("shows an error message if ordering fails", () => {
+    cy.getBySel("material-header-buttons-online-digital-article")
       .should("be.visible")
       .click();
-
-    cy.getBySel("email-order-digital-copy")
-      .should("be.visible")
-      .and("have.value", "test@test.com");
 
     cy.intercept("POST", "**/dpl_das/order", {
       statusCode: 500,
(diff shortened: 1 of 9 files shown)


src/apps/material/order-digital-copy.test.ts
```diff
@@ -0,0 +1,123 @@
...
+  });
+
+  it("order digital copy with succes", () => {
+    cy.getBySel("material-header-buttons-find-on-shelf-digital-article")
+      .should("be.visible")
+      .click();
+
+    cy.getBySel("email-order-digital-copy")
+      .should("be.visible")
+      .and("have.value", "test@test.com");
```
kasperg, line 106: Please remove this assertion. It is already being tested in a previous test.

src/apps/material/order-digital-copy.test.ts
```diff
@@ -0,0 +1,123 @@
...
+      body: {
+        pid: "870971-tsart:34310815",
+        email: "test@test.com"
+      }
+    });
+
+    cy.getBySel("button").should("be.visible").click();
+
+    cy.contains("Digital copy ordered");
+    cy.getBySel("button").should("be.visible").and("contain", "Close").click();
```
kasperg, line 119: Please remove one of the checks here.

The cy.getBySel("button").should("be.visible") is duplicated.

src/apps/material/order-digital-copy.test.ts
```diff
@@ -0,0 +1,123 @@
...
+  });
+
+  it("render modal to order digital copy", () => {
+    cy.getBySel("material-header-buttons-find-on-shelf-digital-article")
+      .should("be.visible")
+      .click();
+
+    cy.getBySel("email-order-digital-copy")
+      .should("be.visible")
+      .and("have.value", "test@test.com");
```
kasperg, line 69: Consider moving this to a separate test:

`it("uses the patron email adress by default", () => {[...]`

This is actually an important assertion as it checks a business requirement and it is not clear from the description of the test why this needs to be tested here. You could also just add an inline comment describing what is tested here and why.

src/apps/material/order-digital-copy.test.ts
```diff
@@ -0,0 +1,123 @@
...
+
+    cy.getBySel("button")
+      .should("be.visible")
+      .and("contain", "Order digital copy");
+  });
+
+  it("order digital copy with fail", () => {
```
kasperg, line 76:
```suggestion
  it("shows an error message if ordering fails", () => {
```

In my opinion this reads better.

src/apps/material/order-digital-copy.test.ts
```diff
@@ -0,0 +1,123 @@
...
+      statusCode: 500,
+      body: {
+        pid: "870971-tsart:34310815",
+        email: "test@test.com"
+      }
+    });
+
+    cy.getBySel("button").should("be.visible").click();
+
+    cy.contains("Error ordering digital copy");
```
kasperg, line 95: Consider how you can add a cy attribute to the error message as this is the approach we are using now.

src/apps/material/order-digital-copy.test.ts
```diff
@@ -0,0 +1,123 @@
...
+    cy.getBySel("button").should("be.visible").and("contain", "Close").click();
+  });
+
+  it("order digital copy with succes", () => {
```
kasperg, line 99:
```suggestion
  it("shows a confirmation message when an order is completed, () => {
```

In my opinion this reads better

src/apps/material/order-digital-copy.test.ts
```diff
@@ -0,0 +1,123 @@
...
+      "/iframe.html?id=apps-material--digital&viewMode=story&type=tidsskriftsartikel"
+    );
+  });
+
+  it("render a material that can be ordered as a digital copy", () => {
+    cy.get("h1")
+      .should("be.visible")
+      .and("contain", "Faglig formidling er ikke kun skriveteknik");
+
+    cy.getBySel("material-header-buttons-find-on-shelf-digital-article").should(
```
kasperg, line 57: I wonder why `find-on-shelf` is part of the selector. I would not expect it to be related to `digital-article`.

src/apps/material/order-digital-copy.test.ts
```diff
@@ -0,0 +1,123 @@
...
+  it("render modal to order digital copy", () => {
+    cy.getBySel("material-header-buttons-find-on-shelf-digital-article")
+      .should("be.visible")
+      .click();
+
+    cy.getBySel("email-order-digital-copy")
+      .should("be.visible")
+      .and("have.value", "test@test.com");
+
+    cy.getBySel("button")
```
kasperg, line 71: Consider if you can add a more specific selector to the button used in the modal. `button` is the fallback value and very generic.

src/apps/material/order-digital-copy.test.ts
```diff
@@ -0,0 +1,123 @@
...
+    cy.getBySel("material-header-buttons-find-on-shelf-digital-article")
+      .should("be.visible")
+      .click();
+
+    cy.getBySel("email-order-digital-copy")
+      .should("be.visible")
+      .and("have.value", "test@test.com");
+
+    cy.intercept("POST", "**/dpl_das/order", {
+      statusCode: 200,
```
kasperg, line 109:
```suggestion
      statusCode: 201,
```

201 is the status code which is actually returned at the moment.

kasperg (approved):
:+1: LGTM

sinejespersen (approved):

PR #268, opened 2022-12-08
Fix order digital copy button

Adamik10:
#### Link to issue
DDFSOEG-345

#### Description
The "Order digital copy" CTA button should only be shown for works that are of the "tidsskriftsartikel" material type, as specified in the project specs.

#### Screenshot of the result
n/a

#### Checklist

n/a

### DIFF
--- a/src/components/material/material-buttons/helper.ts
+++ b/src/components/material/material-buttons/helper.ts
@@ -29,4 +29,13 @@ export const isArticle = (manifestation: Manifestation) => {
   );
 };
 
+export const hasCorrectMaterialType = (
+  desiredMaterialType: string,
+  allMaterialTypes: Manifestation["materialTypes"]
+) => {
+  return allMaterialTypes.some(
+    (type) => type.specific.toLowerCase() === desiredMaterialType.toLowerCase()
+  );
+};
+
 export default {};
--- a/src/components/material/material-buttons/helper.ts
+++ b/src/components/material/material-buttons/helper.ts
@@ -31,9 +31,9 @@ export const isArticle = (manifestation: Manifestation) => {
 
 export const hasCorrectMaterialType = (
   desiredMaterialType: string,
-  allMaterialTypes: Manifestation["materialTypes"]
+  manifestation: Manifestation
 ) => {
-  return allMaterialTypes.some(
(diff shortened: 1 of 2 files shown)


src/components/material/material-buttons/helper.ts
```diff
@@ -29,4 +29,13 @@ export const isArticle = (manifestation: Manifestation) => {
   );
 };
 
+export const hasCorrectMaterialType = (
+  desiredMaterialType: string,
+  allMaterialTypes: Manifestation["materialTypes"]
```
kasperg, line 34: Please change the argument type (and name) to `Manifestation` instead of `Manifestation["materialTypes"]`.

In my opinion this has three advantages:

1. It reads better: We check whether the manifestation has the correct material type "whatever". The location of the material types within the manifestation is an implementation detail.
2. There is less typing for the caller
3. It follows the pattern of the other `hasCorrect` functions in this file.

Consider refactoring `isArticle` to call `hasCorrectMaterialType("tidsskriftsartikel", manifestation), ||  hasCorrectMaterialType("avisartikel", manifestation)`.

kasperg (changes requested):

src/components/material/material-buttons/helper.ts
```diff
@@ -19,13 +18,19 @@ export const hasCorrectAccessType = (
...
-  return (
-    intersection(allMaterialTypes, ["tidsskriftsartikel", "avisartikel"])
-      .length > 0
+};
+
+export const isArticle = (manifestation: Manifestation) => {
+  return !!(
+    hasCorrectMaterialType("tidsskriftsartikel", manifestation) ||
+    hasCorrectMaterialType("avisartikel", manifestation)
   );
```
kasperg, line 34: Why do we need double negation/`!` here?


The following returns a boolean: 

```
(
    hasCorrectMaterialType("tidsskriftsartikel", manifestation) ||
    hasCorrectMaterialType("avisartikel", manifestation)
)
```  

Double-negation results in the original value. We should even require any type juggling.

kasperg (approved):
:+1: LGTM

sinejespersen (approved):

PR #269, opened 2022-12-08
Add space between title and publication year (FindOnShelf)

kasperbirch1:
#### Add space between title and publication year (FindOnShelf)
I have used a template literals to create a string with a space in front of the publication year and changed the rendering logic to not render an empty string if there is no publication year

#### Screenshot of the result
Skærmbillede 2022-12-08 kl  14 42 12

#### Checklist


Adamik10 (approved):
Perf 👍

PR #272, opened 2022-12-12
Hide autosuggest visually instead of not rendering it when it's not open

Adamik10:
#### Link to issue
DDFSOEG-346

#### Description
This PR addresses a failing accessibility test in the dpl-cms once dpl-react is set into drupal. The test was complaining that our search header label for autosuggest was referring to a non-existing DOM element, because we didn't render autosuggest at all unless it should be visible. Now, we instead just hide it visually.

#### Screenshot of the result
image

#### Checklist

n/a


src/components/autosuggest/autosuggest.tsx
```diff
@@ -46,7 +47,7 @@ export const Autosuggest: React.FC<AutosuggestProps> = ({
       {/* eslint-disable react/jsx-props-no-spreading */}
       {/* The downshift combobox works this way by design */}
       <ul
-        className="autosuggest pb-16"
+        className={`autosuggest pb-16 ${clsx({ "hide-visually": !isOpen })}`}
         {...getMenuProps()}
         style={!isOpen ? { display: "none" } : {}}
```
kasperg, line 52: Please consider whether this change is actually needed.

At the moment we already hide the autosuggest if it is not open by setting `display: none`.

src/apps/search-header/search-header.test.ts
```diff
@@ -38,7 +38,7 @@ describe("Search header app", () => {
 
   it("Doesn't show suggestions before 3 characters are typed", () => {
     cy.get(".header__menu-search-input").focus().type("ha");
-    cy.get(".autosuggest").should("not.exist");
+    cy.get(".autosuggest").should("not.be.visible");
```
kasperg, line 41: :+1:

src/components/autosuggest/autosuggest.tsx
Adamik10: Oh my, I totally should have noticed that 🙈

kasperg (approved):
:+1: LGTM

sinejespersen (approved):

PR #273, opened 2022-12-13
Fix aria attributes for the disclosure element

Adamik10:
#### Link to issue
DDFSOEG-347

#### Description
This PR fixes a Pa11y accessibility error where our disclosure component aria attributes were not located on the same HTML element. 

#### Screenshot of the result
n/a

#### Checklist

n/a

### DIFF
--- a/src/components/Disclosures/DisclosureControllable.tsx
+++ b/src/components/Disclosures/DisclosureControllable.tsx
@@ -66,7 +66,9 @@ const DisclosureControllable: FC<DisclosureControllableProps> = ({
           alt=""
         />
       </div>
-      {isOpen && <div id={disclosureId}>{children}</div>}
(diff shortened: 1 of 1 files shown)


src/components/Disclosures/DisclosureControllable.tsx
```diff
@@ -66,7 +66,9 @@ const DisclosureControllable: FC<DisclosureControllableProps> = ({
           alt=""
         />
       </div>
-      {isOpen && <div id={disclosureId}>{children}</div>}
```
spaceo, line 69: I think we should stick to "conditional rendering" instead of class toggling if possible.
No need to render stuff if we can avoid it (explained in cave man language).

spaceo (approved):
Approved but please look at comment

sinejespersen (approved):

PR #276, opened 2022-12-13
Availability label "half" fix (still missing logic for if the material is not costfree) 

kasperbirch1:
#### add useAvailabilityData
useAvailabilityData is a custom hook, which checks the availability of a material. It takes an object with three properties: materialType, faustIds, and isbn. 

The hook uses two other custom hooks, useGetAvailabilityV3 and useGetV1ProductsIdentifier, to make requests to APIs and check the availability of the material at different branches and on the online platform. 

The hook returns an object with a boolean property, isAvailable, which represents the availability of the material.

#### Screenshot of the result
image

##### Example 
http://localhost/iframe.html?args=&id=apps-material--e-bog-prinsen-harry&viewMode=story&type=ebog

#### Checklist

I have moved some of the availability logic from AvailabilityLabel to useAvailabilityData while coding.

### DIFF
--- a/src/components/availability-label/availability-labels.tsx
+++ b/src/components/availability-label/availability-labels.tsx
@@ -37,7 +37,6 @@ export const AvailabiltityLabels: React.FC<AvailabilityLabelsProps> = ({
         const materialType = materialTypes[0].specific;
         const faustId = convertPostIdToFaustId(pid);
         const url = constructMaterialUrl(materialUrl, workId, materialType);
-        const isEReolen = hasCorrectSource(["eReolen"], item);
 
         return (
           <AvailabilityLabel
@@ -61,7 +60,7 @@ export const AvailabiltityLabels: React.FC<AvailabilityLabelsProps> = ({
                 : undefined
             }
             isbn={identifiers?.[0]?.value ?? ""}
-            isEReolen={isEReolen}
+            isTrue={hasCorrectSource(["eReolen"], item)}
           />
         );
       })}
--- a/src/components/availability-label/availability-labels.tsx
+++ b/src/components/availability-label/availability-labels.tsx
@@ -11,7 +11,6 @@ import { WorkId } from "../../core/utils/types/ids";
 import { useUrls } from "../../core/utils/url";
 import { AvailabilityLabel } from "./availability-label";
 import { Manifestation } from "../../core/utils/types/entities";
-import { hasCorrectSource } from "../material/material-buttons/helper";
 
 export interface AvailabilityLabelsProps {
   manifestations: Manifestation[];
@@ -37,7 +36,7 @@ export const AvailabiltityLabels: React.FC<AvailabilityLabelsProps> = ({
         const materialType = materialTypes[0].specific;
         const faustId = convertPostIdToFaustId(pid);
         const url = constructMaterialUrl(materialUrl, workId, materialType);
-
+        const accessTypesCodes = item.accessTypes.map((t) => t.code);
         return (
           <AvailabilityLabel
             key={pid}
@@ -45,6 +44,7 @@ export const AvailabiltityLabels: React.FC<AvailabilityLabelsProps> = ({
             cursorPointer={cursorPointer}
             faustIds={[faustId]}
             manifestText={materialType}
+            accessTypes={accessTypesCodes}
             selected={
               manifestation &&
               materialType === getManifestationType(manifestation)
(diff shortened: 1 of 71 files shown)


src/components/availability-label/availability-labels.tsx
```diff
@@ -56,6 +57,8 @@ export const AvailabiltityLabels: React.FC<AvailabilityLabelsProps> = ({
                   }
                 : undefined
             }
+            isbn={identifiers?.[0]?.value ?? ""}
+            isTrue={hasCorrectSource(["eReolen"], item)}
```
kasperg, line 61: Please rework the naming here. `isTrue` is not a very expressive name for what the prop represents.

If the prop is a way to force an availability label to have a specific value then one approach would be to change the value to `isAvailable`and make it an optional boolean. If it has a value then use that value for availability. If it is undefined then we look it up in the business systems.

src/components/availability-label/helper.ts
```diff
@@ -0,0 +1,83 @@
...
+  const blacklistBranches = config("blacklistedAvailabilityBranchesConfig", {
+    transformer: "stringToArray"
+  });
+
+  // Forcing isAvailable to true if isTrue is true
+  useEffect(() => {
+    if (isTrue) setIsAvailable(true);
+  }, [isTrue]);
+
+  const hasIsbn = Boolean(isbn);
```
kasperg, line 33: Please reconsider this.

Instead of making isbn a string which can be interpreted as a boolean if it is empty I suggest making the isbn parameter nullable or optional.

src/components/availability-label/helper.ts
```diff
@@ -0,0 +1,83 @@
+import { useEffect, useState } from "react";
+import { useGetAvailabilityV3 } from "../../core/fbs/fbs";
+import { useGetV1ProductsIdentifier } from "../../core/publizon/publizon";
+import { useConfig } from "../../core/utils/config";
+
+export const onlineMaterialAllowList = ["lydbog (net)"];
+
+export const isMaterialTypeOnline = (materialType: string) =>
+  onlineMaterialAllowList.includes(materialType);
```
kasperg, line 9: Please reconsider this.

Specifying internet audio books as the only online material type seems like a stretch. If it has been decided by the business that this is the way we do things for now then I guess it is OK but then it needs some kind of inline documentation explaining this.

src/components/availability-label/helper.ts
```diff
@@ -0,0 +1,83 @@
...
+          // Set isAvailable to true if the material is costFree (blue title) ind Publizon
+          // if (res?.product?.costFree) {
+          //   setIsAvailable(true);
+          // }
+          // Check if the material quota is available if the material is not costFree
+        },
+      onError: (error: unknown) => {
+        if (error instanceof Error) {
+          // 128 is the error code for "Bogen er ikke tilgængelig for udlån"
+          if (error.cause && Number(error.cause) === 128) {
```
kasperg, line 72: :+1: Perfect use for an inline comment.

src/components/availability-label/helper.ts
```diff
@@ -0,0 +1,83 @@
...
+            setIsAvailable(false);
+          }
+        }
```
kasperg, line 75: Have you though about what the right way to handle this is? What do we do if the error is not an error code 128?

src/components/availability-label/helper.ts
kasperg, line 72: Please reconsider this. Force typing with `Number` does not seem like a good idea.

src/components/availability-label/helper.ts
```diff
@@ -0,0 +1,83 @@
...
+      }
+    }
+  );
+
+  useGetV1ProductsIdentifier(isbn, {
+    query: {
+      // This should only be enabled if the material is an online material type and the isbn is not empty
+      enabled: isOnline && isTrue === false && hasIsbn,
```
kasperg, line 57: This is an example of non-helpful inline commenting. It restates what the following relatively simple and readable line of code does.

kasperg (changes requested):
I do not think this turned out quite right. Let's look at this together.

src/components/availability-label/helper.ts
kasperbirch1: isAvailable is set to false and is only set to true in onSuccess. therefore, it is only to be explicit to show that if there is an error 128, it must be false

kasperg (approved):
:+1: LGTM

jeppekroghitk (approved):

PR #278, opened 2022-12-14
Feature/cmn 191 modals in loan list

sinejespersen:
#### Link to issue

#### Description

This addresses the issues where a modal is open, but the user can still scroll in the list behind.
- move modals to loan list / root (prop drill functions to open modals)
- hide loan list on open modals
- add new modal ids for more specificity
- remove function not used from date.ts and make only function remaining the default export (changes imports too)
- add modal-helpers.ts in core.

#### Screenshot of the result
 n/a

#### Checklist


jeppekroghitk (approved):

src/apps/loan-list/list/loan-list.tsx
```diff
@@ -70,67 +108,134 @@ const LoanList: FC<LoanListProps> = ({ pageSize }) => {
...
-                header={t("loanListPhysicalLoansTitleText")}
-                amount={physicalLoans.length}
+    <>
+      <div
+        style={modalIds.length > 0 ? { display: "none" } : {}}
+        className="loan-list-page"
+      >
+        <h1 className="text-header-h1 my-32">{t("loanListTitleText")}</h1>
+        {((Array.isArray(physicalLoans) && physicalLoans.length > 0) ||
+          (Array.isArray(digitalLoans) && digitalLoans.length > 0)) && (
```
spaceo, line 162: For the readability you could consider moving this away from the jsx and put it before the return with a variablename called something like: `const listContainsLoans = ...`

src/apps/loan-list/list/loan-list.tsx
```diff
@@ -70,67 +108,134 @@ const LoanList: FC<LoanListProps> = ({ pageSize }) => {
...
+      setDueDate(dueDateInput);
+      open(`${dueDateModal}${dueDateInput}`);
+    },
+    [dueDateModal, open]
+  );
+
   useEffect(() => {
     const modalString = getUrlQueryParam("modal");
+    // if there is a loan details query param, loan details modal should be opened
+    if (modalString && modalString.includes(loanDetails as string)) {
```
spaceo, line 130: I think `modalUrlParam`would be a more descriptive than `modalString` that just states that it is a string concerning a modal.

src/apps/loan-list/materials/selectable-material/selectable-material.tsx
```diff
@@ -7,114 +7,99 @@ import { FaustId } from "../../../../core/utils/types/ids";
...
-    },
-    [faust, identifier, open]
-  );
+  const openLoanDetailsModalHandler = useCallback(() => {
+    if (faust) {
+      openLoanDetailsModal(faust);
+    }
+    if (identifier) {
+      openLoanDetailsModal(identifier);
+    }
```
spaceo, line 38: Can they both be open at the same time?

src/apps/loan-list/materials/selectable-material/selectable-material.tsx
```diff
@@ -7,114 +7,99 @@ import { FaustId } from "../../../../core/utils/types/ids";
...
-          <div className="list-materials__content">
-            <div className="list-materials__content-status">
-              <div className="status-label status-label--outline ">
-                {materialType}
-              </div>
+    <li>
+      <div
+        className={`list-materials ${
+          disabled ? "list-materials--disabled" : ""
+        }`}
```
spaceo, line 46: We have been using `clsx` for this throughout our part of the codebase

src/apps/loan-list/materials/stackable-material/stackable-material.tsx
```diff
@@ -48,73 +41,54 @@ const StackableMaterial: FC<StackableMaterialProps & MaterialProps> = ({
...
-            <MaterialOverdueLink showOn="desktop" dueDate={dueDate} />
-          </MaterialInfo>
-        )}
-        <MaterialStatus loan={loan}>
+    <button
+      type="button"
+      onClick={(e) => openLoanDetailsModalHandler(e)}
+      className={`list-reservation my-32 ${
+        additionalMaterials > 0 ? "list-reservation--stacked" : ""
+      }`}
```
spaceo, line 63: Another clsx comment.... bla bla :)

src/apps/reservation-list/reservation-material/reservation-material.tsx
```diff
@@ -65,7 +65,9 @@ const ReservationMaterial: FC<ReservationMaterialProps & MaterialProps> = ({
         )}
         <ReservationInfo branches={branches} reservationInfo={reservation} />
       </button>
-      <MaterialDetailsModal modalEntity={reservation} material={material}>
+      <MaterialDetailsModal
+        modalId={reservation.faust || reservation.identifier || ""}
```
spaceo, line 69: Could it be done in the same way as [here](https://github.com/danskernesdigitalebibliotek/dpl-react/pull/278/files#diff-3a213a7e4253d895c2972925b93e7c4a0808e596803bcf4d46ce2ba5d5af5ee8L21)?

I am asking with absolutely no overview if it makes sense :) But just reckognized a pattern...

src/core/utils/helpers/modal-helpers.ts
```diff
@@ -0,0 +1,54 @@
+import { getModalIds } from "./general";
+
+export const containsDueDateModalString = (modalString: string) => {
```
spaceo, line 3: Not fond of using the type of variable as a part of the variable. Something that describes the context of the modalId would clarify better what the purpose of the function is, I think.

src/apps/loan-list/materials/selectable-material/selectable-material.tsx
sinejespersen: That cannot happen

src/core/utils/helpers/modal-helpers.ts
```diff
@@ -0,0 +1,54 @@
...
+
+  let idFound = modalString.match(regexIdentifier);
+  if (!idFound) {
+    idFound = modalString.match(regexFaust);
+    if (!idFound) {
+      return null;
+    }
+  }
+  return idFound[0];
+};
```
spaceo, line 39: I am not completely shure what you are trying to match. 
If `modalString` is more than 13 digits it will also match that (if it is possible to be more than 13).
Can there be characters before and/or after?
I think it is more clear if you have one regex. without knowing the problem scope exactly you could also write:
```
const result = modalString.match(/(^\d{8}$)|(^\d{13}$)/);

File: src/core/utils/helpers/modal-helpers.ts [hunk 9]
```diff
@@ -0,0 +1,54 @@
...
+  const regexIdentifier = new RegExp(`${loanDetails}\\d{13}`, "g");
+  const regexFaust = new RegExp(`${loanDetails}\\d{8}`, "g");
+  let modalId = modalString.match(regexIdentifier);
+  if (!modalId) {
+    modalId = modalString.match(regexFaust);
+    if (!modalId) {
+      return null;
+    }
+  }
+  return modalId[0];
```
spaceo, line 53: If you choose to follow the idea that I wrote about the other regex function you could do the same here :)

spaceo (changes requested):
Looks good. I have some comments though. Most important ones are the ones about the modal id matching

src/apps/loan-list/list/loan-list.tsx
sinejespersen: done

src/apps/loan-list/list/loan-list.tsx
sinejespersen: done

src/apps/loan-list/materials/selectable-material/selectable-material.tsx
sinejespersen: no

src/core/utils/helpers/modal-helpers.ts
sinejespersen: there can be characters before and after, does that change this question?

src/apps/reservation-list/reservation-material/reservation-material.tsx
sinejespersen: This sends the modalid into the linked component

src/core/utils/helpers/modal-helpers.ts
```diff
@@ -36,15 +36,12 @@ export const idFromLoanDetailsModalQueryParam = (queryParam: string) => {
...
-    if (!modalId) {
-      return null;
-    }
+  const regexIdentifier = new RegExp(`${loanDetails}(\\d{13})|(\\d{8})`, "g");
+  const modalId = queryParam.match(regexIdentifier);
+  if (modalId) {
+    return [modalId];
   }
-  return modalId[0];
+  return "";
```
spaceo, line 38: Just a suggestion:
```typescript
return modalId ? [modalId] : "";

spaceo (approved):
LGTM 👍

PR #279, opened 2022-12-14
Fix repeating ids for reservation modals

Adamik10:
#### Link to issue
DDFSOEG-353

#### Description
Accessibility Pa11y test found a problem with our reservation modals - they should all have unique ids. The reason they didn't was because there is a modal rendered for each edition of a work (opened by buttons from the "editions" disclosure) + one main modal that is opened by the main reservation button on the top of the page. My solution adds "main" to the main modal id. All the functionality works as prior, and the test is now passing too.

#### Screenshot of the result
n/a

#### Checklist

The "order digital copy" integration test is failing because I don't have the newest changes in this branch - order digital copy changes reside in develop.

### DIFF
--- a/src/apps/material/material.tsx
+++ b/src/apps/material/material.tsx
@@ -242,14 +242,6 @@ const Material: React.FC<MaterialProps> = ({ wid }) => {
             setSelectedPeriodical={setSelectedPeriodical}
             isMainModal
           />
-          <ReservationModal
-            mainManifestation={currentManifestation}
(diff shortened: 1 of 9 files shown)


spaceo:
We had a talk yesterday. And I came up with a possible that, in my eyes, attacks the problem at the root cause. But lets talk about it :)

spaceo (approved):
LGTM 👍

sinejespersen (approved):

PR #282, opened 2022-12-19
Fix autosuggest topic search

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFSOEG-326

#### Description
The bottom 7 suggestions of autosuggest should link to a search result page that is filtered based on material type for - books, ebooks, movies, audio books, music, games, and animated series. The autosuggest itself was already sending a url search parameter, but the search result component needed to be adjusted to parse these on load. 

#### Screenshot of the result
image

#### Checklist

n/a


src/apps/search-result/search-result.tsx
```diff
@@ -76,7 +77,7 @@ const SearchResult: React.FC<SearchResultProps> = ({ q, pageSize }) => {
       name: statistics.searchQuery.name,
       trackedData: q
     });
-    // We actaully just want to track if the query changes.
+    // We actually just want to track if the query changes.
```
kasperg, line 80: :+1: for correcting this but putting it in a separate commit.

src/apps/search-result/search-result.tsx
```diff
@@ -98,6 +99,24 @@ const SearchResult: React.FC<SearchResultProps> = ({ q, pageSize }) => {
...
+          facet: FacetField.MaterialTypes,
+          term: { key: materialTypeUrlFilter, term: materialTypeUrlFilter }
+        },
+        action: "add"
+      });
+    }
+    // We only want to do this once, so we need the dependency array empty
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
+
```
kasperg, line 119: Consider if this can be replaced by [`useSearchParams()`](https://github.com/streamich/react-use/blob/master/docs/useSearchParam.md) from `react-use`. Perhaps it is too complex compared to what we want to do.

src/apps/search-header/search-header.tsx
```diff
@@ -252,8 +252,9 @@ const SearchHeader: React.FC = () => {
             searchUrl,
             selectedItemString,
             filter: {
-              materialType:
+              materialType: encodeURI(
                 autosuggestCategoryList[highlightedCategoryIndex].type
+              )
```
kasperg, line 257: Please move this encoding to `constructSearchUrlWithFilter()`.

This way we avoid forcing each caller to deal with encoding.

Based on the documentation it also seems as if [`encodeURI()` deals with an entire URI](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURI) - not just a single part. This also points to the fact that this responsibility should be push down the call stack where we have the entire URL available.

kasperg (changes requested):
:+1: Nice but please rework the URI encoding.

src/apps/search-result/search-result.tsx
```diff
@@ -98,6 +99,24 @@ const SearchResult: React.FC<SearchResultProps> = ({ q, pageSize }) => {
     }
   }, [campaignFacets, mutate]);
 
+  // Check for material type filters in url on pageload
+  // (this part will be reworked later based on a change request from the client)
```
kasperg, line 103: I do not think "(this part will be reworked later based on a change request from the client)" is very helpful apart from signaling that we expect this to change.

1. It refers to process "change request" and "client" which will mean little to the average reader.
2. It does not explain *why* we expect this to be reworked

I suggest

```suggestion
  // This is an initial, intentionally simple approach supporting what is required by the search header.
  // It could be reworked to support all filters and terms at a later point.
```

src/apps/search-result/search-result.tsx
Adamik10: Thanks for the suggestion, but it wouldn't save many lines of code. I'll keep it as is :)

src/apps/search-header/search-header.tsx
Adamik10: good point! I moved it into the `appendQueryParametersToUrl()` function, where it all goes down.

kasperg (approved):
:+1: LGTM

src/core/utils/types/material-type.ts
```diff
@@ -1,15 +1,15 @@
 const enum MaterialType {
```
sinejespersen, line 1: Where does these "translations" come from? Like why is `game = "playstation 5"`

sinejespersen (approved):

src/core/utils/types/material-type.ts
kasperg: These are term names for the material types provided by Tue.

PR #285, opened 2022-12-20
Order digital article copy - login redirect

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFSOEG-302

#### Description
This PR introduces a helper function that is used when clicking on the reserve button for digital articles and infomedia articles. This function checks for whether a user is logged in and then either opens the given modal if so, or otherwise redirects the user to the login page with the possibility to come back to where the user left off & the modal open.

#### Screenshot of the result
n/a

#### Checklist

n/a

### DIFF
--- a/src/core/guardedRequests.slice.ts
+++ b/src/core/guardedRequests.slice.ts
@@ -9,8 +9,7 @@ import { persistor } from "./store";
 import type { RootState } from "./store";
 import { getCurrentUnixTime } from "./utils/helpers/date";
 import {
-  appendQueryParametersToUrl,
-  getCurrentLocation,
+  currentLocationWithParametersUrl,
   redirectToLoginAndBack,
   turnUrlStringsIntoObjects
 } from "./utils/helpers/url";
@@ -103,12 +102,9 @@ export const guardedRequest = createAsyncThunk(
         // And redirect to external login.
         const { authUrl } = getUrlsFromState(getState() as RootState);
         if (authUrl) {
-          const returnUrl = appendQueryParametersToUrl(
-            new URL(getCurrentLocation()),
-            {
-              [AUTH_PARAM]: "1"
-            }
-          );
+          const returnUrl = currentLocationWithParametersUrl({
+            [AUTH_PARAM]: "1"
+          });
           redirectToLoginAndBack({ authUrl, returnUrl });
         }
       });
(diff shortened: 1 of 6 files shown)


src/core/utils/modal.tsx
```diff
@@ -128,30 +116,36 @@ export type GuardedOpenModalProps = {
...
+          returnUrl,
+          trackingFunction: props.trackOnlineView
+        });
+        return;
+      }
+      // If user is not anonymous we just open the given modal + potentially track it.
+      if (props.trackOnlineView) {
+        props.trackOnlineView();
+      }
+      props.open(props.modalId);
```
spaceo, line 146: As far as I can see it, we do not need the open prop for this function, but we should be able to use:
```typescript
dispatch(openModal({ modalId }));

File: src/core/utils/modal.tsx [hunk 2]
```diff
@@ -128,30 +116,36 @@ export type GuardedOpenModalProps = {
...
+export const useModalButtonHandler = () => {
+  const dispatch = useDispatch();
+  return {
+    open: (modalId: ModalId) => {
+      return dispatch(openModal({ modalId }));
+    },
+    close: (modalId: ModalId) => {
+      return dispatch(closeModal({ modalId }));
+    },
+    guardedOpenModal: (props: GuardedOpenModalProps) => {
```
spaceo, line 128: I don't think it is necesssary to suffix the function with "modal". It is already in the "modal scope". Otherwise we should rename the other functions to: `openModal` and `closeModal` which I find redundant.

I think you could destruct the props like we do in most cases. So it would be:
```typescript
openGuarded: ({modalId, authUrl, trackOnlineView}: GuardedOpenModalProps) => {

**Review by spaceo — CHANGES_REQUESTED** [review]
Great. But I think there can be a couple of improvements

**Review by spaceo — APPROVED** [review]

File: src/core/guardedRequests.slice.ts [hunk 3]
```diff
@@ -103,12 +102,9 @@ export const guardedRequest = createAsyncThunk(
         // And redirect to external login.
         const { authUrl } = getUrlsFromState(getState() as RootState);
         if (authUrl) {
-          const returnUrl = appendQueryParametersToUrl(
-            new URL(getCurrentLocation()),
-            {
-              [AUTH_PARAM]: "1"
-            }
-          );
+          const returnUrl = currentLocationWithParametersUrl({
```
sinejespersen, line 105: I approved, but what does `[AUTH_PARAM]: "1"`mean?

sinejespersen (approved):

PR #286, opened 2022-12-20
Creators are missing" should not appear when there is no value (Remove creatorsAreMissingText)

kasperbirch1:
#### Creators are missing" should not appear when there is no value

The situation is due to the fact that we have transitioned from a situation where dummydata always contained authors/origin to a situation where this is no longer the case.

#### Screenshot of the result
Skærmbillede 2022-12-20 kl  14 04 29 (3)

Skærmbillede 2022-12-20 kl  14 06 32 (3)

Skærmbillede 2022-12-20 kl  14 07 07 (3)

#### Checklist

### DIFF
--- a/src/apps/material/material.dev.tsx
+++ b/src/apps/material/material.dev.tsx
@@ -200,11 +200,6 @@ export default {
       defaultValue: "Genre",
       control: { type: "text" }
     },
-    creatorsAreMissingText: {
-      name: "Creators are missing",
-      defaultValue: "Creators are missing",
-      control: { type: "text" }
-    },
     readArticleText: {
       name: "Read article",
       defaultValue: "Read article",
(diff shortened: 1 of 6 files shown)


src/apps/material/material.dev.tsx
```diff
@@ -200,11 +200,6 @@ export default {
       defaultValue: "Genre",
       control: { type: "text" }
     },
-    creatorsAreMissingText: {
-      name: "Creators are missing",
-      defaultValue: "Creators are missing",
-      control: { type: "text" }
-    },
```
kasperg, line 207: :+1: for remembering to remove the props and controls.

kasperg (approved):
:+1: Excellent. I appreciate the screenshots.

PR #288, opened 2022-12-21
Adds new function `getWorkTitle`

kasperbirch1:
#### Adds new function `getWorkTitle`

This PR adds a new function called `getWorkTitle` to the project. The function takes in a `Work` object and a string indicating the type of title to retrieve ("original" or "full"). If the `Work` object is fiction and the specified title exists, the function returns the title with the work year appended to it. Otherwise, it returns `null`.

#### Checklist


kasperg:
Closed for now as things work as expected.

PR #290, opened 2022-12-22
Update rendering of InfomediaModal and DigitalModal components

kasperbirch1:
#### This pull request refactors the code for rendering the InfomediaModal and DigitalModal components in the material app. 

Changes made:

- InfomediaModal is now always rendered as long as infomediaId is truthy.
- DigitalModal is now only rendered if the result of hasCorrectAccess("DigitalArticleService") is true 

**Previously, the rendering of DigitalModal was dependent on infomediaId, which was not necessary.**

This commit also removes the unnecessary check for currentManifestation and refactors the code to correctly render the InfomediaModal and DigitalModal components based on their respective prop values.

#### Checklist

### DIFF
--- a/src/apps/material/material.tsx
+++ b/src/apps/material/material.tsx
@@ -232,20 +232,20 @@ const Material: React.FC<MaterialProps> = ({ wid }) => {
           />
         </>
       ))}
-      {currentManifestation && infomediaId && (
-        <>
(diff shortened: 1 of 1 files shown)


kasperg (approved):
:+1: LGTM

PR #291, opened 2022-12-22
ITKDev - Feature/5396 favorites list

sinejespersen:
#### Link to issue

https://platform.dandigbib.org/issues/5396

#### Description

- add favorites list
- and tests of said
- add SearchResultListItemAdapter to make favorites list elements comply with search-result-list-item
- change availability label a little bit. alt ="", and only tabindex 0 if a function is present

#### Screenshot of the result

image

#### Checklist


src/apps/favorites-list/FavoritesList.tsx
```diff
@@ -0,0 +1,60 @@
...
+  const { itemsShown, PagerComponent } = usePager(materials.length, pageSize);
+
+  useEffect(
+    () => setDisplayedMaterials([...materials].splice(0, itemsShown)),
+    [itemsShown, materials]
+  );
+
+  useEffect(() => {
+    if (data && data.collections) {
+      setMaterials(data.collections as Pid[]);
```
spaceo, line 27: I think we should avoid type assertions if possible.
I think we cannot avoid it in this case, because the data fetching hook is generated by Orval.
But at least we can try to define the type as close to the source as possible.
Is it possible to do so in the beginning of the component like [this](https://github.com/danskernesdigitalebibliotek/dpl-react/blob/f6b21e6ebf7ccc0eab3e0ba03326b32d06a8660b/src/apps/search-result/search-result.tsx#L114-L118)?

src/apps/favorites-list/FavoritesList.entry.tsx
```diff
@@ -0,0 +1,57 @@
...
+  searchUrl: string;
+  fbsBaseUrl: string;
+  materialOverdueUrl: string;
+  feesPageUrl: string;
+  publizonBaseUrl: string;
+  dplCmsBaseUrl: string;
+  coverBaseUrl: string;
+  materialBaseUrl: string;
+  fbiBaseUrl: string;
+}
```
spaceo, line 22: I think we should make a global interface for the urls somewhere [here](src/core/utils/types). We reuse them a lot of places.

src/apps/favorites-list/FavoritesList.tsx
```diff
@@ -0,0 +1,60 @@
...
+          {t("favoritesListMaterialsText", {
+            placeholders: { "@count": materials.length }
+          })}
+        </p>
+      )}
+      {displayedMaterials.length > 0 && (
+        <ul className="search-result-page__list my-32">
+          {displayedMaterials.map((pid) => (
+            <li key={pid}>
+              <SearchResultListItemAdapter key={pid} pid={pid} />
```
spaceo, line 47: I am sorry but I don't think you can go this way.
I agree in that the search result list and the favourites list have so much similarties that they both can be covered by the same model.
But! It has to be solved then from the ground up by generalise the list both in dpl-design-system and in dpl-react  by calling it something generic like: "card list" or similar. And then use it both for the search results and for the favourites list.
I know it takes some extra work. But that is the way to do it in order to do it to keep the code bases maintainable in the future.

spaceo (changes requested):
Looks good in general.
But there is a major issue that I think should be taken care of.

src/apps/favorites-list/FavoritesList.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+const FavoritesList: React.FC<FavoritesListProps> = ({ pageSize }) => {
+  const t = useText();
+  const { data } = useGetList("default");
+  const [displayedMaterials, setDisplayedMaterials] = useState<Pid[]>([]);
+  const [materials, setMaterials] = useState<Pid[]>([]);
+  const { itemsShown, PagerComponent } = usePager(materials.length, pageSize);
+
+  useEffect(
+    () => setDisplayedMaterials([...materials].splice(0, itemsShown)),
```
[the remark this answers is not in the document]
jeppekroghitk: Okay, but the code you provide here does not achieve the same as the current code, so im not sure why it should replace it.

src/apps/favorites-list/FavoritesList.entry.tsx
jeppekroghitk: I think we should note it, so someone can do it at some point, but i think that a global generalization should be done in one go, and therefore is out of the scope of this PR

yepzdk (approved):

spaceo (approved):
LGTM 👍

PR #292, opened 2022-12-22
Add useGetCleanBranches hook + filters to useGetFacets and useIntelligentFacetsQuery

kasperbirch1:
#### This pull request includes the following changes:

Addition of the useGetCleanBranches hook, which returns an array of clean branches. The hook retrieves an array of branches from the config object, excludes any branches that are blacklisted, and then cleans the IDs of the remaining branches. This is used to filter out blacklisted branches in useGetFacets, useIntelligentFacetsQuery, and useSearchWithPaginationQuery.

- Added filters to intelligentFacets to exclude blacklisted branches. The filters argument was added to the intelligentFacets query to exclude blacklisted branches, as implemented in useGetFacets and useSearchWithPaginationQuery.

- Restructuring of the code to avoid cycle dependencies. 
The code has been modified to remove circular dependencies between modules.

Overall, these changes will improve the efficiency and smoothness of the code. Thank you for considering this pull request.

#### Screenshot of the result
image

image

#### Checklist


src/core/utils/useGetCleanBranches.tsx
```diff
@@ -0,0 +1,43 @@
+import { AgencyBranch } from "../fbs/model";
```
kasperg, line 1: Please rename this file.

I suggest something like `branches.tsx`. In the current form it contains exported functions besides the `useGetCleanBranches` hook and other files use these exported functions as well.

kasperg (approved):
:+1: I have one thing for you to think about before passing this along. Otherwise LGTM.

sinejespersen (approved):

PR #295, opened 2022-12-29
Enable "find on shelf" button if no material specimens available

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFSOEG-359

#### Description
The find on shelf button should always be enabled for physical materials such as books, CDs, etc., even though they are unavailable. 

#### Screenshot of the result
image
image

#### Checklist

Please disregard the failing "loan-list" test. Sine is working on fixing it.

### DIFF
--- a/src/components/availability-label/availability-labels.tsx
+++ b/src/components/availability-label/availability-labels.tsx
@@ -20,7 +20,7 @@ export interface AvailabilityLabelsProps {
   cursorPointer?: boolean;
 }
 
-export const AvailabiltityLabels: React.FC<AvailabilityLabelsProps> = ({
+export const AvailabilityLabels: React.FC<AvailabilityLabelsProps> = ({
   manifestations,
   workId,
   selectedManifestation: manifestation,
--- a/src/components/material/material-buttons/physical/MaterialButtonsFindOnShelf.tsx
+++ b/src/components/material/material-buttons/physical/MaterialButtonsFindOnShelf.tsx
@@ -1,6 +1,5 @@
 import * as React from "react";
 import { FC } from "react";
-import { useGetAvailabilityV3 } from "../../../../core/fbs/fbs";
 import { useModalButtonHandler } from "../../../../core/utils/modal";
 import { useText } from "../../../../core/utils/text";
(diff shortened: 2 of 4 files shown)


src/components/availability-label/availability-labels.tsx
```diff
@@ -20,7 +20,7 @@ export interface AvailabilityLabelsProps {
   cursorPointer?: boolean;
 }
 
-export const AvailabiltityLabels: React.FC<AvailabilityLabelsProps> = ({
+export const AvailabilityLabels: React.FC<AvailabilityLabelsProps> = ({
```
kasperg, line 23: :+1: Nice to get this corrected even though it is not part of task at hand.

src/components/material/material-buttons/physical/MaterialButtonsFindOnShelf.tsx
```diff
@@ -42,19 +42,6 @@ const MaterialButtonsFindOnShelf: FC<MaterialButtonsFindOnShelfProps> = ({
   }
 
   if (size !== "small") {
-    if (!data[0].available) {
-      return (
-        <Button
-          label={t("materialIsLoanedOutText")}
-          buttonType="none"
```
kasperg, line 49: Please consider if fetching availability data with `useGetAvailabilityV3()` on L24 in this file can be removed now.

As far as I can tell with this change the only place the data was actually used is gone. If that is the case then the request is obsolete.

kasperg (changes requested):
:+1: but I have one important consideration for you before this can be moved along.

Also please check(s) for the Find On Shelf button in our tests. The presence (or lack thereof) of buttons is something we have been going back and forth on frequently.

src/components/material/material-buttons/physical/MaterialButtonsFindOnShelf.tsx
Adamik10: That is absolutely correct. It's removed now.

kasperg (approved):
:+1: LGTM

jeppekroghitk (approved):

PR #296, opened 2022-12-30
Feature/5464 recommender

sinejespersen:
#### Link to issue

https://platform.dandigbib.org/issues/5464

#### Description

Created recommender that:

- Recommends based on a current loan
- If no loan, recommends based on current reservation
- If no reservations, search for something that is configured
- Also changed the favorites aria-label and the way fillstate works (so I can change the color)
- 
#### Screenshot of the result

image

image

#### Checklist


jeppekroghitk (approved):

src/apps/recommender/Recommender.tsx
```diff
@@ -0,0 +1,107 @@
...
+      );
+    }
+    if (fbsErrorReservations && !isSuccessFbsReservations) {
+      setPhysicalReservations([]);
+    }
+  }, [fbsErrorReservations, fbsReservations, isSuccessFbsReservations]);
+
+  useEffect(() => {
+    if (physicalLoans !== null) {
+      const newestLoan = sortByLoanDate([...physicalLoans]).reverse();
```
spaceo, line 63: Why the spreading here?

src/apps/recommender/Recommender.tsx
```diff
@@ -0,0 +1,107 @@
...
+      if (newestLoan.length > 0) {
+        setLoanForRecommender(newestLoan[0]);
+      }
+    }
+    if (physicalReservations !== null) {
+      const newestReservation = sortByReservationDate([
+        ...physicalReservations
```
spaceo, line 70: Why spreading?

src/components/icon-favourite/icon-favourite.tsx
```diff
@@ -2,24 +2,31 @@ import React from "react";
...
       width="24"
-      className="icon-favourite"
+      className={classes}
       viewBox="0 0 24 24"
-      fill={fill ? "#000" : "none"}
       xmlns="http://www.w3.org/2000/svg"
     >
       <path
         d="M11.5 20L10.2675 18.921C5.89 15.1035 3 12.5858 3 9.49591C3 6.9782 5.057 5 7.675 5C9.154 5 10.5735 5.66213 11.5 6.70845C12.4265 5.66213 13.846 5 15.325 5C17.943 5 20 6.9782 20 9.49591C20 12.5858 17.11 15.1035 12.7325 18.9292L11.5 20Z"
-        stroke="black"
```
spaceo, line 22: I know that you have had some talk about this with Kasper already (though I do not know about it in detail).
But there will still be a black outline on the un-checked icon right?

src/core/fetchers/helpers.ts
```diff
@@ -38,4 +39,29 @@ export const createAuthHeader = (
...
+        ", "
+      )}`;
+    } else {
+      returnContentString = `${texts.materialByAuthorText} ${creators
+        .slice(0, -1)
+        .join(", ")} ${texts.materialAndAuthorText} ${creators.slice(-1)}`;
+    }
+  }
+  return returnContentString;
+};
```
spaceo, line 65: Could you turn this into a hook instead, so we can use the `useSelector()` hook instead?
I am not fund of including the store directly if we can avoid it.
I have seen it sometimes creating dependency cycles and it would be nice to have it wrapped in some kind of abstraction.

src/core/utils/helpers/general.ts
```diff
@@ -176,6 +176,14 @@ export const sortByLoanDate = (list: LoanType[]) => {
   );
 };
 
+export const sortByReservationDate = (list: ReservationType[]) => {
+  return list.sort(
+    (objA, objB) =>
+      new Date(objA.dateOfReservation || new Date()).getTime() -
```
spaceo, line 182: When are we using dayjs and when not? Just curious.

src/apps/recommender/RecommendMaterial.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+  // For retrieving cover
+  const manifestationPid = getManifestationPid(manifestations);
+
+  const addToListRequest = (materialId: ButtonFavouriteId) => {
+    dispatch(
+      guardedRequest({
+        type: "addFavorite",
+        args: { materialId },
+        app: "search-result"
```
spaceo, line 48: I think you need the specify the recommender app here, to make the system work in that context only.
So it should be: `app: "recommender"` and specified [here](https://github.com/reload/dpl-react/blob/730d8ac11dcb95a86bb625cd8f36f93e7a873add/src/core/utils/types/ids.ts#L6) as well.
And you need to wrap the app in the guardedApp HOC like the [search-result app is](https://github.com/reload/dpl-react/blob/730d8ac11dcb95a86bb625cd8f36f93e7a873add/src/apps/search-result/search-result.entry.tsx#L76-L78).

spaceo (changes requested):
Great stuff. Have some remarks

src/apps/recommender/Recommender.tsx
sinejespersen: I'll remove

src/components/icon-favourite/icon-favourite.tsx
sinejespersen: image

image

src/core/fetchers/helpers.ts
sinejespersen: I've just moved this function, and I will actually return to this later. I'll add a todo?

src/core/utils/helpers/general.ts
sinejespersen: I'm pretty sure this is a copy paste from another function, where I tried sorting using dayjs and for some reason it did not work. I dont quite remember, but I can try again..?

src/apps/recommender/RecommendMaterial.tsx
sinejespersen: done

src/components/icon-favourite/icon-favourite.tsx
spaceo: 👍

src/apps/material/material.entry.tsx
```diff
@@ -77,6 +77,8 @@ interface MaterialEntryTextProps {
   modalReservationFormSmsInputFieldDescriptionText: string;
   modalReservationFormSmsInputFieldLabelText: string;
   numberDescriptionText: string;
+  addToFavoritesAriaLabelText: string;
+  removeFromFavoritesAriaLabelText: string;
```
spaceo, line 81: Please put them in alphabetical order. Since we have so many texts and we need to use them in dpl-cms also, having them sorted alphabetically helps managing them :)

src/core/fetchers/helpers.ts
spaceo: TODO: Make GitHub Copilot please transform all TODO's into code.
Yes a TODO will DO :)

spaceo (changes requested):
Two more comments and we are there I think 👍

spaceo (approved):
Looks good.
But please fix typescript errors.
And, sorry for bad humor, but this was meant as a joke :) 👎 https://github.com/danskernesdigitalebibliotek/dpl-react/pull/296/commits/1c82d4c07b23a77c18625e443b01a872753de1f5

PR #298, opened 2023-01-02
Introduce lazy loading of search results

spaceo:
#### Link to issue

https://reload.atlassian.net/browse/DDFSOEG-356

#### Description

By using the Intersection Observer API
we can distinguish between search result items that the intersecting with the viewport
and thereby change the behaviour depending on the visibility.

#### Checklist

### DIFF
--- a/src/components/search-result-list/search-result-list-item/search-result-list-item.tsx
+++ b/src/components/search-result-list/search-result-list-item/search-result-list-item.tsx
@@ -28,6 +28,7 @@ import { guardedRequest } from "../../../core/guardedRequests.slice";
 import { Work } from "../../../core/utils/types/entities";
 import { useStatistics } from "../../../core/statistics/useStatistics";
 import { statistics } from "../../../core/statistics/statistics";
+import { useItemHasBeenVisible } from "../../../core/utils/helpers/lazy-load";
 
 export interface SearchResultListItemProps {
   item: Work;
@@ -59,6 +60,9 @@ const SearchResultListItem: React.FC<SearchResultListItemProps> = ({
   const { title: seriesTitle, numberInSeries } = firstInSeries || {};
   const materialFullUrl = constructMaterialUrl(materialUrl, workId as WorkId);
   const { track } = useStatistics();
+  // We use hasBeenVisible to determine if the search result
+  // is, or has been, visible in the viewport.
+  const { itemRef, hasBeenVisible: showItem } = useItemHasBeenVisible();
 
   const handleClick = useCallback(() => {
     track("click", {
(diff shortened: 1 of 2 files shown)


src/core/utils/helpers/lazy-load.ts
```diff
@@ -0,0 +1,29 @@
...
+  });
+  const isInViewPort = Boolean(intersection?.isIntersecting);
+  const [hasBeenVisible, setHasBeenVisible] = useState<boolean | null>(null);
+
+  // We need to track if the item has been visible already.
+  // In that way we can make it stay visible when scrolling back up.
+  useEffect(() => {
+    if (hasBeenVisible || hasBeenVisible !== null) {
+      return;
+    }
```
kasperg, line 19: Consider if we can simplify `hasBeenVisible` to a boolean with an initial value of `false`.

To me it seems as if we never set til to `false` anyway and I cannot see why we would need to distinguish between the two values anyway.

src/core/utils/helpers/lazy-load.ts
```diff
@@ -0,0 +1,29 @@
+import { useEffect, useRef, useState } from "react";
+import { useIntersection } from "react-use";
+
+export const useListItemLazyLoad = () => {
+  const listItemRef = useRef(null);
+  const intersection = useIntersection(listItemRef, {
+    root: null,
+    rootMargin: "0%",
+    threshold: 0.33
```
kasperg, line 9: Consider adding an inline comment explaining why you ended up with 0.33.

I can figure out why we do not use 1 (ever pixel) but why not instead of say 0 which would run if even 1 pixel is visible).

src/core/utils/helpers/lazy-load.ts
```diff
@@ -0,0 +1,29 @@
+import { useEffect, useRef, useState } from "react";
+import { useIntersection } from "react-use";
+
+export const useListItemLazyLoad = () => {
```
kasperg, line 4: Naming here is a bit odd to me. Based on the implementation I cannot tell what the hook has to do with list items or lazy loading.

To me it might as well be called `hasBeenVisible()`, return `{ ref, hasBeenVisible }` and perhaps even accept an `IntersectionObserverInit` object.

Another option might be to add a DocBlock explaining the purpose of the hook.

src/components/search-result-list/search-result-list-item/search-result-list-item.tsx
```diff
@@ -83,32 +86,38 @@ const SearchResultListItem: React.FC<SearchResultListItemProps> = ({
...
-          tint={coverTint}
-        />
+        {isVisible && (
+          <SearchResultListItemCover
+            id={manifestationPid}
+            description={String(fullTitle)}
+            url={materialFullUrl}
+            tint={coverTint}
+          />
+        )}
```
kasperg, line 114: It would be nice if we could render an "empty" cover e.g. just with the color/tint, without trying to load any data.

This could also work for skeletons. Perhaps one could trigger it if no pid is passed to the element.

kasperg (approved):
:+1: LGTM. I have some things for you to consider but in general the implementation looks solid.

src/components/search-result-list/search-result-list-item/search-result-list-item.tsx
spaceo: I don't agree in the visual idea of showing a skeleton item that is partly skeleton partly "real rendering". I like the "full skeleton" expression better. But I think we could ask the customer/designer about this as they are the stakeholders.

I think it works like it is now, where the cover initially shows the tint background and when the cover service returns an image url it is loaded.

src/core/utils/helpers/lazy-load.ts
spaceo: Yes indeed. Brain fart removed.

src/core/utils/helpers/lazy-load.ts
spaceo: We go with 0. One could consider even use a negative value but let's stick to 0!

src/core/utils/helpers/lazy-load.ts
spaceo: Renamed it

sinejespersen (approved):

PR #299, opened 2023-01-03
5555: add color-secondary-gray to elements

sinejespersen:
#### Link to issue

https://platform.dandigbib.org/issues/5555

#### Description

Set color explicitly, due to ios default colors
Removed some aria hiddens that are not supposed to be there 

#### Screenshot of the result

Before: 

image

#### Checklist


jeppekroghitk (approved):

spaceo (approved):
LGTM 👍

PR #300, opened 2023-01-04
Update dbc-gateway codegen and workYear query

kasperbirch1:
#### Rewrite code for FBI-API compatibility

This pull request includes changes to the dbc-gateway codegen and the workYear query to ensure compatibility with the latest API update. We had to rewrite some of the code to reflect the changes in the API and to ensure that the system is functioning correctly.

#### Checklist


kasperg:
Closed in favour of #302 .

PR #301, opened 2023-01-05
Settimeout on app mount

spaceo:
#### Link to issue

https://reload.atlassian.net/browse/DDFSOEG-356

#### Description

We noticed in the Chrome performance tab that the mounting of the different apps are being run in the same task which exceeds the 50ms limit. This breaks up the app mounts in chunks.

#### Checklist

### DIFF
--- a/src/core/mount.js
+++ b/src/core/mount.js
@@ -46,7 +46,7 @@ function mount(context) {
     }
   }
 
-  appContainers.forEach(mountApp);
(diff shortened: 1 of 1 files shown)


Adamik10 (approved):
Looks good to me :)

kasperg (approved):
:+1: LGTM

sinejespersen (approved):

PR #302, opened 2023-01-05
Update FBI API with latest schema changes for JED 0.8

kasperg:
#### Link to issue

https://reload.atlassian.net/browse/DSC-16

#### Description

DBC has updated the schema at [fbi-api.dbc.dk/](https://fbi-api.dbc.dk/) for [JED 0.8](https://danbib.dk/broend3_formater). The update contains some breaking changes which forces us to update our client and code accordingly. Key changes for us are related to `workYear`.

The primary purpose of this PR is to update the generated code and updates our code and Cypress fixtures accordingly.

The PR also contains two other changes:

1. We now see empty `access` properties. The code has been updated to handle this case.
2. The search result test has been updated to use `cy` data attributes.

#### Checklist

### DIFF
--- a/cypress/fixtures/material/fbi-api.json
+++ b/cypress/fixtures/material/fbi-api.json
@@ -3,8 +3,12 @@
     "work": {
       "workId": "work-of:870970-basis:52557240",
       "titles": {
-        "full": ["De syv søstre : Maias historie"],
-        "original": []
+        "full": [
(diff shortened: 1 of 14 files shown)


spaceo (approved):
LGTM 👍

sinejespersen (approved):

PR #305, opened 2023-01-08
Remove hitcount from search header when loading

spaceo:
#### Link to issue

https://reload.atlassian.net/browse/DDFSOEG-371

#### Description
Remove hitcount from search header when loading and add "text-loading" class for animation.

#### Screenshot of the result

ezgif-5-91673e5f53

#### Checklist

Depends on:
https://github.com/danskernesdigitalebibliotek/dpl-design-system/pull/180

There is a loan list test in Cypress that is failing. But as far as I can see it has nothing to do with this PR.

### DIFF
--- a/package.json
+++ b/package.json
@@ -135,7 +135,7 @@
     "prop-types": "Since we use former ddb-react components that depend on prop-types we keep this. Should be removed when usage of prop-types is deprecated."
   },
   "dependencies": {
-    "@danskernesdigitalebibliotek/dpl-design-system": "0.0.0-b7f5caa23580a00859acdfcf292439e365b26660",
+    "@danskernesdigitalebibliotek/dpl-design-system": "npm:@reload/dpl-design-system@0.0.0-58c156e1e0fc725aecd1e7f2bf6d9a8a28dcbe1f",
     "@reach/alert": "^0.17.0",
     "@reach/dialog": "^0.17.0",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -135,7 +135,7 @@
     "prop-types": "Since we use former ddb-react components that depend on prop-types we keep this. Should be removed when usage of prop-types is deprecated."
   },
   "dependencies": {
-    "@danskernesdigitalebibliotek/dpl-design-system": "0.0.0-b7f5caa23580a00859acdfcf292439e365b26660",
+    "@danskernesdigitalebibliotek/dpl-design-system": "npm:@reload/dpl-design-system@0.0.0-58c156e1e0fc725aecd1e7f2bf6d9a8a28dcbe1f",
```
Adamik10, line 138: Don't forget to replace this with the ddf version when that gets merged in.

Adamik10 (approved):
Good stuff :) Approved. 
Possibly you could mention the loading class somewhere in the readme file? Do we have a section on handelling loading - this and skeleton screens could be mentioned there. 

PS: Don't forget about the failing integration test (unless it's the same one Sine is already working on 😉 )

package.json
spaceo: 👍

sinejespersen (approved):

PR #306, opened 2023-01-08
Skeleton screens in search results

spaceo:
#### Link to issue

https://reload.atlassian.net/browse/DDFSOEG-374

#### Description

This introduces skeleton screens from the design system in dpl-react.
And adds the first screens at the search result page.

#### Screenshot of the result

image

#### Checklist

Depends on:
https://github.com/danskernesdigitalebibliotek/dpl-design-system/pull/179

There is a loan list test in Cypress that is failing. But as far as I can see it has nothing to do with this PR.


architecture/adr-006-skeleton-screens.md
```diff
@@ -0,0 +1,36 @@
+# Architecture Decision Record: Skeleton Screens
```
Adamik10, line 1: Documentation is always a 👍

docs/skeleton_screens.md
```diff
@@ -0,0 +1,44 @@
+# Skeleton screens
```
Adamik10, line 1: 👏

src/components/search-result-list/search-result-list-item/search-result-list-item-skeleton.tsx
```diff
@@ -0,0 +1,16 @@
+import React from "react";
+
+const SearchResultListItemSkeleton: React.FC = () => {
```
Adamik10, line 3: cool that it's in its own component 👍

Adamik10 (approved):
Good stuff Mikkel. I am very happy that you documented it so well! - Documentation is great.

yarn.lock
```diff
@@ -1754,10 +1754,10 @@
     debug "^3.1.0"
     lodash.once "^4.1.1"
 
-"@danskernesdigitalebibliotek/dpl-design-system@0.0.0-b7f5caa23580a00859acdfcf292439e365b26660":
-  version "0.0.0-b7f5caa23580a00859acdfcf292439e365b26660"
-  resolved "https://npm.pkg.github.com/download/@danskernesdigitalebibliotek/dpl-design-system/0.0.0-b7f5caa23580a00859acdfcf292439e365b26660/477ea708a0679d53038ca3f8bb29d803e0af40fb#477ea708a0679d53038ca3f8bb29d803e0af40fb"
-  integrity sha512-9AcMD63fA2UvSXObYT8sRenFNNT9Cy+AyQxKaWUdjKIT83PsXzdWX6lKsSbbeKvC9DF5+BSyUSvBRYo0nu7ZpQ==
+"@danskernesdigitalebibliotek/dpl-design-system@npm:@reload/dpl-design-system@0.0.0-6b30a8d9adaecb097a0b655cab37d514dd908ef7":
```
kasperg, line 1757: Remember to remove this change before this is merged.

src/components/facet-line/FacetLine.tsx
```diff
@@ -34,6 +34,11 @@ const FacetLine: React.FunctionComponent<FacetLineProps> = ({
 
   return (
     <>
+      {!data && (
+        <div className="ssc mt-48">
+          <div className="ssc-head-line" />
+        </div>
+      )}
```
kasperg, line 41: I know this is simple but to keep consistency we could consider placing this in a separate `FacetLineFiltersSkeleton` component.

docs/skeleton_screens.md
```diff
@@ -0,0 +1,44 @@
...
+
+The skeleton screens are rendered with help from the
+[skeleton-screen-css library](https://www.npmjs.com/package/skeleton-screen-css?activeTab=readme).
+ By using [ssc classes](https://github.com/nullilac/skeleton-screen-css#example---card)
+ you can easily compose screens
+ that simulate the look of a "real" rendering with real data.
+
+### Simple example
+
+In this example we have a HelloWorld app with three text props attached:
```
kasperg, line 26:
```suggestion
In this example we have a search result item component consisting of a cover, a headline and two lines of text:
```

docs/skeleton_screens.md
```diff
@@ -0,0 +1,44 @@
...
+
+```tsx
+import React from "react";
+
+const SearchResultListItemSkeleton: React.FC = () => {
+  return (
+    <article className="search-result-item ssc">
+      <div className="ssc-square">&nbsp;</div>
+      <div className="ssc-wrapper w-80">
+        <div className="ssc-head-line w-60 mb" />
```
kasperg, line 36: Can we provide any help, hints or guidelines about how developers should use standard classes like `.search-result-item` and skeleton classes?

One example might be: "Use BEM block classes (for layout in relation to other elements etc.) but use SSC classes instead of elements." Note: I do not know if this example will work in practice.

architecture/adr-006-skeleton-screens.md
kasperg: Does (some of) this belong in the design system? Parts of this refer to a package - `skeleton-screen-css` which is not even a direct dependency of this package.

It seems like a decision which really spans the two projects. Using skeleton screens is something related to the async/clientside nature of the React components but the effect of using use `skeleton-screen-css` goes in the design system. We even have stories for skeleton elements in the design system. 🤔

architecture/adr-006-skeleton-screens.md
```diff
@@ -0,0 +1,36 @@
...
+
+## Context
+
+In the work of trying to improve the performance
```
kasperg, line 5: Totally nitpicking but I wonder why these lines of text are so short. Markdownlint enforces a maximum line length of 80 characters but many of these lines are <50 for no reason that I could tell :smile:.

src/components/search-result-list/SearchResultList.tsx
```diff
@@ -10,21 +11,25 @@ export interface SearchResultListProps {
...
-          <SearchResultListItem
-            item={item}
-            coverTint={getCoverTint(i)}
-            resultNumber={i + 1}
-          />
-        </li>
-      ))}
+      {/* Show skeleton search result items if no data is available yet */}
+      {!worksAreLoaded &&
+        [...Array(5)].map(() => (
```
kasperg, line 18: A comment about why 5 might be good. Is it a matter of having enough elements to fill out most user screens?

architecture/adr-006-skeleton-screens.md
```diff
@@ -0,0 +1,36 @@
...
+We could also have chosen to use images or GIF's
+to render the screens. But by using the simple toolbox
+of skeleton-screen-css we should be able to make screens
+for all the different use cases in the different apps.
+
+## Consequences
+
+It is now possible, with a limited amount of work,
+to construct skeleton screens in the loading state
+of the various user interfaces.
```
kasperg, line 36:
```suggestion
It is now possible, with a limited amount of work,
to construct skeleton screens in the loading state
of the various user interfaces.

Because we use library where skeletons are implemented purely in CSS we also provide a solution which can be consumed in any technology already using the design system without any additional dependencies, clientside or serverside.
```

architecture/adr-006-skeleton-screens.md
spaceo: Well another problem connected to the fragmented architecture.
I'll move it to the design system and make a PR there.
Easy peasy no problem.

yarn.lock
spaceo: yes 👍

docs/skeleton_screens.md
spaceo: Done

docs/skeleton_screens.md
spaceo: done

sinejespersen (approved):

PR #310, opened 2023-01-09
Fix accessibility for reservation modal

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFSOEG-372

#### Description
This PR fixes a Pa11y accessibility issue for the reservation modal - not all the IDs are unique. Turns out there were two modal components rendered for every single reservation modal - so when the modal opens there would be a modal inside a modal.

#### Screenshot of the result
image

#### Checklist

Please only review the last 3 commits (with my face on them)
Disregard FBI API changes ( #302 )

### DIFF
--- a/src/components/reservation/ReservationModalBody.tsx
+++ b/src/components/reservation/ReservationModalBody.tsx
@@ -6,7 +6,6 @@ import {
   getManifestationType,
   materialIsFiction
 } from "../../core/utils/helpers/general";
-import Modal from "../../core/utils/modal";
 import { useText } from "../../core/utils/text";
 import { FaustId, WorkId } from "../../core/utils/types/ids";
 import { Button } from "../Buttons/Button";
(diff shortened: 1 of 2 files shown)


src/components/reservation/ReservationModalBody.tsx
```diff
@@ -6,7 +6,6 @@ import {
   getManifestationType,
   materialIsFiction
 } from "../../core/utils/helpers/general";
-import Modal from "../../core/utils/modal";
```
kasperg, line 9: We have talked about cleaning up the commit history as a part of working with a PR. This is an example of a commit that it would be nice to have merged into the previous commit 63c71fc1b1b5449649f46c3de769dae6e6b196c6 which made the import obsolete.

kasperg (approved):
:+1: LGTM but I have only focused on the last 3 commits. Consider rebasing this PR on `develop` now that #302 have landed.

src/core/utils/modal.tsx
```diff
@@ -72,12 +72,12 @@ function Modal({
...
         )}
         role="dialog"
-        aria-labelledby={`modal-${modalId}`}
+        aria-labelledby={`modal-${modalId}-description`}
         data-cy={dataCy}
       >
         <div
           className="modal__screen-reader-description"
-          id={`modal-${modalId}`}
+          id={`modal-${modalId}-description`}
```
sinejespersen, line 80: I am not sure I understand this, why is description added?

Adamik10: That's just because it makes the ID more case specific, since it's used on the element that contains the description. But it's not the key change and I can revert it back if you don't think it makes sense.

sinejespersen (approved):

PR #311, opened 2023-01-11
Feature/handling dom focus

jeppekroghitk:
#### Link to issue

https://platform.dandigbib.org/issues/5533

#### Description

When a modal is opened, the current activeElement in the dom is stored in an array. When said modal is closed, the element is popped from the array, and returned focus.

#### Screenshot of the result

Here is a screenshot of the focus being returned to the second element in the list, after closing the modal it opened:
Screenshot 2023-01-11 at 10 52 14

#### Checklist

### DIFF
--- a/src/apps/loan-list/list/loan-list.test.ts
+++ b/src/apps/loan-list/list/loan-list.test.ts
@@ -326,7 +326,7 @@ describe("Loan list", () => {
       .eq(2)
       .scrollIntoView()
       .find(".list-reservation__information a")
-      .should("be.visible")
+      .should("exist")
       .should(
         "have.text",
         "You will be charged a fee, when the item is returned"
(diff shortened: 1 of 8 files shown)


src/apps/loan-list/list/loan-list.tsx
```diff
@@ -158,7 +158,9 @@ const LoanList: FC<LoanListProps> = ({ pageSize }) => {
   return (
     <>
       <div
-        style={modalIds.length > 0 ? { display: "none" } : {}}
+        style={
+          modalIds.length > 0 ? { overflow: "hidden", height: "100vh" } : {}
```
spaceo, line 162: Wouldn't it be prettier to flip a class instead of hardcoding the style?

src/apps/reservation-list/list/reservation-list.tsx
```diff
@@ -256,27 +256,30 @@ const ReservationList: FC<ReservationListProps> = ({ pageSize }) => {
...
-              reservedReservationsPublizon={reservedReservationsPublizon}
-              pageSize={pageSize}
-            />
-          )}
-        </div>
-      )}
+      <div
+        className="reservation-list-page"
+        style={
+          modalIds.length > 0 ? { overflow: "hidden", height: "100vh" } : {}
```
spaceo, line 262: Wouldn't it be prettier to flip a class instead of hardcoding the style?
And since you are doing the same [here](https://github.com/danskernesdigitalebibliotek/dpl-react/pull/311/files#diff-2cdd62020bea6d73b0ed5c5b7ead29e571bdb06dedcfeb649598dcc202fe42c3R162) I think you should create a function for it.

src/core/modal.slice.ts
```diff
@@ -10,6 +10,17 @@ interface PayloadProps {
...
+const handleModalFocus = (elementToStore?: Element) => {
+  if (elementToStore) {
+    return focusContainerArray.push(elementToStore);
+  }
+  const element = focusContainerArray.pop() as HTMLElement;
+  if (element) {
+    element.focus();
+  }
+  return element;
+};
```
spaceo, line 23: I like the idea of it but I think you could split the function in two functions (names are just suggestions):
SetElementFocus and FocusLatestElement
Just to make clear what is happening.

spaceo: Another question:
What is the idea behind that it is an array?
Why can there be multiple elements in the history?
Would it be more safe and less prone to errors to just have one element?

spaceo (changes requested):
Looks great. Nice initiative. I have some questions/suggestions

src/core/modal.slice.ts
jeppekroghitk: Some modals lead into other modals, which means that we potentially can have x modals open on top of each other at a time. We need to store an array of elements in the order that the respective modals were opened, in order to return the correct element into focus when a modal is closed.
 
I get the idea of splitting the functionality into two seperate functions. One for storing and one for returning. I'll look into that.

sinejespersen (approved):

src/apps/loan-list/list/loan-list.test.ts
```diff
@@ -448,242 +448,242 @@ describe("Loan list", () => {
...
-      .should("have.length", 2);
-
-    // 2.c.iv. If more than 10 loans -> pagination (because of pageSizeDesktop/pageSizeMobile the limit is 2 not 25)
-    cy.get(".loan-list-page").find(".result-pager").should("have.length", 2);
-    cy.get(".list-reservation-container")
-      .eq(1)
-      .find(".list-reservation")
-      .should("have.length", 2);
-  });
+  // it("Loan list basics (digital loans)", () => {
```
spaceo, line 451: Please use `it.skip()` or `it.only` if you want to disable one or more test

src/core/utils/helpers/general.ts
```diff
@@ -320,4 +320,8 @@ export const filterLoansSoonOverdue = (loans: LoanType[]) => {
 export const getManifestationType = (manifestation: Manifestation) =>
   manifestation?.materialTypes?.[0]?.specific;
 
+export const useScrollLock = (modalIds: string[]) => {
```
spaceo, line 323: Im sorry to be picky about this. But it is a convention to prefix custom hooks with "use*". So I think we should avoid using it for other functions. Suggestion: `getScrollClass` ...or whatever you find appropriate.

src/core/modal.slice.ts
spaceo: > Some modals lead into other modals, which means that we potentially can have x modals open on top of each other at a time. We need to store an array of elements in the order that the respective modals were opened, in order to return the correct element into focus when a modal is closed.

Ok. Thank you for the explanation!

> I get the idea of splitting the functionality into two seperate functions. One for storing and one for returning. I'll look into that.

👍

spaceo (approved):
LGTM 👍

PR #314, opened 2023-01-12
Fix loan list test

kasperbirch1:
#### This PR is to fix a failing loan-list test by adding .scrollIntoView() to ensure element visibility for Cypress. 
Also, it renames the loan list test and removes .tsx extension as they are not React components.

#### Checklist


kasperg:
It is worth noting that [this started failing when #305 was merged](https://github.com/danskernesdigitalebibliotek/dpl-react/actions/workflows/integration_test.yml?query=branch%3Adevelop). For some reason actions did not run when the latest commits to this PR were pushed.

jeppekroghitk (approved):

PR #315, opened 2023-01-16
Change incorrect value for coverTints

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFSOEG-389

#### Change incorrect value for coverTints
There was an issue with one of the values for the coverTints css class, which was set to 100. However, there is no css class that fits a tint value of 100. I assume that it should be 60 instead. because it was missing.

image

#### Checklist


spaceo (approved):
LGTM 👍

PR #317, opened 2023-01-17
MK14 ”Se Værkvisnings-siden”: Filmatiseringer vises ikke

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFSOEG-392

#### Description

#### Screenshot of the result

#### Checklist


src/components/material/MaterialDescription.tsx
```diff
@@ -17,20 +17,30 @@ export interface MaterialDescriptionProps {
...
+  const subjectsList = subjects.all.map((item) => {
     return {
       url: constructSearchUrl(searchUrl, item.display),
       term: item.display
     };
   });
-  const { fictionNonfiction } = work;
+
+  const filmAdaptationsList =
+    manifestations.bestRepresentation?.ownerWork.titles.main.map((item) => {
```
kasperg, line 38: Please reconsider this.

The original request says: Look in `relations.hasAdaptation` and get the title from the owner work of one of these manifestations - if available. That is not what has been implemented here.

As I see it this implementation takes the title(s) of the owner work of any best representation and prints that.

src/components/material/MaterialDescription.tsx
```diff
@@ -17,20 +17,30 @@ export interface MaterialDescriptionProps {
...
-  const inSeries = work.series;
-  const seriesMembersList = work.seriesMembers.map((item) => {
+  const { fictionNonfiction, series, subjects, seriesMembers, manifestations } =
+    work;
+
+  const seriesMembersList = seriesMembers.map((item) => {
     return {
       url: constructMaterialUrl(materialUrl, item.workId as WorkId),
-      term: item.titles.main[0]
+      term: item.titles.main.join(" / ")
```
kasperg, line 26: What is this change about?

As far as I can tell it does not have any relation to the change at hand and I do not see any explanation of it in the commit comments either.

kasperg (changes requested):
Based on the conversation in JIRA and my interpretation of the implementation I do not think you got this quite right.

jeppekroghitk (approved):

PR #318, opened 2023-01-18
Promt for login when ordering physical materials

Adamik10:
#### Link to issue
[DSC-9](https://reload.atlassian.net/browse/DSC-9)

#### Description
This PR adds our already developed prompt-for-login functionality to the reservation button for physical materials.

#### Screenshot of the result
n/a

#### Checklist

-

### DIFF
--- a/src/apps/material/login-redirect.test.ts
+++ b/src/apps/material/login-redirect.test.ts
@@ -8,19 +8,12 @@ describe("Material", () => {
       .getBySel("material-header-buttons-physical")
       .click();
 
-    cy.interceptRest({
-      aliasName: "user",
-      url: "**/agencyid/patrons/patronid/v2",
-      fixtureFilePath: "material/user.json"
-    });
-
-    cy.interceptRest({
-      aliasName: "holdings",
-      url: "**/agencyid/catalog/holdings/**",
-      fixtureFilePath: "material/holdings.json"
-    });
-
-    cy.url().should("include", "modal=reservation-modal-46615743");
+    cy.url()
+      .should("include", "modal=reservation-modal-46615743")
+      .then(() => {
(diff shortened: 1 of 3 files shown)


src/apps/material/login-redirect.test.ts
```diff
@@ -0,0 +1,67 @@
...
+
+    cy.interceptRest({
+      aliasName: "holdings",
+      url: "**/agencyid/catalog/holdings/**",
+      fixtureFilePath: "material/holdings.json"
+    });
+
+    cy.url().should("include", "modal=reservation-modal-46615743");
+    cy.getBySel("modal").should("be.visible");
+  });
```
spaceo, line 23: Fails locally both when running:
`yarn test` and `yarn cypress:open`
Does it work in GH actions?

spaceo: Maybe we need a way to intercept the redirect to Adgangsplatformen?

spaceo (changes requested):
have a question that needs to be clarified 🙏

src/apps/material/login-redirect.test.ts
Adamik10: mmmmm works just fine for me
image

As far as I know the setup for testing is that you need to run `yarn start:storybook:test` and then `yarn cypress:open`

Adamik10: Det fungerer på GH også:
image

Adamik10: I think it could be the user token still present before pressing the reserve button (in which case no redirect happens). I made some changes and you can try again 🤞

spaceo (approved):
LGTM 👍

sinejespersen (approved):

PR #319, opened 2023-01-19
Update getWorkDescriptionListData + Run codegen:dbc:gateway

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DSC-23

#### This pull request updates the function getWorkDescriptionListData to include the following changes:

- Modifications to the order of fields.
- Updates to certain fields to contain the desired data.

##### Additionally, the following changes have been made:
- Removal of the hostPublication field and the publisher will now be obtained directly from the manifestation.
- Run codegen:dbc:gateway to update these changes.

**fixtures**:
Reflect the change in the hostPublication field being swapped with the publisher.

**ManifestationsSimple' GraphQL fragment**:

Add the bestRepresentation field to serve as a fallback option for the 'getWorkDescriptionListData' query in cases where no 'currentManifestation' is available.

**'getWorkManifestation' function**:

Updated to accept a type parameter, allowing for the selection of either 'bestRepresentation' or 'latest' representation.

#### Checklist


src/apps/material/helper.ts
```diff
@@ -11,7 +11,13 @@ import {
 import { UseTextFunction } from "../../core/utils/text";
 import { Manifestation, Work } from "../../core/utils/types/entities";
 
-export const getWorkManifestation = (work: Work) => {
+export const getWorkManifestation = (
+  work: Work,
+  type: "bestRepresentation" | "latest"
+) => {
+  if (type === "bestRepresentation") {
```
spaceo, line 18: I think we could create an object const for bestRepresentation" and "latest" globally.
And then it should be possible to make a type of it as well.
Maybe the iidiom could be called ManifestationType?

src/apps/material/helper.ts
```diff
@@ -11,7 +11,13 @@ import {
...
+    return work.manifestations.bestRepresentation as Manifestation;
+  }
   return work.manifestations.latest as Manifestation;
```
spaceo, line 21: Isn't it simple just to return:
```typescript
return work.manifestations[type];
```

src/apps/material/helper.ts
```diff
@@ -48,52 +54,70 @@ export const getWorkDescriptionListData = ({
   const allLanguages = mainLanguages
     .map((language) => language.display)
     .join(", ");
-  const fallBackManifestation = getWorkManifestation(work);
+  const fallBackManifestation = getWorkManifestation(
+    work,
+    "bestRepresentation"
```
spaceo, line 59: Please look at comment above about having a global entity describing the manifestation type

src/apps/material/helper.ts
```diff
@@ -48,52 +54,70 @@ export const getWorkDescriptionListData = ({
...
+    },
+    {
+      label: t("playTimeText"),
       value:
-        (manifestation?.materialTypes?.[0]?.specific && "") ||
-        (fallBackManifestation?.materialTypes?.[0]?.specific && ""),
+        String(manifestation?.physicalDescriptions?.[0]?.playingTime ?? "") ||
+        String(
+          fallBackManifestation?.physicalDescriptions?.[0]?.playingTime ?? ""
+        ),
```
spaceo, line 78: Please create a function for this:
```typescript
const getPlayingTime = (manifestation: Manifestation) => {
  return String(manifestation?.physicalDescriptions?.[0]?.playingTime ?? "");
}
```
and then use it like:
```typescript
getPlayingTime(manifestation) || getPlayingTime(fallBackManifestation)

File: src/apps/material/helper.ts [hunk 5]
```diff
@@ -48,52 +54,70 @@ export const getWorkDescriptionListData = ({
...
       type: "standard"
     },
     {
-      label: t("languageText"),
-      value: allLanguages,
+      label: t("editionText"),
+      value:
+        (manifestation?.edition?.summary ?? "") ||
+        (fallBackManifestation?.edition?.summary ?? ""),
```
spaceo, line 85: Please see previous comment

src/apps/material/helper.ts
```diff
@@ -48,52 +54,70 @@ export const getWorkDescriptionListData = ({
...
       type: "standard"
     },
+
```
spaceo, line 88: Please see previous comment

src/apps/material/helper.ts
```diff
@@ -48,52 +54,70 @@ export const getWorkDescriptionListData = ({
...
       value:
         (manifestation?.genreAndForm?.[0] ?? "") ||
         (fallBackManifestation?.genreAndForm?.[0] ?? ""),
       type: "standard"
     },
-    { label: t("contributorsText"), value: creatorsText, type: "link" },
     {
       label: t("originalTitleText"),
-      value: titles && workYear ? `${titles?.original} ${workYear.year}` : "",
+      value: titles?.original?.[0] ?? "",
```
spaceo, line 98: Could we make a function for this? `getWorkTitle()`?

src/apps/material/helper.ts
```diff
@@ -48,52 +54,70 @@ export const getWorkDescriptionListData = ({
...
       type: "standard"
     },
     {
-      label: t("isbnText"),
+      label: t("publisherText"),
       value:
-        (manifestation?.identifiers?.[0]?.value ?? "") ||
-        (fallBackManifestation?.identifiers?.[0]?.value ?? ""),
+        (manifestation?.publisher.join(" / ") ?? "") ||
+        (fallBackManifestation?.publisher.join(" / ") ?? ""),
```
spaceo, line 105: Please see previous comment about using functions for this

src/apps/material/helper.ts
```diff
@@ -48,52 +54,70 @@ export const getWorkDescriptionListData = ({
...
       type: "standard"
     },
     {
-      label: t("editionText"),
+      label: t("firstEditionYearText"),
+      value: String(workYear?.year ?? t("unknownText")),
```
spaceo, line 110: Both text identifiers are pretty generic, especially "unknownText". Could the be prefixd with the context?
Please remember to put them in the proptypes of the app.

src/apps/material/helper.ts
```diff
@@ -48,52 +54,70 @@ export const getWorkDescriptionListData = ({
...
+      type: "standard"
+    },
+    {
+      label: t("typeText"),
       value:
-        (manifestation?.edition?.summary ?? "") ||
-        (fallBackManifestation?.edition?.summary ?? ""),
+        (manifestation?.materialTypes?.[0]?.specific ?? "") ||
+        (fallBackManifestation?.materialTypes?.[0].specific ?? ""),
```
spaceo, line 117: See previous comment about functions

src/apps/material/material.tsx
```diff
@@ -110,7 +110,7 @@ const Material: React.FC<MaterialProps> = ({ wid }) => {
     const type = getUrlQueryParam("type");
     // if there is no type in the url, getWorkManifestation is used to set the state and url type parameters
     if (!type) {
-      const workManifestation = getWorkManifestation(work);
+      const workManifestation = getWorkManifestation(work, "latest");
```
spaceo, line 113: See previous comment about manifestation type

spaceo (changes requested):
Looks good. I have some remarks though

Adamik10 (commented):
LGTM 👍 
The only comment I have is that I think the fixtures are unnecessarily long. I normally just delete parts that are not necessary for the tests so the files aren't giant.

spaceo (approved):
LGTM :+1

jeppekroghitk (approved):

PR #324, opened 2023-01-24
Feature/5591 thresholds configurable

sinejespersen:
#### Link to issue

https://platform.dandigbib.org/issues/5591

#### Description

- clean up some props 
- add config dangerThresholdConfig and warningThresholdConfig
#### Screenshot of the result

#### Checklist

### DIFF
--- a/src/apps/intermediate-list/stackable-fees/fee-status.tsx
+++ b/src/apps/intermediate-list/stackable-fees/fee-status.tsx
@@ -2,7 +2,6 @@ import dayjs from "dayjs";
 import React, { FC } from "react";
 import { dateFormatDefault } from "../../../core/configuration/date-format.json";
 import { useText } from "../../../core/utils/text";
-import StatusBadge from "../../loan-list/materials/utils/status-badge";
 
 interface FeeStatusProps {
   dueDate: string;
@@ -15,7 +14,7 @@ const FeeStatus: FC<FeeStatusProps> = ({ dueDate, reasonMessage }) => {
   return (
     <div>
       <div className="list-reservation__deadline">
-        <StatusBadge dueDate={dueDate} dangerText={reasonMessage} />
+        <div className="status-label status-label--danger">{reasonMessage}</div>
         <p className="text-small-caption" id="due-date">
           {t("feeCreatedText", {
             placeholders: { "@date": dueDateFormatted }
--- a/src/apps/loan-list/list/loan-list.dev.tsx
+++ b/src/apps/loan-list/list/loan-list.dev.tsx
@@ -28,13 +28,10 @@ export default {
       control: { type: "number" }
     },
     // Config
-    dangerThresholdConfig: {
-      defaultValue: "0",
+    thresholdConfig: {
+      defaultValue:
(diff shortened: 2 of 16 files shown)


src/apps/intermediate-list/stackable-fees/fee-status.tsx
```diff
@@ -15,7 +14,7 @@ const FeeStatus: FC<FeeStatusProps> = ({ dueDate, reasonMessage }) => {
   return (
     <div>
       <div className="list-reservation__deadline">
-        <StatusBadge dueDate={dueDate} dangerText={reasonMessage} />
+        <div className="status-label status-label--danger">{reasonMessage}</div>
```
spaceo, line 17: This seems like a generic way to show a "danger message" and it is used multiple places?
In that case I think it should be a seperate component we can reuse.

spaceo (changes requested):
Looks good. One thing I would like to be addressed...

src/apps/loan-list/list/loan-list.dev.tsx
```diff
@@ -27,6 +27,15 @@ export default {
       defaultValue: 5,
       control: { type: "number" }
     },
+    // Config
+    dangerThresholdConfig: {
+      defaultValue: "0",
+      control: { type: "text" }
+    },
+    warningThresholdConfig: {
```
spaceo, line 35: I think the config keys should be more specific. They are globally shared (like the text config keys).

src/apps/intermediate-list/stackable-fees/fee-status.tsx
sinejespersen: I can change it back to using the generic component, if you like, it just doesnt ever change color (the fee-status), which is sort of what StatusBadge is good at

src/apps/loan-list/list/loan-list.dev.tsx
sinejespersen: any idea for a name? I think they are a little generic, but they should also be the same globally. there will not be separate configurations for them in drupal.

jeppekroghitk (approved):

src/apps/loan-list/list/loan-list.dev.tsx
spaceo: Ok. If it is that global then how about making just one configuration config object?
Since it is possible to use a json string we could have one config object called
`patronAlertTresholds`. Does that make sense?
branchesConfig is an example you can look at:
https://github.com/reload/dpl-react/blob/e452336fbb07371befe258ce2cc8fcac81e5577b/src/apps/search-result/search-result.dev.tsx#L93-L98
You can parse the config like this:
https://github.com/reload/dpl-react/blob/e452336fbb07371befe258ce2cc8fcac81e5577b/src/apps/patron-page/util/BranchesDropdown.tsx#L21-L23

sinejespersen: In one of the components, only one of the two are used

spaceo (approved):
LGTM :+1

PR #325, opened 2023-01-24
Add data-cy property for cypress testing to search header components

Adamik10:
#### Link to issue
[DDFSOEG-307](https://reload.atlassian.net/browse/DDFSOEG-307)

#### Description
This PR adds data-cy property for cypress testing to search header components so that they can be easily targeted in our cypress tests moving forward. (Prompted by [this PR](https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/216) )

#### Screenshot of the result
n/a

#### Checklist

-

### DIFF
--- a/src/components/autosuggest-text/autosuggest-text-item.tsx
+++ b/src/components/autosuggest-text/autosuggest-text-item.tsx
@@ -12,6 +12,7 @@ export interface AutosuggestTextItemProps {
   index: number;
   generateItemId: (objectItem: Suggestion) => string;
   getItemProps: UseComboboxPropGetters<Suggestion>["getItemProps"];
+  dataCy?: string;
 }
 
 const AutosuggestTextItem: React.FC<AutosuggestTextItemProps> = ({
@@ -19,7 +20,8 @@ const AutosuggestTextItem: React.FC<AutosuggestTextItemProps> = ({
   item,
   index,
   generateItemId,
-  getItemProps
+  getItemProps,
+  dataCy = "autosuggest-text-item"
 }) => {
   const t = useText();
   return (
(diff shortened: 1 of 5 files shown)


src/components/autosuggest-text/autosuggest-text-item.tsx
```diff
@@ -30,6 +30,7 @@ const AutosuggestTextItem: React.FC<AutosuggestTextItemProps> = ({
         className={classes.textSuggestion}
         key={generateItemId(item)}
         {...getItemProps({ item, index })}
+        data-cy="autosuggest-text-item"
```
spaceo, line 33: Normally when it is a component that needs a data-cy attribute on the outer/first level we make it possible to overwrite via a prop. Like [this](https://github.com/reload/dpl-react/blob/cdfa637892e619a032f5a924e92fd7f5964cf328/src/components/availability-label/availability-label.tsx#L32) and [this](https://github.com/reload/dpl-react/blob/cdfa637892e619a032f5a924e92fd7f5964cf328/src/components/availability-label/availability-label.tsx#L93-L94). In that way it canvary according to the context it is used in.

src/components/autosuggest/autosuggest.tsx
```diff
@@ -49,6 +49,7 @@ export const Autosuggest: React.FC<AutosuggestProps> = ({
         className="autosuggest pb-16"
         {...getMenuProps()}
         style={!isOpen ? { display: "none" } : {}}
+        data-cy="autosuggest"
```
spaceo, line 52: Please see previous comment

spaceo (changes requested):
Looks. Tiny remark

jeppekroghitk (approved):

spaceo (approved):

PR #326, opened 2023-01-25
Feature/5408 something like this

sinejespersen:
#### Link to issue

#### Description

#### Screenshot of the result

#### Checklist

### DIFF
--- a/src/apps/something-similar/SomethingSimilar.entry.tsx
+++ b/src/apps/something-similar/SomethingSimilar.entry.tsx
@@ -1,0 +1,21 @@
+import React, { FC } from "react";
+import SomethingSimilar from "./SomethingSimilar";
+import { withText } from "../../core/utils/text";
+import { withConfig } from "../../core/utils/config";
+import { withUrls } from "../../core/utils/url";
+
+export interface SomethingSimilarProps {
+  fbsBaseUrl: string;
+  publizonBaseUrl: string;
+  dplCmsBaseUrl: string;
+  coverBaseUrl: string;
+  materialBaseUrl: string;
+  fbiBaseUrl: string;
+  materialUrl: string;
+}
+
+const SomethingSimilarEntry: FC<SomethingSimilarProps> = () => (
+  <SomethingSimilar />
+);
+
+export default withUrls(withConfig(withText(SomethingSimilarEntry)));
--- a/src/apps/something-similar/SomethingSimilar.entry.tsx
+++ b/src/apps/something-similar/SomethingSimilar.entry.tsx
@@ -12,6 +12,11 @@ export interface SomethingSimilarProps {
   materialBaseUrl: string;
   fbiBaseUrl: string;
   materialUrl: string;
+  somethingSimilarTitleText: string;
+  materialByAuthorText: string;
+  materialAndAuthorText: string;
+  somethingSimilarByTheSameAuthorText: string;
+  somethingSimilarSomethingSimilarAuthorText: string;
 }
 
 const SomethingSimilarEntry: FC<SomethingSimilarProps> = () => (
--- a/src/apps/something-similar/SomethingSimilar.tsx
+++ b/src/apps/something-similar/SomethingSimilar.tsx
@@ -1,0 +1,27 @@
+import React, { FC, useEffect, useState } from "react";
+import { getWorkIdFromQueryParam } from "../../core/utils/helpers/url";
+import { FaustId } from "../../core/utils/types/ids";
+import SomethingSimilarList from "./SomethingSimilarList";
+
+const SomethingSimilar: FC = () => {
+  const [loanForRecommender, setLoanForRecommender] = useState<string | null>(
+    null
+  );
+
+  useEffect(() => {
+    setLoanForRecommender(getWorkIdFromQueryParam());
+  }, []);
+
+  return (
+    <div className="recommender recommender--bright">
+      {loanForRecommender && (
+        <SomethingSimilarList
+          faust={loanForRecommender as FaustId}
+          id={loanForRecommender as FaustId}
+        />
+      )}
+    </div>
+  );
+};
+
+export default SomethingSimilar;
--- a/src/apps/something-similar/SomethingSimilarList.tsx
+++ b/src/apps/something-similar/SomethingSimilarList.tsx
@@ -1,0 +1,98 @@
+import React, { useEffect, FC, useState } from "react";
+import {
+  useRecommendFromFaustQuery,
+  RecommendFromFaustQuery,
+  useSearchWithPaginationQuery,
+  SearchWithPaginationQuery
+} from "../../core/dbc-gateway/generated/graphql";
+import { useText } from "../../core/utils/text";
+import { Work } from "../../core/utils/types/entities";
+import fetchMaterial, {
+  MaterialProps
+} from "../loan-list/materials/utils/material-fetch-hoc";
+import RecommendMaterial from "../recommender/RecommendMaterial";
+
+export interface LikeThisListProps {
+  id: string;
+}
+
+const LikeThisList: FC<LikeThisListProps & MaterialProps> = ({
+  id,
+  material
+}) => {
+  const t = useText();
+  const [recommendView, setRecommendView] = useState<boolean>(true);
+  const { data: somethingSimilarData } = useRecommendFromFaustQuery({
+    faust: id,
+    limit: 4
+  });
+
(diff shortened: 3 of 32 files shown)


jeppekroghitk (approved):

src/apps/something-similar/SomethingSimilar.entry.tsx
```diff
@@ -0,0 +1,26 @@
...
+import { withUrls } from "../../core/utils/url";
+
+export interface SomethingSimilarProps {
+  fbsBaseUrl: string;
+  publizonBaseUrl: string;
+  dplCmsBaseUrl: string;
+  coverBaseUrl: string;
+  materialBaseUrl: string;
+  fbiBaseUrl: string;
+  materialUrl: string;
```
spaceo, line 14: Think we should have a global interface for the serviceUrls

src/apps/something-similar/SomethingSimilar.entry.tsx
```diff
@@ -0,0 +1,26 @@
...
+  somethingSimilarTitleText: string;
+  materialByAuthorText: string;
+  materialAndAuthorText: string;
+  somethingSimilarByTheSameAuthorText: string;
+  somethingSimilarSomethingSimilarAuthorText: string;
+}
+
+const SomethingSimilarEntry: FC<SomethingSimilarProps> = () => (
+  <SomethingSimilar />
```
spaceo, line 23: I think you should use wid as a prop here instead of trying to get it from the url later on. It brings the flexibility of being able to use the app in other contexts as well.
We did the same in the [work/material app](https://github.com/danskernesdigitalebibliotek/dpl-react/blob/f6b21e6ebf7ccc0eab3e0ba03326b32d06a8660b/src/apps/material/material.entry.tsx#L152-L156)

src/apps/something-similar/SomethingSimilar.tsx
```diff
@@ -0,0 +1,27 @@
...
+  useEffect(() => {
+    setLoanForRecommender(getWorkIdFromQueryParam());
+  }, []);
+
+  return (
+    <div className="recommender recommender--bright">
+      {loanForRecommender && (
+        <SomethingSimilarList
+          faust={loanForRecommender as FaustId}
+          id={loanForRecommender as FaustId}
```
spaceo, line 20: It can be solved differently than use type assertions here

src/apps/something-similar/SomethingSimilar.tsx
```diff
@@ -0,0 +1,27 @@
...
+import { FaustId } from "../../core/utils/types/ids";
+import SomethingSimilarList from "./SomethingSimilarList";
+
+const SomethingSimilar: FC = () => {
+  const [loanForRecommender, setLoanForRecommender] = useState<string | null>(
+    null
+  );
+
+  useEffect(() => {
+    setLoanForRecommender(getWorkIdFromQueryParam());
```
spaceo, line 12: Please see previous comment about feeding the app with a work id instead of fetching it from the url

src/apps/something-similar/SomethingSimilarList.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+const LikeThisList: FC<LikeThisListProps & MaterialProps> = ({
+  id,
+  material
+}) => {
+  const t = useText();
+  const [recommendView, setRecommendView] = useState<boolean>(true);
+  const { data: somethingSimilarData } = useRecommendFromFaustQuery({
+    faust: id,
+    limit: 4
```
spaceo, line 27: This should be a global configuration/prop

src/apps/something-similar/SomethingSimilarList.tsx
```diff
@@ -0,0 +1,97 @@
...
+  });
+
+  const { data: byAuthorData } = useSearchWithPaginationQuery({
+    limit: 4,
```
spaceo, line 31: This should be a global configuration/prop

src/apps/something-similar/SomethingSimilarList.tsx
```diff
@@ -0,0 +1,97 @@
...
+    q: {
+      all: material?.firstAuthor
+    },
+    offset: 0
+  });
+
+  const [somethingSimilar, setSomethingSimilar] =
```
spaceo, line 38: It is more of a cosmetic thing, but I think it would be nice to group the useState hooks in the top.

src/apps/something-similar/SomethingSimilarList.tsx
```diff
@@ -0,0 +1,97 @@
...
+    }
+  }, [byAuthorData]);
+
+  return (
+    <>
+      <h2 className="recommender__title text-header-h1 recommender__title--left">
+        {t("somethingSimilarTitleText")}
+      </h2>
+      <div className="recommender__buttons">
+        <button
```
spaceo, line 62: Isn't it possible to use our global button component?

src/apps/something-similar/SomethingSimilarList.tsx
```diff
@@ -0,0 +1,97 @@
...
+          type="button"
+          onClick={() => setRecommendView(true)}
+          className={`button-link button-link--bright ${
+            recommendView ? "button-link--selected" : ""
+          }`}
+        >
+          {t("somethingSimilarSomethingSimilarAuthorText")}
+        </button>
+        <button
```
spaceo, line 71: Same as prev.

src/apps/something-similar/SomethingSimilarList.tsx
```diff
@@ -0,0 +1,97 @@
...
+          }`}
+        >
+          {t("somethingSimilarByTheSameAuthorText")}
+        </button>
+      </div>
+      <ul className="recommender__grid">
+        {recommendView &&
+          somethingSimilar &&
+          somethingSimilar.recommend.result.map(({ work }) => (
+            <RecommendMaterial bright work={work as Work} />
```
spaceo, line 85: Should be possible to map the Work type to the data earlier on

src/apps/something-similar/SomethingSimilarList.tsx
```diff
@@ -0,0 +1,97 @@
...
+          ))}
+        {!recommendView &&
+          authorMaterials &&
+          authorMaterials.search.works.map((work) => (
+            <RecommendMaterial bright work={work as Work} />
```
spaceo, line 90: Should be possible to map the Work type to the data earlier on

src/core/utils/helpers/url.ts
```diff
@@ -23,6 +23,19 @@ export const getUrlQueryParam = (param: string): null | string => {
...
+  const regexIdentifier = /(?<=work-of:\d{6}-.*:)(\d{8})/g;
+  const workId = decodedeQueryParams.match(regexIdentifier);
+
+  if (workId) {
+    const [returnId] = workId;
+    return returnId;
+  }
+  return "";
+};
+
```
spaceo, line 38: I don't think we need this function because we can use work id as a prop.

spaceo (changes requested):
You are spawning new apps. Great :)
I have some comments though.

src/apps/something-similar/SomethingSimilar.entry.tsx
sinejespersen: done

src/apps/something-similar/SomethingSimilarList.tsx
sinejespersen: yep

src/apps/something-similar/SomethingSimilarList.tsx
sinejespersen: yep

src/apps/something-similar/SomethingSimilarList.tsx
sinejespersen: It does not look like a button but a link :)

src/apps/something-similar/SomethingSimilarList.tsx
sinejespersen: It does not look like a button but a link :)

src/apps/something-similar/SomethingSimilarList.tsx
sinejespersen: yep

src/apps/something-similar/SomethingSimilarList.tsx
sinejespersen: yep

spaceo (approved):
LGTM 👍

PR #331, opened 2023-01-26
use 'getNumberedSeries' to render series on manifestation and search results 

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFSOEG-395

#### This pull request adds a new function 'getNumberedSeries'
'getNumberedSeries' takes in an array of objects (series) and returns a sorted version of the array based on the 'numberInSeries' property. It first filters out any objects that do not have a 'numberInSeries' property, and then sorts the remaining objects based on the numerical value of their 'numberInSeries' property.

#### Screenshot of the result
**After**:
image

**Before**:
image

#### Checklist

### DIFF
--- a/src/apps/material/helper.ts
+++ b/src/apps/material/helper.ts
@@ -350,3 +350,18 @@ export const getBestMaterialTypeForWork = (work: Work) => {
 export const reservationModalId = (faustIds: FaustId[]) => {
   return constructModalId("reservation-modal", faustIds.sort());
 };
+
+export const getNumberedSeries = (series: Work["series"]) => {
+  // Filter out series that don't have a 'numberInSeries' property
+  const seriesWithNumber = series.filter(
+    (serie) => serie.numberInSeries?.number
+  );
+
+  // Sort the series based on their 'numberInSeries' property
+  return seriesWithNumber.sort((a, b) => {
+    if (a?.numberInSeries?.number && b?.numberInSeries?.number) {
+      return Number(a.numberInSeries.number) - Number(b.numberInSeries.number);
+    }
+    return 0;
+  });
+};
--- a/src/apps/material/helper.ts
+++ b/src/apps/material/helper.ts
@@ -352,15 +352,4 @@ export const reservationModalId = (faustIds: FaustId[]) => {
 };
 
 export const getNumberedSeries = (series: Work["series"]) =>
-  series
-    // Filter out series that don't have a 'numberInSeries' property
-    .filter((serie) => serie.numberInSeries?.number)
-    // Sort the list in chronological order to prioritize the most representative series items first.
-    .sort((a, b) => {
-      if (a?.numberInSeries?.number && b?.numberInSeries?.number) {
-        return (
-          Number(a.numberInSeries.number) - Number(b.numberInSeries.number)
-        );
-      }
-      return 0;
-    });
+  series.filter((serie) => serie.numberInSeries?.number);
--- a/src/apps/material/helper.ts
+++ b/src/apps/material/helper.ts
@@ -352,4 +352,4 @@ export const reservationModalId = (faustIds: FaustId[]) => {
 };
 
 export const getNumberedSeries = (series: Work["series"]) =>
-  series.filter((serie) => serie.numberInSeries?.number);
+  series.filter((seriesEntry) => seriesEntry.numberInSeries?.number);
--- a/src/apps/material/helper.ts
+++ b/src/apps/material/helper.ts
@@ -351,17 +351,16 @@ export const reservationModalId = (faustIds: FaustId[]) => {
   return constructModalId("reservation-modal", faustIds.sort());
 };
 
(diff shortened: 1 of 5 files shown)


src/components/material/MaterialDescription.tsx
```diff
@@ -42,11 +42,11 @@ const MaterialDescription: React.FC<MaterialDescriptionProps> = ({ work }) => {
       )}
       <div className="material-description__links mt-32">
         {inSeries &&
-          inSeries.map((seriesItem, i) => {
-            return (
+          inSeries.map((seriesItem, i) =>
+            seriesItem.numberInSeries?.number ? (
```
spaceo, line 46: I think the same is happening in the [search result item](https://github.com/reload/dpl-react/blob/730d8ac11dcb95a86bb625cd8f36f93e7a873add/src/components/search-result-list/search-result-list-item/search-result-list-item.tsx#L122-L135)?
 I think you should put an effort in making a component fro showing this part, which can be used in both places.

spaceo (changes requested):
Looks good. But I think you need to refactor stuff.

src/components/material/MaterialDescription.tsx
```diff
@@ -17,20 +18,23 @@ export interface MaterialDescriptionProps {
...
+  const seriesList = hasNumberInSeries(series);
+
+  const seriesMembersList = seriesMembers.map((item) => {
     return {
       url: constructMaterialUrl(materialUrl, item.workId as WorkId),
-      term: item.titles.main[0]
+      term: item.titles.main.join(" / ")
     };
   });
-  const subjectsList = work.subjects.all.map((item) => {
```
kasperg, line 27: These somewhat unrelated changes will conflict with changes in #317 😕 .

src/core/utils/helpers/general.ts
```diff
@@ -345,4 +345,19 @@ export const dataIsNotEmpty = (data: unknown[]) => Boolean(data.length);
...
+  );
+
+  // Sort the series based on their 'numberInSeries' property
+  return seriesWithNumber.sort((a, b) => {
+    if (a?.numberInSeries?.number && b?.numberInSeries?.number) {
+      return Number(a.numberInSeries.number) - Number(b.numberInSeries.number);
+    }
+    return 0;
+  });
+};
```
kasperg, line 361: Please reconsider the this function:

1. The naming is off. Functions starting with `has` or `is` typically return a boolean. That is not the case here. Better examples would be `numberedSeries` or `getNumberedSeries`.
2. Personally I do not understand *why* we need to do any sorting. It is not mentioned anywhere in the requirements. As I see it we should be fine working with any numbered series entry.
3. This function is placed in `utils/helpers/general.ts`. It could not get any more generic :smile:. It would be good to move all our work/manifestation helpers to a separate file. Note that moving this around may be outside the scope of this PR.

src/components/search-result-list/search-result-list-item/search-result-list-item.tsx
```diff
@@ -56,8 +57,7 @@ const SearchResultListItem: React.FC<SearchResultListItemProps> = ({
     t
   );
   const manifestationPid = getManifestationPid(manifestations);
-  const firstInSeries = series?.[0];
-  const { title: seriesTitle, numberInSeries } = firstInSeries || {};
+  const firstItemInSeries = hasNumberInSeries(series)[0];
```
kasperg, line 60: Consider using `shift()` instead of indexed access.

This should ensure that Typescript sees the return value as `Work["series"] | undefined`. You add a proper check for this in L122 anyway so we should be good but using `shift()` will let TypeScript help you remember this.

kasperg (approved):
:+1: This needs a little more consideration but afterwards you can push this along.

src/apps/material/material.test.ts
```diff
@@ -41,7 +41,7 @@ describe("Material", () => {
 
     cy.visit("/iframe.html?id=apps-material--default&viewMode=story&type=bog");
 
-    cy.getBySel("material-description-series-1")
+    cy.getBySel("material-description-series-0")
```
kasperg, line 44: This test specifically tests for series. Consider updating the title to reflect that.

src/apps/material/helper.ts
```diff
@@ -350,3 +350,6 @@ export const getBestMaterialTypeForWork = (work: Work) => {
 export const reservationModalId = (faustIds: FaustId[]) => {
   return constructModalId("reservation-modal", faustIds.sort());
 };
+
+export const getNumberedSeries = (series: Work["series"]) =>
+  series.filter((serie) => serie.numberInSeries?.number);
```
sinejespersen, line 355:
```suggestion
  series.filter((seriesEntry) => seriesEntry.numberInSeries?.number);
```

sinejespersen (approved):

spaceo (approved):

PR #333, opened 2023-01-26
Test/material buttons

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFSOEG-375

#### Description
This PR introduces the material-buttons cypress test that checks whether the material page action/reservation buttons and the find on shelf button behave as they should. 
I also adjusted and added some fixtures to e.g. be able to reuse material page fbi-api.json fixture in my test. 

#### Screenshot of the result
image

#### Checklist

n/a

### DIFF
--- a/src/components/material/material-buttons/online/MaterialButtonOnlineExternal.tsx
+++ b/src/components/material/material-buttons/online/MaterialButtonOnlineExternal.tsx
@@ -8,6 +8,7 @@ import { getMaterialTypes } from "../../../../core/utils/helpers/general";
 import { useText } from "../../../../core/utils/text";
 import { ButtonSize } from "../../../../core/utils/types/button";
 import { Manifestation } from "../../../../core/utils/types/entities";
+import MaterialTypes from "../../../../core/utils/types/material-type";
 import { LinkNoStyle } from "../../../atoms/link-no-style";
 import { Button } from "../../../Buttons/Button";
 
@@ -32,7 +33,9 @@ export const getOnlineMaterialType = (
     return "emovie";
   }
   if (
-    materialTypes.find((element) => element.toLowerCase().includes("lydbog"))
+    materialTypes.find((element) =>
+      element.toLowerCase().includes(MaterialTypes.audioBookGeneric)
+    )
   ) {
     return "audiobook";
   }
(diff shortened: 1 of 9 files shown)


src/components/material/material-buttons/online/MaterialButtonOnlineExternal.tsx
```diff
@@ -25,13 +25,15 @@ export const getOnlineMaterialType = (
...
     return "ebook";
   }
-  if (sourceName.includes("filmstriben")) {
+  if (sourceName.toLowerCase().includes("filmstriben")) {
     return "emovie";
   }
-  if (materialTypes.find((element) => element.includes("lydbog"))) {
+  if (
+    materialTypes.find((element) => element.toLowerCase().includes("lydbog"))
+  ) {
```
spaceo, line 36: Don't we have a global place for these material types? I think we should use that

spaceo (changes requested):
Looks good. One comment...

src/components/material/material-buttons/online/MaterialButtonOnlineExternal.tsx
Adamik10: I mean, no, but it defo could be added ;)

src/components/material/material-buttons/online/MaterialButtonOnlineExternal.tsx
```diff
@@ -44,7 +49,7 @@ const MaterialButtonOnlineExternal: FC<MaterialButtonOnlineExternalProps> = ({
   size,
   trackOnlineView,
   manifestations,
-  dataCy = "material-button-online-external"
+  dataCy = "material-buttons-online-external"
```
spaceo, line 52: Why is this changed, btw?
This component is called MaterialButtonOnlineExternal (button in singular)
and the cy attribute is passed along to the Button component (which is also singular/one button.
It is not a biggie i just wondered.
I have tried in the other cases to follow a pattern where the cy attribute is named after the component, in kebab case of course. This would be a deviance of that pattern

Adamik10: It is because the dataCy lineage starts in the component called MaterialButtons, and it passes down a string called "material-buttons" which is then appended "-online" to, which is then appended "-external" to in this component. With my change the string is more consistent as it in all circumstances stays the same for the component.

spaceo (approved):
You agreed on changing the dataCy default value back again.
But in any case it is not a blocker and therefor approved.

src/components/material/material-buttons/online/MaterialButtonOnlineExternal.tsx
Adamik10: Gotcha ;)

jeppekroghitk (approved):

PR #334, opened 2023-01-26
Editions grouped by type and then by age

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DSC-31

#### This pull request adds a new function getManifestationsOrderByTypeAndYear that sorts an array of Manifestation objects by year and material type.

- Added a new function getManifestationsOrderByTypeAndYear that takes an array of Manifestation objects as an argument.
- Used two helper functions, orderManifestationsByYear and groupBy, to sort the array of Manifestation objects.
- Returned a new array with the sorted Manifestation objects.

#### Screenshot of the result

image

#### Checklist

### DIFF
--- a/src/apps/material/helper.ts
+++ b/src/apps/material/helper.ts
@@ -30,10 +30,12 @@ export const getManifestationsOrderByTypeAndYear = (
 
   const materialsMappedBytype = groupBy(
     orderedByYear,
-    "materialTypes[0].specific"
+    // all manifestations that not have a material type will be grouped under "unknown"
+    (m) => m?.materialTypes[0]?.specific ?? "unknown"
   );
 
   return (
+    // Get the keys for each material type.
     Object.keys(materialsMappedBytype)
       // Sort the material types alphabetically.
       .sort()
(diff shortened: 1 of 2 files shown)


src/apps/material/helper.ts
```diff
@@ -23,6 +23,28 @@ export const getLatestWorkManifestation = (work: Work) => {
...
 };
 
+export const getManifestationsOrderByTypeAndYear = (
+  manifestations: Manifestation[]
+) => {
+  const orderedByYear = orderManifestationsByYear(manifestations);
+
+  const materialsMappedBytype = groupBy(
+    orderedByYear,
+    "materialTypes[0].specific"
```
spaceo, line 33: I think the type should be an argument to the function. You could use [this](https://github.com/reload/dpl-react/blob/730d8ac11dcb95a86bb625cd8f36f93e7a873add/src/core/utils/types/material-type.ts#L1) with `typeof`

spaceo (approved):
Looks good. On e minor comment

spaceo (changes requested):
One comment ;)

spaceo (approved):
LGTM 👍

PR #340, opened 2023-01-31
Performance optimisations

spaceo:
#### Link to issue

https://reload.atlassian.net/browse/DDFSOEG-400

#### Description

* Material page skeleton screens
* Material cover preload tag
* Lazy loading of info in disclosures and modals on material page

#### Checklist

#### Additional notes
This PR is depending on a dpl-design-system PR. When the design system has been merged the package.json dependency refrence will be updated.


src/apps/material/login-redirect.test.ts
```diff
@@ -5,15 +5,20 @@ describe("Material", () => {
     window.sessionStorage.removeItem("user");
 
     cy.visit("/iframe.html?id=apps-material--default&type=bog&")
+      .getBySel("material-description")
+      .scrollIntoView()
```
kasperg, line 9: Consider if we can use `material-header-content` instead of `material-description`.

I am a bit surprised that we need to scroll to the description )which I think would be far down the page) when we want to use the reservation buttons which are placed in the header. 

Note that this occurs in many places throughout this PR.

src/apps/material/material.tsx
```diff
@@ -160,112 +157,122 @@ const Material: React.FC<MaterialProps> = ({ wid }) => {
...
-      </Disclosure>
-      {reviews && reviews.length >= 1 && (
-        <DisclosureControllable
-          id="reviews"
-          title={t("reviewsText")}
-          mainIconPath={CreateIcon}
-          showContent={shouldOpenReviewDisclosure}
-          cyData="material-reviews-disclosure"
+    <>
+      <CoverPreloadLink id={pid} size="xlarge" />
```
kasperg, line 161: Please consider if preloading covers here is actually useful.

From the documentation:

> The preload value of the [<link>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/link) element's [rel](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/link#attr-rel) attribute lets you declare fetch requests in the HTML's [<head>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/head), specifying resources that your page will need very soon, which you want to start loading early in the page lifecycle, before browsers' main rendering machinery kicks in.

This places the preload link in the HTML body. With that in mind I would not think it made any difference. On the other hand it adds another request (if it is not cached by Orval) and in any case added complexity.

We might be able to move the `<link>` tag to the `<head>` with [help from third party libraries](https://github.com/nfl/react-helmet) but it would render clientside anyway and I do not think it would make much of a difference.

It is also not something we use for the search page which from a performance perspective is even more taxing.

src/apps/material/order-digital-copy.test.ts
```diff
@@ -134,6 +144,55 @@ describe("Material - Order digital copy", () => {
       .and("contain", "Close")
       .click();
   });
+
+  beforeEach(() => {
+    cy.interceptGraphql({
+      operationName: "getMaterial",
+      fixtureFilePath: "material/order-digital-copy/order-digital-fbi-api"
+    });
```
kasperg, line 152: Please reconsider this `beforeEach` implementation.

To me it looks like what is already defined on L4 in this file. If they are different then one has to go.

Perhaps this is a merge conflict gone wrong?

src/components/cover/coverPreloadLink.tsx
```diff
@@ -0,0 +1,42 @@
+import React from "react";
```
kasperg, line 1: If you agree that `CoverPreloadLink` is unnecessary then remember to remove this component.

kasperg (changes requested):
A lot of work went into this. I have a few comments for you to look into.

You do not mention memoization of the components but I expect this also had a considerable effect on the performance. 

On the other hand I suggest that we drop cover preloading.

src/apps/material/material.tsx
spaceo: Yeah. Think you are right. I don't think the cover preload link here is giving us a noticeable gain. I'll remove it again.

src/components/cover/coverPreloadLink.tsx
spaceo: yep

src/apps/material/order-digital-copy.test.ts
spaceo: Nice spotted!
I have corrected it.

sinejespersen (approved):
assuming @danskernesdigitalebibliotek/dpl-design-system will be reverted in package.json....

kasperg (approved):
:+1: LGTM

PR #341, opened 2023-02-02
Fix/release date

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFSOEG-404

#### Description
This PR corrects the way we determine release year for search result items. Doing that also required updating our "manifestation simple fields" fragment (graphQL).

#### Screenshot of the result
before:
image

after:
image

#### Checklist

n/a

### DIFF
--- a/src/components/search-result-list/search-result-list-item/search-result-list-item.tsx
+++ b/src/components/search-result-list/search-result-list-item/search-result-list-item.tsx
@@ -143,7 +143,7 @@ const SearchResultListItem: React.FC<SearchResultListItemProps> = ({
           <Link href={materialFullUrl}>{fullTitle}</Link>
         </h2>
 
-        {author && (
+        {author && item && (
           <p className="text-small-caption" data-cy="search-result-item-author">
             {`${t("byAuthorText")} ${author}`}
             {getReleaseYear(item) ? ` (${getReleaseYear(item)})` : ""}
--- a/src/components/search-result-list/search-result-list-item/search-result-list-item.tsx
+++ b/src/components/search-result-list/search-result-list-item/search-result-list-item.tsx
@@ -13,7 +13,8 @@ import {
   creatorsToString,
   filterCreators,
   flattenCreators,
-  getManifestationPid
+  getManifestationPid,
+  getReleaseYear
 } from "../../../core/utils/helpers/general";
 import SearchResultListItemCover from "./search-result-list-item-cover";
 import HorizontalTermLine from "../../horizontal-term-line/HorizontalTermLine";
@@ -37,13 +38,13 @@ export interface SearchResultListItemProps {
 }
 
 const SearchResultListItem: React.FC<SearchResultListItemProps> = ({
+  item,
   item: {
     titles: { full: fullTitle },
     series,
     creators,
     manifestations: { all: manifestations },
-    workId,
-    workYear
+    workId
   },
   coverTint,
   resultNumber
@@ -145,7 +146,7 @@ const SearchResultListItem: React.FC<SearchResultListItemProps> = ({
         {author && (
           <p className="text-small-caption" data-cy="search-result-item-author">
             {`${t("byAuthorText")} ${author}`}
-            {workYear && ` (${workYear.year})`}
+            {getReleaseYear(item) ? ` (${getReleaseYear(item)})` : ""}
           </p>
         )}
       </div>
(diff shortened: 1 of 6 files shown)


src/components/search-result-list/search-result-list-item/search-result-list-item.tsx
```diff
@@ -142,10 +143,10 @@ const SearchResultListItem: React.FC<SearchResultListItemProps> = ({
           <Link href={materialFullUrl}>{fullTitle}</Link>
         </h2>
 
-        {author && (
+        {author && item && (
           <p className="text-small-caption" data-cy="search-result-item-author">
             {`${t("byAuthorText")} ${author}`}
-            {workYear && ` (${workYear.year})`}
+            {getReleaseYear(item) ? ` (${getReleaseYear(item)})` : ""}
```
spaceo, line 149: Please consolidate these lines with
[this](https://github.com/danskernesdigitalebibliotek/dpl-react/pull/341/files#diff-a09d47a1eb335411b7f4c67799d5a4d59bac0b55e8dcf3d3142c99f86d309378R348), [this](https://github.com/reload/dpl-react/blob/3a23ad1240698df192979453ce2e2111268b0915/src/components/reservation/helper.ts#L128) and [this](https://github.com/reload/dpl-react/blob/3a23ad1240698df192979453ce2e2111268b0915/src/core/utils/helpers/general.ts#L18)

src/core/utils/helpers/general.ts
```diff
@@ -345,4 +345,28 @@ export const dataIsNotEmpty = (data: unknown[]) => Boolean(data.length);
...
+    case false:
+      return (
+        latest.edition?.publicationYear?.display ||
+        latest.workYear?.year ||
+        null
+      );
+    default:
+      return null;
+  }
+};
```
spaceo, line 370: I don't think it makes sense to use a switch case here.
`materialIsFiction` is always returning a `boolean` and the default case would never be executed.
Suggestion (har ikke testet:
```typescript
export const getReleaseYear = (work: Work) => {
  const existsOnWork = work.fictionNonfiction;
  const { latest, bestRepresentation } = work.manifestations;
  const manifestation = bestRepresentation || latest;

  if (existsOnWork && materialIsFiction(work)) {
    return work.workYear?.year || null;
  }

  if (materialIsFiction(manifestation)) {
    return (
      manifestation.dateFirstEdition?.display ||
      manifestation.dateFirstEdition?.year ||
      null
    );
  }

  return (
    latest.edition?.publicationYear?.display || latest.workYear?.year || null
  );
};

**Review by spaceo — CHANGES_REQUESTED** [review]
I think we need to consolidate places in the codebase that does the same.

File: src/core/utils/helpers/general.ts (hunk 2 above)
Comment (spaceo) [reply src/core/utils/helpers/general.ts:370]: Hov har faktisk en bedre version, hvor vi sikrer os at return er: `string | null`:
```typescript
export const getReleaseYear = (work: Work) => {
  const { latest, bestRepresentation } = work.manifestations;
  const manifestation = bestRepresentation || latest;

  // If the work tells us that it is fiction.
  if (materialIsFiction(work)) {
    return stringifyValue(work.workYear?.year) || null;
  }

  // If the manifestation tells us that it is fiction.
  if (materialIsFiction(manifestation)) {
    return (
      stringifyValue(
        manifestation.dateFirstEdition?.display ||
        manifestation.dateFirstEdition?.year
      ) || null
    );
  }

  // Otherwise try to
  // use the publication year of the latest manifestation edition.
  // or use the work year of the latest manifestation.
  // or surrender with a null.
  return (
    stringifyValue(
      latest.edition?.publicationYear?.display || latest.workYear?.year
    ) || null
  );
};
```
Men det kræver en ændring af `stringifyValue()`:
```typescript
export const stringifyValue = (value: number | string | null | undefined) =>
  value ? String(value) : "";

Adamik10: I like your other solution :) 
But without stringifying the value. It is possibly null/undefined, we don't want that rendered as a string.

src/components/search-result-list/search-result-list-item/search-result-list-item.tsx
Adamik10: I incorporated `getManifestationPublicationYear()` helper into my function and renamed it + commented on it so that people know it's a function specific to search result page, as there we are very particular what years we want to show for what works.

spaceo (approved):
Great.
Although I still think we for instance could have a functional component that we could use in various places for showing author/year, this is an improvement.
👍

jeppekroghitk (approved):

PR #345, opened 2023-02-07
Close modals by clicking outside

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DSC-18

#### Description
This PR updates the design system package which shrinks all full screen modals so that backdrop is always visible, and introduces functionality to close modals by clicking on the backdrop. 

#### Screenshot of the result
https://user-images.githubusercontent.com/28546954/217244910-1138879c-dcfc-4cc8-a005-cd7785ff04d9.mov

#### Checklist

The design system package still needs to be updated before this is merged - [this PR](https://github.com/danskernesdigitalebibliotek/dpl-design-system/pull/196) needs to get merged into the develop branch.

### DIFF
--- a/src/core/utils/modal.tsx
+++ b/src/core/utils/modal.tsx
@@ -61,7 +61,7 @@ function Modal({
   return (
     <FocusTrap>
       <div>
-        {/* The backdrop doesn't have a tab index or keyboard listener because it barely duplicates
+        {/* The backdrop doesn't have a role or keyboard listener because it barely duplicates
           the close button's functionality which possesses both. */}
         {/* eslint-disable jsx-a11y/click-events-have-key-events, jsx-a11y/no-static-element-interactions */}
         <div
--- a/src/core/utils/modal.tsx
+++ b/src/core/utils/modal.tsx
@@ -63,7 +63,7 @@ function Modal({
       <div>
         {/* The backdrop doesn't have a tab index or keyboard listener because it barely duplicates
           the close button's functionality which possesses both. */}
-        {/* eslint-disable jsx-a11y/click-events-have-key-events, jsx-a11y/interactive-supports-focus */}
+        {/* eslint-disable jsx-a11y/click-events-have-key-events, jsx-a11y/no-static-element-interactions */}
         <div
           className="modal-backdrop"
           style={{
@@ -73,14 +73,13 @@ function Modal({
             // the remaining modals
             zIndex: modalIds.indexOf(modalId) + 10
           }}
-          role="button"
           onClick={() => {
             close();
           }}
         >
           {" "}
         </div>
-        {/* eslint-enable jsx-a11y/click-events-have-key-events, jsx-a11y/interactive-supports-focus */}
+        {/* eslint-enable jsx-a11y/click-events-have-key-events, jsx-a11y/no-static-element-interactions */}
         <div
           className={clsx(
             "modal",
--- a/src/core/utils/modal.tsx
+++ b/src/core/utils/modal.tsx
@@ -76,8 +76,7 @@ function Modal({
           onClick={() => {
             close();
           }}
-        >
-        </div>
+        />
         {/* eslint-enable jsx-a11y/click-events-have-key-events, jsx-a11y/no-static-element-interactions */}
         <div
           className={clsx(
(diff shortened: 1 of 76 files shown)


src/core/utils/modal.tsx
```diff
@@ -52,19 +53,33 @@ function Modal({
...
+          style={{
+            // some elements are designed with z-index which means they pop up over the modal
+            // so I add 10 to the z-index of the modal
+            // the index of the modalid is used, so the newest modal is always on top of
+            // the remaining modals
+            zIndex: modalIds.indexOf(modalId) + 10
+          }}
+          role="button"
+          onClick={() => {
+            close();
```
kasperg, line 77: Consider the usage of the role button.

I appreciate the commentary regarding disabling the accessibility linting issues but I wonder if you got this 100% right with the `role="button"` for the backdrop when we at the same time want assistive technology to ignore the element entirely and defer to the actual close button.

I took a look at how [the Reach UI project (accessible React UI elements) handles dialogs](https://reach.tech/dialog). They do not seem to hav any accessibility-related props on their overlay but still register a click handler to close the dialog when clicking on it.

I expect removing the `button` role might affect which linting errors we get but then I suggest that we adjust these accordingly.

If we want to investigate this we could also try to use some assistive technology ourselves to see how it interprets the current situation.

kasperg (changes requested):
:+1: I think a small change is needed here and then some food for thought.

src/core/utils/modal.tsx
Adamik10: Yeah, it's a bit tricky. And you're totally right about the role button. I removed it from the backdrop, but then we just get different errors (as you mention you'd expect).

src/core/utils/modal.tsx
```diff
@@ -19,6 +19,7 @@ type ModalProps = {
   screenReaderModalDescriptionText: string;
   classNames?: string;
   dataCy?: string;
+  orientation?: "right";
```
kasperg, line 22: Please reconsider this structure.

Instead of making the prop an optional literal type with a single value I suggest using a non-optional literal with multiple values e.g. "center" | "right" and provide a default value of "center".

This way you also avoid the current detection where any `orientation` value will add the `modal-btn-close--offset` class. I would have expected that only the `right` value would need the `modal-btn-close--offset` class.

src/core/utils/modal.tsx
```diff
@@ -52,19 +54,32 @@ function Modal({
...
-          // so I add 10 to the z-index of the modal
-          // the index of the modalid is used, so the newest modal is always on top of
-          // the remaining modals
-          zIndex: modalIds.indexOf(modalId) + 10
-        }}
-      >
+      <div>
+        {/* The backdrop doesn't have a tab index or keyboard listener because it barely duplicates
+          the close button's functionality which possesses both. */}
+        {/* eslint-disable jsx-a11y/click-events-have-key-events, jsx-a11y/no-static-element-interactions */}
```
kasperg, line 66: Consider whether the comment still reflects the rules being disabled.

src/core/utils/modal.tsx
kasperg, line 22: Please reconsider this structure.

Instead of making the prop an optional literal type with a single value I suggest using a non-optional literal with multiple values e.g. "center" | "right" and provide a default value of "center".

This way you also avoid the current detection where any `orientation` value will add the `modal-btn-close--offset` class. I would have expected that only the `right` value would need the `modal-btn-close--offset` class.

Adamik10: I remade it a bit differently - to a boolean. We currently only have two types of modals - centered with offset and sliders that come from the right side. I overcomplicated it with the last implementation.

kasperg (approved):
:+1: LGTM

src/core/utils/modal.tsx
```diff
@@ -52,19 +54,32 @@ function Modal({
...
+            // so I add 10 to the z-index of the modal
+            // the index of the modalid is used, so the newest modal is always on top of
+            // the remaining modals
+            zIndex: modalIds.indexOf(modalId) + 10
+          }}
+          onClick={() => {
+            close();
+          }}
+        >
+          {" "}
```
sinejespersen, line 80:
```suggestion

```

sinejespersen (approved):

PR #346, opened 2023-02-07
Use redux to facets

kasperbirch1:
#### Link to issue

#### Description

#### Screenshot of the result

#### Checklist


src/apps/search-result/search-result.tsx
```diff
@@ -28,33 +27,34 @@ import FacetLine from "../../components/facet-line/FacetLine";
...
+import { addToFilter } from "../../core/filter.slice";
 
 interface SearchResultProps {
   q: string;
   pageSize: number;
 }
 
 const SearchResult: React.FC<SearchResultProps> = ({ q, pageSize }) => {
+  const dispatch = useDispatch();
+  const filterState = useSelector((state: RootState) => state.filter);
```
kasperg, line 40: This gets the filters directly from the store but there does not seem to be any checks that confirm that the stored filters actually match the query. This results in the following behavior:

1. User searches for "Harry Potter"
2. User filters on facet term "Magic"
3. User goes to material "Harry Potter and the Philosophers Stone"
4. User goes back to the search result for "Harry Potter"
5. The result will be filtered by the facet term "Magic"
6. The user goes back to the material "Harry Potter and the Philosophers Stone"
7. The user follows a link to the search result for "Fiction" e.g. by clicking on a term.
8. The search result will be filtered by the facet term "Magic"

While the filter in 5 is intentional the filter in 8 is not.

We need a way to distinguish the two and clear the stored filters if a new query is made.

kasperg (changes requested):

kasperbirch1:
Some notes:
* You could introduce the useFilterHandler again. Then you don't need to dispatch in all the places you want to update the filters and the dependency on redux is only defined in that hook, which makes it easier to swap in the future.
* You could investigate if you can use the generic where Kasper suggest a change. `...map<BlaBla>(`.
* I think it is better to maintain the const name "filters" indstead of "filterState". In my view it is not necesarry to indicate that we are using state in the variable name.

src/components/facet-line/FacetLineSelected.tsx
```diff
@@ -1,22 +1,20 @@
...
+  const filterState = useSelector((state: RootState) => state.filter);
 
-const FacetLineSelected: React.FunctionComponent<FacetLineSelectedProps> = ({
-  filters,
-  filterHandler
-}) => {
   return (
     <ul className="facet-line-selected-terms">
-      {Object.entries(filters).map(([facet, value]) => {
+      {Object.entries(filterState as Filter).map(([facet, value]) => {
```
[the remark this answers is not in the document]
kasperbirch1: See this https://github.com/danskernesdigitalebibliotek/dpl-react/pull/346/commits/70b554e4c1153006b82efc3a182f784d37511a4c

src/core/filter.slice.ts
```diff
@@ -0,0 +1,61 @@
...
+      }
+
+      // If there are more terms in the facet, return the updated state
+      return {
+        ...state,
+        [facet]: selectedFacet
+      };
+    },
+    clear() {
+      return {};
```
spaceo, line 55: I think conceptually it is prettier to return the initialState (which is the same result). But if you should choose to change the initialState in the future, clear() would return to that.

src/core/utils/helpers/url.ts
```diff
@@ -34,6 +34,12 @@ export const setQueryParametersInUrl = (parameters: {
   window.history.replaceState(null, "", processedUrl);
 };
 
+export const removeUrlQueryParam = (param: string) => {
+  const processedUrl = new URL(getCurrentLocation());
+  processedUrl.searchParams.delete(param);
+  window.history.replaceState(null, "", processedUrl);
+};
+
```
spaceo, line 42: The is already a `removeQueryParametersFromUrl()` in the same file. Please consolidate the two functions.

src/apps/search-result/search-result.tsx
```diff
@@ -164,22 +158,23 @@ const SearchResult: React.FC<SearchResultProps> = ({ q, pageSize }) => {
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [campaignData]);
 
+  // This is a hack to clear filters when the user navigates
+  // to the search page from links where there should be no filters.
+  useEffect(() => {
+    const filtersUrlParam = getUrlQueryParam("filters");
+    if (filtersUrlParam !== "withFilters") clearFilter();
```
spaceo, line 165: Isn't it possible to maintain the filters so if the user goes back in history they are still intact the moment "withFilters" is in the URL?
With that model it is a matter about to ignore/use the stored filters depending on the "withFilters" param.
Does it makes sense?

spaceo: If we stick to that model. This `useEffect` would not be needed.

src/apps/search-result/useFilterHandler.tsx
```diff
@@ -1,47 +1,43 @@
...
-      } else {
-        setFilters({
-          ...filters,
-          [facet]: { [term.term]: term }
-        });
-      }
-    }
+  const addToFilter = useCallback(
+    (payload: FilterPayloadType) => {
+      setQueryParametersInUrl({
```
spaceo, line 27: Think it would be less processing if we add a check wether the "withFilters" is already there

src/apps/search-result/search-result.tsx
spaceo: Actually I think it could be solved with just returning empty filters from the filterHandler if "withFilters" is not in the URL

spaceo (changes requested):
Looks great.
Nice to see that you are getting to know redux and that you managed to eliminate som ts assertions.

jeppekroghitk (approved):

src/apps/search-result/search-result.tsx
```diff
@@ -168,22 +158,23 @@ const SearchResult: React.FC<SearchResultProps> = ({ q, pageSize }) => {
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [campaignData]);
 
+  // This is a hack to clear filters when the user navigates
+  // to the search page from links where there should be no filters.
+  useEffect(() => {
+    const filtersUrlParam = getUrlQueryParam("filters");
+    if (filtersUrlParam !== "usePersistedFilters") clearFilter();
+  }, [clearFilter]);
```
kasperg, line 166: I had to think a bit about why this is necessary.

Letting the user navigate to a new search result should not necessarily clear the filters for any other search result. However we only have a single place to store the selected filters so if the user adds more filters on the new search result it may result in filters getting mixed up between searches.

Consequently we have to clear them when starting a new search. This will however result in the following situation:

1. The user does a search
2. The user applies a filter
3. The user goes to a material
4. The user clicks on a term in a material
5. The user is shown a search result
6. Filters are cleared
7. The user clicks the back button twice and returns to the search result in 1
8. Filters from 2 are not applied as they have been cleared in 6

😵

src/apps/search-header/search-header.tsx
```diff
@@ -194,6 +196,7 @@ const SearchHeader: React.FC = () => {
   }
 
   function handleInputValueChange(changes: UseComboboxStateChange<Suggestion>) {
+    clearFilter();
```
kasperg, line 199: Please consider if we need to clear the filter at all given the addition to the search result in L163.

As far as I can tell this will clear the filter if no `usePersistedFilters` query parameter is set and no query parameter will be set by default.

This also means we keep the filter handler out of the search header, which I think is nice for separation of concerns.

kasperg (changes requested):
:+1: but I still think this needs a bit more work before it is ready.

kasperg (approved):
:+1: LGTM

PR #347, opened 2023-02-07
Fix useGetFacets re-render bug

spaceo:
#### Description

The search result and  facetbrowsermodal component is rerendering infinetely. This fixes the bug.

#### Checklist


kasperg (approved):
:+1: but I wonder why the implementation caused the rerenders.

spaceo:
> 👍 but I wonder why the implementation caused the rerenders.

The setState and the useEffect was not needed so we skipped it.
I didn't double check but my guess is that the useEffect hook kept being called with setState and the caused the infinite loop of renders.

jeppekroghitk (approved):

PR #348, opened 2023-02-08
Change reviews based on API changes

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFSOEG-417

#### Description
Based on the new changes in the API the graphQL fragments, data structure for fetching works, and react components have been adjusted. One bigger change is that we now call for review data only if the disclosure is opened on material page, as opposed to getting reviews as a part of the work data on pageload. 

#### Screenshot of the result
image

https://user-images.githubusercontent.com/28546954/217556086-3fcb1fa0-4647-4171-980a-338f040fd1ce.mov

#### Checklist

n/a


src/apps/material/material.test.ts
```diff
@@ -174,6 +174,24 @@ describe("Material", () => {
...
+    });
+    cy.interceptGraphql({
+      operationName: "getReviewManifestations",
+      fixtureFilePath: "material/reviews.json"
+    });
+    cy.visit("/iframe.html?id=apps-material--default&viewMode=story&type=bog");
+
+    cy.getBySel("material-reviews-disclosure").should("be.visible").click();
+    cy.getBySel("material-reviews").should(
+      "contain",
```
spaceo, line 190: NB: We probably need some scrolling here when the performance PR has been merged, because of lazy loading..

src/apps/material/material.tsx
```diff
@@ -207,11 +204,7 @@ const Material: React.FC<MaterialProps> = ({ wid }) => {
           cyData="material-reviews-disclosure"
         >
           <MaterialReviews
-            listOfReviews={
-              reviews as Array<
-                LibrariansReview | ExternalReview | InfomediaReview
-              >
-            }
+            pids={hasReview.map((review) => review.pid as Pid)}
```
spaceo, line 207: I think you should look into the `Work`type and see if you can add the `Pid`type there on the review. I think we should avoid type assertions as much as possible. Because with type assertions we risk the possibility of enforcing a wrong type to a value.

src/components/material/MaterialReviews.tsx
```diff
@@ -1,30 +1,63 @@
...
+    pid: pids
+  });
+
+  if (isLoading) {
+    return <ReviewSkeleton />;
+  }
+
+  if (!data) {
+    return null;
+  }
```
spaceo, line 28: Have you tried throttling the network? Returning a null here could potentially lead to a glitch in the rendering.

src/components/material/MaterialReviews.tsx
```diff
@@ -1,30 +1,63 @@
...
+
+  const { manifestations: reviews } = data;
+
   return (
-    <ul className="reviews">
-      {listOfReviews.map((review) => {
-        if (review.__typename === "ExternalReview") {
-          return <ReviewExternal review={review} />;
+    <ul className="reviews" data-cy={dataCy}>
+      {reviews.map((review) => {
```
spaceo, line 34: Suggestion:
```typescript
{reviews.map(({access}) => {

File: src/components/material/MaterialReviews.tsx [hunk 5]
```diff
@@ -1,30 +1,63 @@
...
+        if (
```
spaceo, line 35: I think it would be pettier with a function returning the review type with the same precedense as you have here. Then you could use a switch case outside the render function to determine which component to return.

src/components/material/ReviewSkeleton.tsx
```diff
@@ -0,0 +1,14 @@
+import * as React from "react";
```
spaceo, line 1: Yay skeleton!! 💀

src/components/material/ReviewExternal.tsx
```diff
@@ -1,26 +1,44 @@
...
+        accessUrls.map(({ url, origin }, index) => {
           return (
-            <a href={url} className="link-tag text-small-caption mb-8">
-              {origin}
-            </a>
+            <>
+              <span>{index > 0 ? ", " : ""}</span>
+              <a href={url} className="link-tag text-small-caption mb-8">
+                {origin}
+              </a>
```
spaceo, line 40: Can we use the Link component here?

spaceo (changes requested):
Looks good! But please look at my comments ;)

src/apps/material/material.test.ts
Adamik10: Thanks for heads up! :)

src/components/material/MaterialReviews.tsx
Adamik10: Yes I have tried Slow 3G :) all works fine. 
It's because if it's still loading we just show the skeleton.

src/components/material/MaterialReviews.tsx
Adamik10: It's not possible to write it like that because the access prop is possibly undefined. Typescript doesn't like that.

src/components/material/MaterialReviews.tsx
Adamik10: Mmmm I must say I don't think it's prettier, but I did it, and it's commited.

spaceo (approved):
LGTM 👍

jeppekroghitk (approved):

PR #352, opened 2023-02-10
Add getDbcVerifiedSubjectsFirst

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DSC-38

#### This PR adds a new function getDbcVerifiedSubjectsFirst
which takes an array of subjects as an argument and returns a new array with unique subjects where dbcVerified subjects appear first.

#### Checklist


src/apps/material/helper.ts
```diff
@@ -350,3 +350,12 @@ export const getBestMaterialTypeForWork = (work: Work) => {
 export const reservationModalId = (faustIds: FaustId[]) => {
   return constructModalId("reservation-modal", faustIds.sort());
 };
+
+export const getDbcVerifiedSubjectsFirst = (subjects: Work["subjects"]) =>
+  Array.from(
+    new Set([
+      // dbcVerified needs to be first, because it is the most accurate
```
kasperg, line 357: :+1: Nice commenting.

src/apps/material/helper.ts
```diff
@@ -350,3 +350,12 @@ export const getBestMaterialTypeForWork = (work: Work) => {
 export const reservationModalId = (faustIds: FaustId[]) => {
   return constructModalId("reservation-modal", faustIds.sort());
 };
+
+export const getDbcVerifiedSubjectsFirst = (subjects: Work["subjects"]) =>
+  Array.from(
+    new Set([
```
kasperg, line 356: Consider using `_.uniq` instead of combining `Array.form` + `Set`. We have Lodash in our toolchain anyway and `_.uniq` makes your intention more readable.

Also consider adding a comment explaining why we need remove duplicates.

kasperg (approved):
:+1: One thing for you to consider before passing this along.

kasperg (changes requested):
Cypress tests fail here. I do not think it is related to other failures.

jeppekroghitk (approved):

kasperg (approved):
:+1: LGTM

PR #355, opened 2023-02-15
Add mutation for ordering digital copy and update test + Refactor 'interceptGraphql'

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DSC-32

#### This pull request adds a new mutation for ordering a digital copy of an article, replacing the usage of the old 'useDplDasDigitalArticleOrderPOS' with 'usePlaceCopyMutation'. 

The new mutation returns a status, which is used by the 'getResponseMessage' function to determine the appropriate translation string to display.

'getResponseMessage' takes two parameters, articleResponse and t. The function checks if articleResponse is defined, and if so, it creates an object with the keys being the different statuses of 'CopyRequestStatus' and the values being strings returned from the useTextFunction. Finally, it returns the value of the key in the object that matches the status of articleResponse.

Additionally, this PR updates the test for ordering a digital copy to handle all known status responses, including:

- OK
- ERROR_UNAUTHENTICATED_USER
- ERROR_AGENCY_NOT_SUBSCRIBED
- ERROR_PID_NOT_RESERVABLE
- ERROR_MISSING_CLIENT_CONFIGURATION.

The 'InterceptGraphqlParams' type now includes the 'statusCode' field, and the 'fixtureFilePath' field is made optional. The 'interceptGraphql' command has been modified to use these options and set the 'statusCode' default to a 200 status code if no 'statusCode' is specified. These changes make it easier to test GraphQL requests in Cypress by allowing developers to specify the HTTP status code and whether to use a fixture file.

#### Checklist

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -2,3 +2,4 @@ src/core/cover-service-api/model/
 src/core/cover-service-api/cover-service.ts
 src/core/dpl-cms/model
 src/core/dpl-cms/dpl-cms.ts
+src/core/fbs/fbs.ts
--- a/.storybook/preview.js
+++ b/.storybook/preview.js
@@ -57,8 +57,9 @@ const WrappedStory = (app) =>
   });
 
 const App = ({ story }) => <Store>{WrappedStory(story)}</Store>;
-// TODO: Using addon-redux would be much nicer, but it doesn't seem to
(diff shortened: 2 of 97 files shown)


kasperg (approved):
:+1: LGTM

PR #358, opened 2023-02-17
Usetext vitest unit test

spaceo:
Had too many conflicts (because a lot happened in develop since the changes were made) so I cherry picked changes into a clone of develop.
You can see history/conversations [here](https://github.com/danskernesdigitalebibliotek/dpl-react/pull/222).

---
#### Description

The code base is growing and so does the number of functions and custom hooks.
I missed unit testing while developing the extended text functionality and expansion of the useText hook.

So this is a suggestion for introducing unit tests by using [Vitest](https://vitest.dev/) and [testing-library/react-hooks](https://github.com/testing-library/react-hooks-testing-library) for hook testing.

I started out with Jest but had problems with having both Jest and Cypress in the same codebase. They have colliding test function names and Typescript could not figure it out.

I got Vitest recommended and it seems very fast and just as capable as Jest.

#### Screenshot of the result

image

#### Checklist

This is first iteration of bringing unit tests into our setup.
Next step is to bring it into our CI workflow.


sinejespersen (approved):

architecture/adr-006-unit-tests.md
```diff
@@ -0,0 +1,38 @@
...
+While we have a good coverage in our UI tests from Cypress we are lacking
+something to tests the inner workings of the applications.
+
+With unit tests added we can test bits of functionality that is shared
+between different areas of the application and make sure that we get the
+expected output giving different variations of input.
+
+## Decision
+
+We decided to go for Vitest which is an easy to use and very fast
```
kasperg, line 16:
```suggestion
We decided to go for [Vitest](https://vitest.dev/) which is an easy to use and very fast
```

architecture/adr-006-unit-tests.md
```diff
@@ -0,0 +1,38 @@
...
+unit testing tool.
+
+It has more or less the same capabilities as Jest
+which is another popular testing framework which is similar.
+
+Vitest is framework agnostic so in order to make it possible to test hooks
+we found `@testing-library/react-hooks` that works in conjunction with Vitest.
```
kasperg, line 23:
```suggestion
we found [`@testing-library/react-hooks`](https://react-hooks-testing-library.com/) that works in conjunction with Vitest.
```

architecture/adr-006-unit-tests.md
```diff
@@ -0,0 +1,38 @@
...
+With unit tests added we can test bits of functionality that is shared
+between different areas of the application and make sure that we get the
+expected output giving different variations of input.
+
+## Decision
+
+We decided to go for Vitest which is an easy to use and very fast
+unit testing tool.
+
+It has more or less the same capabilities as Jest
```
kasperg, line 19:
```suggestion
It has more or less the same capabilities as [Jest](https://jestjs.io/)
```

cypress.json
```diff
@@ -2,9 +2,12 @@
   "baseUrl": "http://localhost:57021",
   "integrationFolder": "./src",
   "testFiles": "**/*.test.*",
+  "ignoreTestFiles": [
+    "**/tests/unit/*.test.*"
+  ],
```
kasperg, line 7: Consider changing our `testFiles` to `./src/**/*.test.*` instead of using `ignoreTestFiles`.

[`ignoreTestFiles` seems to be legacy configuration](https://docs.cypress.io/guides/references/legacy-configuration#Folders--Files). It works because we are currently using an older version of Cypress but not when we upgrade.

package.json
```diff
@@ -34,6 +34,7 @@
     "start:storybook:test": "NODE_ENV=test start-storybook --port 57021 --quiet --ci",
     "start:storybook:dev": "NODE_ENV=development  start-storybook --port 80",
     "dev": "concurrently --raw \"yarn start:storybook:dev\" \"yarn watch\"",
+    "test:unit": "vitest",
     "test": "cypress run",
```
kasperg, line 38: Consider changing the script for running Cypress to `test:integration`.

This would match the pattern that we use for other types of task. We might even then create a "top-level" `test` script which runs both types of tests but I am not sure whether it would be used.

src/tests/unit/text.test.tsx
```diff
@@ -0,0 +1,78 @@
...
+});
+
+const Wrapper = ({ children }: { children: ReactNode }) => (
+  <Provider store={store}> {children} </Provider>
+);
+
+test("Should strings with placeholders", () => {
+  const { result } = renderHook(() => useText(), { wrapper: Wrapper });
+  // We name it t, because that is how we normally use it in the code.
+  const t = result.current;
```
kasperg, line 31: :+1: Nice commenting. Normally single letter variable names would be discouraged.

src/tests/unit/text.test.tsx
```diff
@@ -0,0 +1,78 @@
...
+      }
+    }
+  }
+});
+
+const Wrapper = ({ children }: { children: ReactNode }) => (
+  <Provider store={store}> {children} </Provider>
+);
+
+test("Should strings with placeholders", () => {
```
kasperg, line 28:
```suggestion
test("Should handle strings with placeholders", () => {
```

The wording of the test description does not seem right to me.

docs/code_guidelines.md
```diff
@@ -294,3 +294,19 @@ cy.getBySel("reservation-success-title-text")
...
+By using that we can test functions (and therefore also hooks) and classes.
+
+### Where do I place my tests?
+
+They have to be placed in src/tests/unit.
+
+### Testing hooks
+
+For testing hooks we are using the library: `@testing-library/react-hooks`
+you can read more about it [here](https://vitest.dev) and you can also take a
```
kasperg, line 310:
```suggestion
For testing hooks we are using the library [`@testing-library/react-hooks`](https://react-hooks-testing-library.com/) and you can also take a
```

I do not understand how the link to Vitest fits into the current sentence.

docs/code_guidelines.md
```diff
@@ -294,3 +294,19 @@ cy.getBySel("reservation-success-title-text")
...
 ```
+
+## Writing Unit Tests
+
+We are using Vitest as framework for running unit tests.
+By using that we can test functions (and therefore also hooks) and classes.
+
+### Where do I place my tests?
+
+They have to be placed in src/tests/unit.
```
kasperg, line 305:
```suggestion
They have to be placed in `src/tests/unit`.
```

docs/code_guidelines.md
```diff
@@ -294,3 +294,19 @@ cy.getBySel("reservation-success-title-text")
   // Another assertion.
   .and("contain", "Material is available and reserved for you!");
 ```
+
+## Writing Unit Tests
+
+We are using Vitest as framework for running unit tests.
```
kasperg, line 300:
```suggestion
We are using [Vitest](https://vitest.dev/) as framework for running unit tests.
```

vitest.config.ts
```diff
@@ -0,0 +1,14 @@
...
+
+export default defineConfig({
+  test: {
+    environment: "happy-dom",
+    include: ["./src/tests/unit/*.test.(ts|tsx)"],
+    coverage: {
+      all: true,
+      provider: "istanbul",
+      reporter: ["clover"],
+      include: ["src/**/*.{ts,tsx}"]
```
kasperg, line 11: We probably want to exclude (some of) the code in `src/core` from code coverage.

vitest.config.ts
```diff
@@ -0,0 +1,14 @@
+import { defineConfig } from "vitest/config";
+
+export default defineConfig({
+  test: {
+    environment: "happy-dom",
+    include: ["./src/tests/unit/*.test.(ts|tsx)"],
```
kasperg, line 6: Consider if we want to use the `tsx` suffix for test files.

I am not really sure whether they can/should be considered `tsx` even though we generally seem to use the `ts` and `tsx` suffixes without much consistency and it does not seem to make much difference functionally either.

docs/code_guidelines.md
kasperg: I wonder if we could place test files along with the code that they test like we do for Cypress tests.

I guess there is a possibility for mixing up the two frameworks if all test files use the format `[scope].test.ts`. If we want to separate the two then one approach might be to use `vite` for `src/core` and Cypress for `src/apps` and `src/components`. That would also clearly communicate the responsibility of the different tools and folders. It might be too strict though.

kasperg (commented):
:+1: Good stuff here. There are a lot of comments but most of them are thoughts, quibbles and markdown suggestions.

One major thing: The code coverage drops 10% when introducing this. I am not sure why but it does not seem to be intentional.

I see to reasons why:

1. We are now intentionally analyzing more code than before so the current number is actually the right one.
2. Something is wrong about our updated setup.

Please look into what is the case here.

package.json
spaceo: Makes sense. I'll change the command.

src/tests/unit/text.test.tsx
spaceo: 👍

vitest.config.ts
spaceo: The convention is to use tsx when jsx syntax is present in the file.
We might need to test components in the future, so I think we should keep tsx for now.

cypress.json
spaceo: I changed the globbing pattern :)

kasperg (approved):
:+1: LGTM

PR #362, opened 2023-02-20
Fix/paralel reservations

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DSC-40

#### Description
This PR fixes an bug introduced by unifying material availability buttons per type. A parallelle reservation on all manifestations of selected type should be performed for fictional works, otherwise we select the latest reservation and reserve that one. 

#### Screenshot of the result
image

#### Checklist

ITK is looking into the failing reservation-details test.


src/components/reservation/ReservationModalBody.tsx
```diff
@@ -141,7 +144,14 @@ export const ReservationModalBody = ({
...
     reservationResponse?.reservationResults[0]?.reservationDetails;
-  const manifestation = selectedManifestations[0];
+  const manifestation =
+    manifestationsToReserve?.[0] || selectedManifestations[0];
+  const getEditionText = () => {
+    if (!materialIsFiction(work) || manifestationsToReserve?.length === 1) {
+      return manifestation.edition?.summary;
+    }
+    return t("firstAvailableEditionText");
+  };
```
kasperg, line 154: Is there a reason why we want to define a function at this level? It does not seem to be a pattern we use much elsewhere. I mean: Either functions are placed in `helper.ts` or the variable is defined directly as `const editionText`.

src/components/reservation/helper.ts
```diff
@@ -149,4 +150,21 @@ export const getAuthorLine = (
...
+) => {
+  if (isPeriodical) {
+    return reservableManifestations;
+  }
+  if (!reservableManifestations || reservableManifestations.length < 1) {
+    return null;
+  }
+  // If the work isn't fictional we only want to reserve the latest manifestation.
+  if (!materialIsFiction(reservableManifestations[0])) {
+    return [getLatestManifestation(reservableManifestations)];
```
kasperg, line 165: Another approach to this would be `orderManifestationsByYear(reservableManifestations, "desc").slice(0,1)`. This would also handle a empty `reservableManifestations` quite elegantly. It is not as urgent now as we explicitly check on L160.

kasperg, line 165: The comment writes out what the code does in text. That is nice if the code is complex but more importantly is always *why* - especially if it refers to business logic.

src/components/reservation/helper.ts
```diff
@@ -149,4 +150,21 @@ export const getAuthorLine = (
     : [t("materialHeaderAuthorByText"), author, year].join(" ");
 };
 
+export const getManifestationsToReserve = (
+  reservableManifestations: Manifestation[] | null,
+  isPeriodical?: boolean
+) => {
+  if (isPeriodical) {
+    return reservableManifestations;
```
kasperg, line 158: Consider adding a comment explaining *why* we want to do parallel reservations for periodicals. When I think of it that does not make much sense to me. If the user wants Time Magazine February 2023, then doing a parallel reservation for another version seems odd.

src/components/material/material-buttons/material-buttons.test.ts
```diff
@@ -7,33 +7,25 @@ describe("Material buttons", () => {
       url: "**/availability/v3?recordid=**",
       fixtureFilePath: "material/unavailability.json"
     });
-    cy.visit("/iframe.html?id=apps-material--default&viewMode=story&type=bog")
-      .getBySel("material-description")
-      .scrollIntoView();
-
+    cy.visit("/iframe.html?id=apps-material--default&viewMode=story&type=bog");
+    cy.scrollTo("bottom");
```
kasperg, line 11: Scrolling across the entire page seems a bit odd to me compared to scrolling to the element which is responsible for the element. Perhaps it is necessary 🤔 .

kasperg (approved):
:+1: There is a lot of business logic embedded in this PR which could use some commentary. Consider this before passing this along.

src/components/reservation/helper.ts
```diff
@@ -149,4 +150,21 @@ export const getAuthorLine = (
     : [t("materialHeaderAuthorByText"), author, year].join(" ");
 };
 
+export const getManifestationsToReserve = (
+  reservableManifestations: Manifestation[] | null,
```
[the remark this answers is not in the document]
Adamik10: You are absolutely correct about the wonky return type of `UseReservableManifestations`
Will do :)

src/components/reservation/helper.ts
Adamik10: I think it's more readable as it is currently.

src/components/material/material-buttons/material-buttons.test.ts
Adamik10: Yeah, unfortunately it is necessary, because the elements don't exist before they enter viewport. The test was failing for `.scrollIntoView()` 😿

jeppekroghitk (approved):

PR #365, opened 2023-02-24
swapped exclamation mark in userisanonymous check + cypress testing

jeppekroghitk:
#### Link to issue

N/A (fix for existing: https://platform.dandigbib.org/issues/5377)

#### Description

This PR simply swaps the exclamation on the two checks for userIsAnonymous, in the main menu rendering. An earlier oversight. Furthermore fixed some issues in cypress tests, and added a TODO for checking unauthorized menu access.

#### Screenshot of the result

N/A

#### Checklist


spaceo (approved):
LGTM 👍 
But this is so much a fit for a simple Cypress test ;)

spaceo (changes requested):
There is a failing test...

cableman (approved):

src/apps/menu/menu.test.ts
```diff
@@ -1,9 +1,10 @@
-import { TOKEN_LIBRARY_KEY } from "../../core/token";
+import { TOKEN_LIBRARY_KEY, TOKEN_USER_KEY } from "../../core/token";
 
-describe("Menu", () => {
+describe("Menu (authenticated))", () => {
   beforeEach(() => {
     cy.window().then((win) => {
       win.sessionStorage.setItem(TOKEN_LIBRARY_KEY, "random-token");
+      win.sessionStorage.setItem(TOKEN_USER_KEY, "random-user");
```
spaceo, line 7: I think you should use `cy.createFakeAuthenticatedSession()` instead.
The reason why it was made was that then we don't need to know what is happening in order to create a fake user session. It might even change or be extended in the future, and then we only have to change it in that function.

spaceo (approved):
Approved. But please consider what I write in the comment ;)

PR #366, opened 2023-02-24
Scroll down before checking user blocked button in material buttons test

spaceo:
It isn't rendered before it enters the viewport.

#### Checklist


jeppekroghitk (approved):

kasperg (approved):
:+1: LGTM

PR #369, opened 2023-02-27
Cypress coverage fix

spaceo:
#### Description

This PR solves the missing Cypress coverage problem by being specific about which Cypress files to include.
By doing that  we also automatically exclude Vitest files.
Second commit is due to a ts warning about the construction of the vitest config.

#### Checklist

### DIFF
--- a/cypress.json
+++ b/cypress.json
@@ -1,7 +1,7 @@
 {
   "baseUrl": "http://localhost:57021",
   "integrationFolder": "./src",
-  "testFiles": "!(tests)/**/*.test.@(ts|tsx)",
+  "testFiles": "@(apps|components)/**/*.test.@(ts|tsx)",
   "defaultCommandTimeout": 10000,
   "retries": {
     "runMode": 3,
--- a/vitest.config.ts
+++ b/vitest.config.ts
@@ -12,3 +12,5 @@ export default defineConfig({
     }
   }
 });
+
+export {};


cypress.json
```diff
@@ -1,10 +1,10 @@
 {
   "baseUrl": "http://localhost:57021",
   "integrationFolder": "./src",
-  "testFiles": "!(tests)/**/*.test.@(ts|tsx)",
+  "testFiles": "@(apps|components)/**/*.test.@(ts|tsx)",
```
kasperg, line 4: This reraises my question from last week: 

> I wonder if we could place test files along with the code that they test like we do for Cypress tests?

> I guess there is a possibility for mixing up the two frameworks if all test files use the format `[scope].test.ts`. If we want to separate the two then one approach might be to use vite for `src/core` and Cypress for `src/apps` and `src/components`. That would also clearly communicate the responsibility of the different tools and folders. It might be too strict though.

With this change we seem to be halfway there anyway.

kasperg (approved):
:+1: I see that coverage is back. Nice.

I wonder why the change to the test files is necessary. As far as I can tell all [tests are run even without this change](https://github.com/danskernesdigitalebibliotek/dpl-react/actions/runs/4284443392/jobs/7461406195#step:5:1370). Oh well.

I have one suggestion for you to consider.

jeppekroghitk (approved):

PR #371, opened 2023-03-01
Fix/material buttons breaking material page

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFSOEG-470

#### Description
This PR fixes the broken material page where we need to check whether the user is blocked in order to show the correct material page buttons. However, if the user isn't logged in the query fails which our brand new error boundary considers enough a reason to break the whole page. 
The PR also fixes a small flaw on the search result page not showing release dates properly.

#### Screenshot of the result
n/a

#### Checklist

n/a


orval.config.ts
```diff
@@ -114,6 +121,8 @@ export default defineConfig({
           useQuery: true
         },
         operations: {
+          // The reason why we add this here is to be able to use "enabled" option in the
+          // proxy-url:GET query. This lets us call it conditionally.
```
spaceo, line 125: 👍

spaceo (approved):
LGTM 👍

jeppekroghitk (approved):

PR #373, opened 2023-03-02
Handle 0-hit search result

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFSOEG-468

#### Description
This PR introduces functionality to handle 0-hit search in the search result UI.

#### Screenshot of the result
Before (never ending loading screen):
image

After:
image

#### Checklist

n/a

### DIFF
--- a/src/components/facet-line/FacetLine.tsx
+++ b/src/components/facet-line/FacetLine.tsx
@@ -9,17 +9,11 @@ import useFilterHandler from "../../apps/search-result/useFilterHandler";
 
 type FacetLineProps = {
   q: string;
-  hitCount: number;
 };
 
-const FacetLine: React.FunctionComponent<FacetLineProps> = ({
-  q,
-  hitCount
-}) => {
+const FacetLine: React.FunctionComponent<FacetLineProps> = ({ q }) => {
   const { filters } = useFilterHandler();
-
   const cleanBranches = useGetCleanBranches();
-
   const { data, isLoading } = useIntelligentFacetsQuery({
     q: { all: q },
     facetsLimit: 6,
@@ -30,9 +24,7 @@ const FacetLine: React.FunctionComponent<FacetLineProps> = ({
   return (
     <>
       {isLoading && <FacetLineFiltersSkeleton />}
-      {data && hitCount > 0 && (
-        <FacetLineFilters facets={data.search.intelligentFacets} />
-      )}
+      {data && <FacetLineFilters facets={data.search.intelligentFacets} />}
       <FacetLineSelected />
     </>
   );
--- a/src/components/facet-line/FacetLine.tsx
+++ b/src/components/facet-line/FacetLine.tsx
@@ -10,19 +10,17 @@ import useFilterHandler from "../../apps/search-result/useFilterHandler";
 type FacetLineProps = {
   q: string;
   hitCount: number;
-  isLoading: boolean;
 };
 
 const FacetLine: React.FunctionComponent<FacetLineProps> = ({
   q,
-  hitCount,
-  isLoading
+  hitCount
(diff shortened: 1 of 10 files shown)


src/components/facet-line/FacetLine.tsx
```diff
@@ -25,8 +31,10 @@ const FacetLine: React.FunctionComponent<FacetLineProps> = ({ q }) => {
 
   return (
     <>
-      {!data && <FacetLineFiltersSkeleton />}
-      {data && <FacetLineFilters facets={data.search.intelligentFacets} />}
+      {(isLoading || isLoadingFacets) && <FacetLineFiltersSkeleton />}
+      {data && hitCount > 0 && (
+        <FacetLineFilters facets={data.search.intelligentFacets} />
+      )}
```
kasperg, line 37: Please consider if passing data about the number of hits is really necessary for the facet line to render the facet line.

As I see it it passes information about the state of the search result (`hitcount` and `isLoading`) which should not really be necessary for rendering the facet line. I think the situation might be caused by the `useIntelligentFacetsQuery()` always returning the same result. That should no longer be the case now. Thus I would think that if the search has 0 results then there will be no facets to render in the line either causing it to be empty.

If that holds true then we should be able to reduce this to using a `isLoadingFacets` variable to determine whether we should render a skeleton or the result.

Instead of passing `hitcount` and `isLoading` down we should be able to use these to control whether subcomponents are rendered or not in the `SearchResult` component.

src/components/search-result-list/SearchResultList.tsx
```diff
@@ -1,38 +1,44 @@
...
-            <SearchResultListItem
-              item={item}
-              coverTint={getCoverTint(i)}
-              resultNumber={i + 1}
-            />
-          </li>
-        ))}
+      {/* We fill out empty page without results a little bit to visually convey
+      to the user that the results are empty. */}
+      {(!hitcount || isDataEmpty(resultItems)) && <div className="my-112" />}
```
kasperg, line 32: Please reconsider adding this spacer here.

This mixes in considerations about spacing for empty results (which is not really a list) into a  component which is responsible for rendering a list. It also means that we have to pass information about the state of the search result (`hitcount` and `isLoading`) into the component.

src/core/utils/helpers/general.ts
```diff
@@ -372,7 +372,7 @@ export const getAllFaustIds = (manifestations: Manifestation[]) => {
 export const getScrollClass = (modalIds: string[]) => {
   return modalIds.length > 0 ? "scroll-lock-background" : "";
 };
-export const dataIsNotEmpty = (data: unknown[]) => Boolean(data.length);
+export const isDataEmpty = (data: unknown[]) => Boolean(!data.length);
```
kasperg, line 375: :+1:

We could also consider replacing this with [`_.isEmpty()`](https://lodash.com/docs/4.17.15#isEmpty).

kasperg (changes requested):
I do not agree with the approach chosen here.

As I see it you introduce additional complexity within multiple components which could have been solved with a simple new component and a top level conditional within `search-result` in the lines of:

```
if (!isLoading && hitcount == 0) {
  return <SearchResultEmpty>
} else {
  return (
    <div className="search-result-page">  
   // ...
```

src/components/facet-line/FacetLine.tsx
Adamik10: You're completely right. I made it more difficult than it needed to be.

kasperg (changes requested):
We discussed this in-person. These are some of my notes - primarily for reference.

Adamik10:
Follow this PR in #390

PR #375, opened 2023-03-03
Refactor `constructMaterialUrl` to use 'new URL' object

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFSOEG-474

####  This PR fixes an issue with the `constructMaterialUrl` function that modifies the original URL object instead of creating a new one, potentially causing bugs in other parts of the code.

The code changes include:
- Creating a new URL object materialUrl using the original URL object.

These changes ensure that the constructMaterialUrl function behaves as intended and avoids unintended side effects.

#### Checklist

### DIFF
--- a/src/core/utils/helpers/url.ts
+++ b/src/core/utils/helpers/url.ts
@@ -82,7 +82,7 @@ export const constructMaterialUrl = (
   workId: WorkId,
   type?: string
 ) => {
-  const materialUrl = new URL(url.href);
+  const materialUrl = new URL(url);
 
   // Replace placeholders with values.
   materialUrl.pathname = processUrlPlaceholders(materialUrl.pathname, [
--- a/src/core/utils/helpers/url.ts
+++ b/src/core/utils/helpers/url.ts
@@ -82,9 +82,10 @@ export const constructMaterialUrl = (
   workId: WorkId,
   type?: string
 ) => {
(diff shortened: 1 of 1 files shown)


src/core/utils/helpers/url.ts
```diff
@@ -82,9 +82,10 @@ export const constructMaterialUrl = (
   workId: WorkId,
   type?: string
 ) => {
-  const materialUrl = url;
+  const materialUrl = new URL(url.href);
```
spaceo, line 85: It is a minor thing, and it is more or less the same. But I think you should clone the URL object like we have done in the other [url helper functions](https://github.com/danskernesdigitalebibliotek/dpl-react/blob/50cf62e4ff18d8f1bd771b0a1693f84bcd314532/src/core/utils/helpers/url.ts)

So instead of:
```typescript
const materialUrl = new URL(url.href);
```
You write: 
```typescript
const materialUrl = new URL(url);
```
The stringify method of `URL`makes sure that the string representation of `url` is sent to the new constructor.

spaceo (changes requested):
Looks good. Just a minor comment

spaceo (approved):
LGTM 👍

sinejespersen (approved):

PR #376, opened 2023-03-03
Fix release year on search result page

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFSOEG-404

#### Description
This PR fixes an issue with not choosing the correct release year to render on the search result page. 

#### Screenshot of the result
n/a

#### Checklist

n/a


kasperg (approved):
:+1: 

It would be nice to have some sort of test for this. We have been changing this around a couple of this. Perhaps consider adding an test case for it in our existing Cypress tests.

jeppekroghitk (approved):

PR #377, opened 2023-03-06
Add selected materialType to `constructMaterialUrl`

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFSOEG-463

#### This pull request adds a new function called `getFirstMaterialTypeFromFilters` 

its retrieves the first materialType from filters that are included in manifestations. it is then used to construct the materialFullUrl in the `SearchResultListItem`

#### Screenshot of the result

https://user-images.githubusercontent.com/49920322/223070074-346bc298-6bc1-4465-a6f9-53c92eb3aa0f.mov

#### Checklist

### DIFF
--- a/src/components/facet-browser/helper.ts
+++ b/src/components/facet-browser/helper.ts
@@ -1,12 +1,10 @@
-import { head, mapValues } from "lodash";
+import { mapValues } from "lodash";
 import {
   FacetField,
   useSearchFacetQuery
 } from "../../core/dbc-gateway/generated/graphql";
 import useGetCleanBranches from "../../core/utils/branches";
 import { Filter, FilterItemTerm } from "../../core/filter.slice";
-import { getMaterialTypes } from "../../core/utils/helpers/general";
-import { Manifestation } from "../../core/utils/types/entities";
 
 export const allFacetFields = [
   FacetField.MainLanguages,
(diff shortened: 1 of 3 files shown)


src/components/facet-browser/helper.ts
```diff
@@ -100,4 +102,15 @@ export function getAllFilterPathsAsString(filterObject: {
   return allFilterPathsAsString;
 }
 
+export const getFirstMaterialTypeFromFilters = (
+  filters: Filter,
+  manifestations: Manifestation[]
+) => {
+  const materialTypeFilter = Object.keys(filters.materialTypes || {}).sort()[0];
```
kasperg, line 109: Please look into if we can do better here.

Filters is currently just a map of strings and `materialTypes` is a value what we know should contain data. If this name changes our code will stille run but return no types. We get an exported enum `FacetField` with the value `materialTypes`. We should be able to use that.

Also: Grabbing the first entry with `[0]` should be replaced with `_.head()` from Lodash.

src/components/facet-browser/helper.ts
```diff
@@ -100,4 +102,15 @@ export function getAllFilterPathsAsString(filterObject: {
   return allFilterPathsAsString;
 }
 
+export const getFirstMaterialTypeFromFilters = (
```
kasperg, line 105: This helper is placed in `src/components/facet-browser` but is used in `src/components/search-result-list[...]`.

Please move the component to an appropriate helper.

kasperg (approved):
:+1: LGTM

jeppekroghitk (approved):

PR #380, opened 2023-03-07
Add `LinkButton` and refactor  `Link` component

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFSOEG-450

####  This PR make some changes to the Link component and added a new LinkButton component

Here are the details of my changes:

- Renamed and moved Link + LinkNoStyle to src/components/atoms/links/Link.tsx.

- Refactored Link component to improve maintainability and accessibility, including:
-- Set the isNewTab prop default to false.
-- Moved onClick and onKeyUp handlers to the anchor tag for better accessibility and HTML semantics.
-- Simplified the logic for handling the trackClick prop.
-- Added e.preventDefault() to the handleClick and handleKeyUp event handlers to prevent the default action of the anchor tag when the trackClick prop is provided.

- Added LinkButton component that provides a button-like UI for links with customizable styling, size, and variant. The LinkButton component is designed to work with the refactored ButtonIcon component, allowing for the inclusion of icons within the button.

- Fixed the falling 'Campaign' test by removing the unnecessary check for the "role" attribute with the value "button". This check is no longer needed because the Link element will always be represented by an "a" tag instead of a "span" tag with the "button" role.

#### Checklist


kasperg (approved):
:+1: lots of good stuff here. A few things needs changing but I trust you to fix these things yourself before passing this along.

jeppekroghitk (approved):

PR #381, opened 2023-03-08
Itkdev - dashboard clean up

jeppekroghitk:
#### Link to issue

n/a

#### Description

added missing texts
removed 0 from interface
added styling
remove hardcoded text
refactor WarningBar to accommodate two looks
move pager (because valid html)

#### Screenshot of the result

image

image

#### Checklist


spaceo (changes requested):
Is in draft mode - does it need a review?

sinejespersen:
> Is in draft mode - does it need a review?

Yes, but not yet. so draft is on purpose.

sinejespersen:
> Looks ok. I am just in doubt of the "my-_" classes. What are they and why could "mb-_" not be used?

it could. But my can also be used, is one of them preferable?

spaceo (approved):
LGTM

PR #382, opened 2023-03-08
Use 'h4' for review headline

kasperbirch1:
# See https://github.com/danskernesdigitalebibliotek/dpl-react/pull/387

#### Link to issue
https://reload.atlassian.net/browse/DDFSOEG-452

#### Checklist

### DIFF
--- a/src/components/material/ReviewInfomedia.tsx
+++ b/src/components/material/ReviewInfomedia.tsx
@@ -64,7 +64,7 @@ const ReviewInfomedia: React.FC<ReviewInfomediaProps> = ({
       <li className="review text-small-caption" data-cy={dataCy}>
         {(authors || date) && <ReviewMetadata author={authors} date={date} />}
         {review?.rating && <ReviewHearts amountOfHearts={review.rating} />}
-        <div className="review__headline mb-8">
+        <h4 className="review__headline mb-8">
           {infomedia.error === "BORROWER_NOT_LOGGED_IN" ? (
             <Button
               label={t("loginToSeeReviewText")}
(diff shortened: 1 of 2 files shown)


src/components/material/ReviewInfomedia.tsx
```diff
@@ -80,7 +80,7 @@ const ReviewInfomedia: React.FC<ReviewInfomediaProps> = ({
           ) : (
             t("cantViewReviewText")
           )}
-        </div>
+        </h4>
```
kasperg, line 83: Please reconsider using a `<h4>` here.

As I see it this is not an actual headline. It is a button to show that the user cannot see the review without logging in.

kasperg (changes requested):
I would like a bit more information before I agree that this can be merged

First of all: At least one place the content is not an actual headline.

Second: This uses the `<h4>` headline but I am unsure why.

When checking a material display I would expect that if a `<h4>` is to be used for heading for the individual reviews then the heading for the review list (the disclosure) would be `<h3>`. That is not the case. In fact I could not find any `<h3>` as a part of the material page. Perhaps you expect to rework the headings for the disclosures but I cannot tell from the current commit.

Thus as is I do not think the current PR is ready for merging.

There are a number of tickets regarding heading structure for materials. There are a couple of things that might make sense here:

- The heading structure could be defined in the design system. Perhas we can even update [the accessibility test](https://github.com/chanzuckerberg/axe-storybook-testing) to check for this also.
- Perhaps it makes sense to fix all the issues in combination to ensure we get it right. 

Perhaps it would

PR #383, opened 2023-03-09
Add 'aria-pressed' attribute to `AvailabilityLabel`

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFSOEG-449

#### Add 'aria-pressed' attribute to AvailabilityLabel

The aria-pressed attribute has been added to the AvailabilityLabel component in the availability-label.tsx file. The selected prop is used to determine the value of the aria-pressed attribute. Additionally, two Cypress tests have been added to ensure that the aria-pressed attribute is correctly set for both selected and unselected availability labels.

Changes Made:
- Added aria-pressed attribute to the AvailabilityLabel component
- Added two Cypress tests to ensure correct behavior of aria-pressed attribute

#### Checklist

### DIFF
--- a/src/apps/material/material.test.ts
+++ b/src/apps/material/material.test.ts
@@ -225,7 +225,7 @@ describe("Material", () => {
     );
   });
 
-  it("Check if selected availability label have aria-pressed=true", () => {
+  it("Has a selected availability label based on url parameter", () => {
     cy.interceptGraphql({
       operationName: "getMaterial",
       fixtureFilePath: "material/fbi-api.json"
@@ -242,7 +242,7 @@ describe("Material", () => {
       .should("have.attr", "aria-pressed", "true");
   });
 
-  it("Check if unselected availability label have aria-pressed=false", () => {
+  it("Does not have selected availability labels which does not match url parameter", () => {
     cy.interceptGraphql({
       operationName: "getMaterial",
       fixtureFilePath: "material/fbi-api.json"
--- a/src/apps/material/material.test.ts
+++ b/src/apps/material/material.test.ts
@@ -225,6 +225,40 @@ describe("Material", () => {
     );
   });
 
+  it("Check if selected availability label have aria-pressed=true", () => {
+    cy.interceptGraphql({
+      operationName: "getMaterial",
+      fixtureFilePath: "material/fbi-api.json"
+    });
+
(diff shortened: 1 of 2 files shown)


Adamik10 (approved):
Very nice. Great that you're testing it as well!

src/components/availability-label/availability-label.tsx
```diff
@@ -91,6 +91,7 @@ export const AvailabilityLabel: React.FC<AvailabilityLabelProps> = ({
       role="button"
       tabIndex={0}
       data-cy={dataCy}
+      aria-pressed={selected}
     >
       <div className={classes.triangle} />
       <img className={classes.check} src={CheckIcon} alt="check-icon" />
```
kasperg, line 97: Considering this bug report I think two more changes are needed:

1. The `alt` attribute for the `CheckIcon` should be removed. It is included in the screen reader output as shown by the bug report and it is currently not translatable. Any information which could be provided using the icon is better represented through `aria-pressed` so we might as well remove it entirely.
2. The original change and the change in 1. must be reflected in the design system which also have a story for selected availability labels.

src/apps/material/material.test.ts
```diff
@@ -225,6 +225,40 @@ describe("Material", () => {
     );
   });
 
+  it("Check if selected availability label have aria-pressed=true", () => {
```
kasperg, line 228:
```suggestion
  it("Has a selected availability label based on url parameter", () => {
```

src/apps/material/material.test.ts
```diff
@@ -225,6 +225,40 @@ describe("Material", () => {
...
+    cy.getBySel("material-description").scrollIntoView();
+
+    cy.getBySel("availability-label")
+      .find('[data-cy="availability-label-type"]')
+      .contains("bog")
+      .parent()
+      .should("have.attr", "aria-pressed", "true");
+  });
+
+  it("Check if unselected availability label have aria-pressed=false", () => {
```
kasperg, line 245:
```suggestion
  it("Does not have selected availability labels which does not match url parameter", () => {
```

kasperg (changes requested):
:+1: for adding the test.

Please reconsider the changes based on the bug report.

kasperbirch1:
This will be fixed in https://github.com/danskernesdigitalebibliotek/dpl-react/pull/385

PR #385, opened 2023-03-09
Improve Accessibility in `AvailabilityLabel`

kasperbirch1:
#### Link to issue

https://reload.atlassian.net/browse/DDFSOEG-437
https://reload.atlassian.net/browse/DDFSOEG-449

#### This PR is to improve accessibility in `AvailabilityLabel`

it replacing the container 'div with type="button"' with either an outside `LinkNoStyle` or `button`. This change ensures that the AvailabilityLabel component is semantically correct and prevents double tabbing when navigating with the keyboard on the search-result page.

#### Checklist

### DIFF
--- a/src/apps/material/material.test.ts
+++ b/src/apps/material/material.test.ts
@@ -225,6 +225,40 @@ describe("Material", () => {
     );
   });
 
+  it("Has a selected availability label based on url parameter", () => {
+    cy.interceptGraphql({
+      operationName: "getMaterial",
+      fixtureFilePath: "material/fbi-api.json"
+    });
+
+    cy.visit("/iframe.html?id=apps-material--default&viewMode=story&type=bog");
+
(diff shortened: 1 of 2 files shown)


kasperg (changes requested):
I think the core problem here has to do with the structure of the component.

The component can either render a link (using `<LinkNoStyle>` with a button (using `<div role="button">` but without button functionality) within or just a button (using `<div role="button">`).

The current approach tries to fix this by removing more of the default button behavior with `tabIndex`.

Instead I suggest that you toggle between two different wrappers around the core availability label content (`<div className={classes.triangle} />check-icon[etc]`)

1. The button. You might even change from using a `<div role="button">` to using an actual `<button type="button">` element.
2. The link

`classes.parent` is probably applicable to both forms.

kasperg (approved):
:+1: LGTM

jeppekroghitk (approved):

PR #386, opened 2023-03-09
Fix/autosuggest accessibility

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFSOEG-433
https://reload.atlassian.net/browse/DDFSOEG-432
https://reload.atlassian.net/browse/DDFSOEG-431

#### Description
According to https://www.w3.org/TR/wai-aria-1.1/#option & https://www.w3.org/TR/wai-aria-1.1/#listbox our autosuggest wasn't living up to accessibility criteria, where unordered lists of role "listbox" needs to have direct list items that are of role "option" OR own them. Plus an unordered list without a role needs to be a parent to list items.
This PR fixes these problems.

#### Screenshot of the result
n/a

#### Checklist

n/a

### DIFF
--- a/src/components/autosuggest-category/autosuggest-category.tsx
+++ b/src/components/autosuggest-category/autosuggest-category.tsx
@@ -25,33 +25,36 @@ const AutosuggestCategory: FC<AutosuggestCategoryProps> = ({
 }) => {
   const t = useText();
   return (
-    <>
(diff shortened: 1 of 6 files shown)


Adamik10:
In the end we chose to solve this in the design system and change HTML structure for autosuggest.
https://github.com/danskernesdigitalebibliotek/dpl-design-system/pull/208

PR #387, opened 2023-03-09
Refactor headings tree structure for `Material Page`

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFSOEG-452
https://reload.atlassian.net/browse/DDFSOEG-422
https://reload.atlassian.net/browse/DDFSOEG-423

#### This PR is refactoring the headings tree structure for the `Material Page`
To improve user navigation and content comprehension

#### Screenshot of the result
image

#### Checklist


src/components/Heading/Heading.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+type HeadingProps = {
+  children: React.ReactNode;
+  level?: HeadingLevelType;
+  className?: string;
+};
+
+const Heading: React.FunctionComponent<HeadingProps> = ({
+  children,
+  level = "h2",
```
kasperg, line 13: Please remove the default value for this component.

Having a proper heading structure is important and using default values prevents the developer from having to take it into consideration.

src/components/Heading/Heading.tsx
```diff
@@ -0,0 +1,26 @@
...
+  className
+}) => {
+  const HeadingTag = {
+    h2: "h2",
+    h3: "h3",
+    h4: "h4",
+    h5: "h5"
+  }[level] as keyof JSX.IntrinsicElements;
```
kasperg, line 21:
```suggestion
  const HeadingTag = level as keyof JSX.IntrinsicElements;
```

Please reconsider the structure above. It takes a string literal and uses an object with the same keys to map to strings with the same values. All in all the allowed values are repeated three times. That should not be necessary. The suggestion above tries to fix that. Perhaps try that out or consider other options.

kasperg (changes requested):
:+1: but a bit more work is needed here.

kasperg (approved):
:+1: LGTM

PR #388, opened 2023-03-13
Update Publizon URL + Refactor components to avoid re-renders

kasperbirch1:
#### Refactor components to avoid re-renders

This pull request includes the following changes:

- Made changes to `ModalReservationFormText`, `useAlternativeAvailableManifestation`, `ModalReservationFormText` and `handleSetImageLoaded` to improve performance.

#### Checklist

### DIFF
--- a/src/components/cover/cover.tsx
+++ b/src/components/cover/cover.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from "react";
+import React, { useCallback, useState } from "react";
 import clsx from "clsx";
 import { useGetCoverCollection } from "../../core/cover-service-api/cover-service";
 import { GetCoverCollectionType } from "../../core/cover-service-api/model";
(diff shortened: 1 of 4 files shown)


src/components/cover/cover.tsx
```diff
@@ -28,7 +28,9 @@ export const Cover = ({
   shadow
 }: CoverProps) => {
   const [imageLoaded, setImageLoaded] = useState<boolean | null>(null);
-  const handleSetImageLoaded = () => setImageLoaded(true);
+  const handleSetImageLoaded = useCallback(() => {
+    setImageLoaded(true);
+  }, []);
```
kasperg, line 33: I appreciate the intent here but I wonder if this callback should in fact not have any dependencies at all - such as the `id` or `idType` which will effect the source of the final image. If you think it is OK then move along :smile:

kasperg (changes requested):
Without further reasoning I do not agree with the change to the publizon tokens.

Other changes are good - probably :smile:

src/components/cover/cover.tsx
kasperbirch1: I am not sure if they should be there? 😅 If I put them in the dependency array I got this warning:
`React Hook useCallback has unnecessary dependencies: 'id' and 'idType'. Either exclude them or remove the dependency array.eslint`

kasperg (approved):
:+1: LGTM

jeppekroghitk (approved):

PR #389, opened 2023-03-13
Add Global Escape Key Handler for Closing Modals

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFSOEG-453

#### This PR adds a global escape key handler to close modals.
When the user presses the escape key the `closeLastModal` action is dispatched, and it closes the last opened modal.

By implementing this feature, we ensure that the application is more user-friendly, and users can easily close modals without having to use tap key to search for a close button. Furthermore, this addition helps maintain consistency in user interactions across different applications, as the escape key is widely used for closing modals.

#### Checklist

##### This PR will prevent the re-renders
https://github.com/danskernesdigitalebibliotek/dpl-react/pull/388

### DIFF
--- a/src/core/mount.js
+++ b/src/core/mount.js
@@ -75,6 +75,11 @@ function reset() {
 }
 
 function init() {
+  // We only want to close the single last modal when the user hits escape.
+  // Consequently we only want a single event listener. We cannot guarantee this
+  // on the app (or component) level as a page may contain multiple apps. 
+  // init provides a single entry point for loading all apps and is suitable for our
+  // purpose.
   document.addEventListener("keydown", (e) => {
     if (e.key === "Escape") {
       store.dispatch(closeLastModal());
--- a/src/core/mount.js
+++ b/src/core/mount.js
@@ -3,8 +3,9 @@ import { render } from "react-dom";
 import { withErrorBoundary } from "react-error-boundary";
 import { setToken } from "./token";
 import Store from "../components/store";
-import { persistor } from "./store";
+import { persistor, store } from "./store";
 import ErrorBoundaryAlert from "../components/error-boundary-alert/ErrorBoundaryAlert";
+import { closeLastModal } from "./modal.slice";
 
 /**
  * We look for containers and corresponding applications.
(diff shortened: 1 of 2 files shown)


src/core/utils/useKeyPress.ts
kasperg, line 1: Please replace this implementation with the hooks provided by [`react-use`](https://github.com/streamich/react-use).

They have a few different options. I encourage you to try them out in their storybook to see the difference and choose accordingly.

src/core/utils/modal.tsx
```diff
@@ -40,6 +41,16 @@ function Modal({
   const dispatch = useDispatch();
   const { modalIds } = useSelector((s: ModalIdsProps) => s.modal);
 
+  const close = useCallback(() => {
+    // This is to ensure that it only closes the top modal (multiple modals can be open at once)
+    const index = modalIds.indexOf(modalId);
+    if (index >= 0 && modalIds[index] === modalId) {
```
kasperg, line 47: Please reconsider this code. 

I appreciate the intention in the comment but I do not understand the implementation. If `const index = modalIds.indexOf(modalId)` then I cannot see when `modalIds[index] === modalId` would not be true. If that is the case then it is only a test for `index >= 0`. I had expected a test for whether the modal id was the last entry in the array.

kasperg (changes requested):

src/core/mount.js
```diff
@@ -74,6 +75,12 @@ function reset() {
 }
 
 function init() {
+  document.addEventListener("keydown", (e) => {
+    if (e.key === "Escape") {
+      store.dispatch(closeLastModal());
+    }
+  });
+
```
kasperg, line 83:
```suggestion
function init() {
  // We only want to close the single last modal when the user hits escape.
  // Consequently we only want a single event listener. We cannot guarantee this
  // on the app (or component) level as a page may contain multiple apps. 
  // init provides a single entry point for loading all apps and is suitable for our
  // purpose.
  document.addEventListener("keydown", (e) => {
    if (e.key === "Escape") {
      store.dispatch(closeLastModal());
    }
  });

```

We agree on the solution but I think this change warrants some documentation/inline code comment about why it is needed and why it is placed at this level in the codebase.

kasperg (approved):
:+1: but please consider additional commentary.

jeppekroghitk (approved):

PR #391, opened 2023-03-13
Add `aria-label` to `ReservationFormListItem`

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFSOEG-434

#### This PR is a proposed solution to add an `aria-label` to the `ReservationFormListItem` component. 
As screen readers need to know what the change buttons do in the reservation model, this change will improve accessibility and user experience for those who rely on assistive technologies.

#### Checklist

### DIFF
--- a/src/apps/material/material.entry.tsx
+++ b/src/apps/material/material.entry.tsx
@@ -14,6 +14,10 @@ interface MaterialEntryTextProps {
   approveReservationText: string;
   cantReserveText: string;
   cantViewReviewText: string;
+  changeEmailText: string;
+  changeInterestPeriodText: string;
+  changePickupLocationText: string;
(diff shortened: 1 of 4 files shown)


src/apps/material/material.entry.tsx
```diff
@@ -36,6 +40,7 @@ interface MaterialEntryTextProps {
   editionsText: string;
   editionText: string;
   etAlText: string;
+  expandMoreText: string;
```
kasperg, line 43: :+1: Appreciate this cleanup.

kasperg (approved):
:+1: LGTM

jeppekroghitk (approved):

PR #393, opened 2023-03-16
Fix/accessibility for find on shelf

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFSOEG-440

#### Description
This PR fixes an issue with the find on shelf modal content. Accessibility for screen readers for the list of locations wasn't taken into account. 

#### Screenshot of the result
n/a

#### Checklist

n/a


kasperg (approved):
:+1: Perhaps this structure could be improved even more by viewing all the manifestation lists as one giant table but let's see how this goes.

PR #394, opened 2023-03-16
Fix/accessibility for selects

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFSOEG-424

#### Description
On the material page inside the reservation modal - two sub-modals had <select> elements without any aria-labels, which was not living up to accessibility requirements - https://www.w3.org/Translations/WCAG21-da/#headings-and-labels .

### DIFF
--- a/src/components/reservation/forms/ModalReservationFormSelect.tsx
+++ b/src/components/reservation/forms/ModalReservationFormSelect.tsx
@@ -14,7 +14,7 @@ export interface ModalReservationFormSelectProps {
   items: { label: string; value: string }[];
   defaultSelectedItem: string;
   selectHandler: (value: string) => void;
-  ariaLabel?: string;
+  ariaLabel: string;
 }
 
 const modalProps = (
@@ -34,7 +34,7 @@ const ModalReservationFormSelect = ({
   items,
   defaultSelectedItem,
   selectHandler,
-  ariaLabel = "modal-reservation-form-select"
+  ariaLabel
 }: ModalReservationFormSelectProps) => {
   const { close } = useModalButtonHandler();
   const t = useText();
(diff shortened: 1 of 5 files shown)


src/components/reservation/forms/ModalReservationFormSelect.tsx
```diff
@@ -32,7 +33,8 @@ const ModalReservationFormSelect = ({
   header,
   items,
   defaultSelectedItem,
-  selectHandler
+  selectHandler,
+  ariaLabel = "modal-reservation-form-select"
```
kasperg, line 37: Please reconsider making `ariaLabel` optional.

I do not think this makes for a suitable default value to be handled by screen readers.

src/components/reservation/forms/NoInterestAfterModal.tsx
```diff
@@ -33,6 +33,7 @@ const NoInterestAfterModal = ({
       items={formatInterestPeriods}
       defaultSelectedItem={String(selectedInterest)}
       selectHandler={(value: string) => setSelectedInterest(Number(value))}
+      ariaLabel={t("modalReservationFormNoInterestAfterHeaderTitleText")}
```
kasperg, line 36: I understand what you are getting at with this approach of duplicating the title but I do not think it makes for a very good solution.

Either we should:

- Create a separate text entry for the dropdown aria label `modalReservationFormNoInterestAfterLabelText`
- Mark the dropdown as labelled by the header of the modal using `aria-labelled-by`. This requires setting a unique id for the modal header and passing this to the children. It gets somewhat complex so if you want to go that route come talk to me about it.

kasperg (changes requested):
:+1: this needs a bit more consideration.

src/components/reservation/forms/ModalReservationFormSelect.tsx
Adamik10: You're right, it's an absolutely terrible value. I for some reason thought I was working with our good old dataCy props.

kasperg (approved):
:+1: LGTM

jeppekroghitk (approved):

PR #396, opened 2023-03-21
Feature/5369 create patron

sinejespersen:
#### Link to issue

#### Description

#### Screenshot of the result

#### Checklist


yepzdk (approved):

src/apps/create-patron-user-info/CreatePatron.dev.tsx
```diff
@@ -0,0 +1,134 @@
...
+import pincodeArgs from "../../core/storybook/pincodeArgs";
+
+export default {
+  title: "Apps / Create patron",
+  component: CreatePatron,
+  argTypes: {
+    ...serviceUrlArgs,
+    ...pincodeArgs,
+    blacklistedPickupBranchesConfig: {
+      defaultValue: "FBS-751032,FBS-751031,FBS-751009,FBS-751027,FBS-751024",
```
cableman, line 14: Don't known these values all (to me known) agency id's (ISIL number) starts with `DK-xxxxxx`?

cableman (approved):

src/apps/create-patron-user-info/CreatePatron.dev.tsx
sinejespersen: I copy pasted, used throughout the system and I am actually not sure it makes a difference as it seems to work in the drupal setup :)

src/apps/create-patron-user-info/CreatePatron.entry.tsx
```diff
@@ -0,0 +1,55 @@
...
+  createPatronConfirmButtonText: string;
+  createPatronCancelButtonText: string;
+  patronContactEmailLabelText: string;
+  patronContactEmailCheckboxText: string;
+  createPatronChangePickupHeaderText: string;
+  createPatronChangePickupBodyText: string;
+  createPatronHeaderText: string;
+  createPatronInvalidSSNHeaderText: string;
+  createPatronInvalidSSNBodyText: string;
+}
```
spaceo, line 42: Just an advise: It is easier for the person that implements the data props in dpl-cms if the props are in alphabetic order.

src/apps/create-patron-user-info/UserInfo.tsx
```diff
@@ -0,0 +1,128 @@
...
+      mutate(
+        {
+          data: { cprNumber: cpr, patron, pincode: pin }
+        },
+        {
+          onSuccess: () => {
+            redirectTo(redirectOnUserCreatedUrl);
+          },
+          // todo error handling, missing in figma
+          onError: () => {}
```
spaceo, line 58: Since the ErrorBoundary system is catching services throwing an error, there is no need for an error handler here

src/core/utils/helpers/general.ts
```diff
@@ -405,6 +405,44 @@ export const filterLoansNotOverdue = (loans: LoanType[], warning: number) => {
...
+    if (Number(year) < 21) {
+      prefix = "20";
+    } else {
+      prefix = "19";
+    }
+    const yearWithPrefix = Number(`${prefix}${year}`);
+
+    return new Date(
+      Date.UTC(
+        Number(yearWithPrefix),
```
spaceo, line 424: Should not be necessary to cast `yearWithPrefix` to an integer twice.

src/core/utils/helpers/general.ts
```diff
@@ -405,6 +405,44 @@ export const filterLoansNotOverdue = (loans: LoanType[], warning: number) => {
...
+  const dateSegments = cpr.substring(0, 6).match(/.{1,2}/g);
+
+  if (dateSegments) {
+    const [day, month, year] = dateSegments;
+    let prefix = "";
+    if (Number(year) < 21) {
+      prefix = "20";
+    } else {
+      prefix = "19";
+    }
```
spaceo, line 419: So this is valid for a little under 100 years. Let's see how long the application stands ;)

src/core/utils/helpers/general.ts
```diff
@@ -405,6 +405,44 @@ export const filterLoansNotOverdue = (loans: LoanType[], warning: number) => {
...
+  return null;
+}
+
+export const isCprValid = (cpr: string, minAge: number) => {
+  const cprDate = getDateFromCpr(cpr);
+  if (cprDate === null) return false;
+
+  const age = dayjs().diff(dayjs(cprDate), "year");
+  return age > minAge;
+};
```
spaceo, line 444: What is being considered being a "valid" cpr? Is it just the age of the patron? Otherwise you have to follow the documentation for validating CPR's.

spaceo (changes requested):
I have some questions I would like to have clarified before approval ;)

src/core/utils/helpers/general.ts
sinejespersen: good point

src/core/utils/helpers/general.ts
sinejespersen: fixed

src/apps/create-patron-user-info/UserInfo.tsx
```diff
@@ -0,0 +1,128 @@
...
+  const [validCpr] = useState<boolean>(isCprValid(cpr, minAge));
+  const { mutate } = useCreateV4();
+  const [patron, setPatron] = useState<PatronSettingsV3>({
+    preferredPickupBranch: "",
+    receiveEmail: false,
+    receivePostalMail: false,
+    receiveSms: false,
+    phoneNumber: "",
+    emailAddress: ""
+  });
```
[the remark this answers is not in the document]
sinejespersen: What should the defaults be?

src/apps/create-patron-user-info/UserInfo.tsx
```diff
@@ -0,0 +1,128 @@
...
+        {
+          onSuccess: () => {
+            redirectTo(redirectOnUserCreatedUrl);
+          },
+          // todo error handling, missing in figma
+          onError: () => {}
+        }
+      );
+    }
+  };
```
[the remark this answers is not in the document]
sinejespersen: the change patron updates the patron that is later saved. Do you have a suggestion for a better name?

"Why are you not creating a common place in core for handling user data, when we have talked about generalizing this in several other PR´s?" -> As we talked about, they differ. I am not sure any one person has all the details to create a general function for this. it seem to differ slightly for every new use case.

src/apps/create-patron-user-info/UserInfo.tsx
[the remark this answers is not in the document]
spaceo: I might as well be these defaults. I just think they should be centralized

src/apps/create-patron-user-info/CreatePatron.tsx
```diff
@@ -1,14 +1,17 @@
...
+import { useConfig } from "../../core/utils/config";
 
 interface CreatePatronProps {
   userToken: string;
 }
 
 const CreatePatron: FC<CreatePatronProps> = ({ userToken }) => {
   const [cpr, setCpr] = useState<string | null>(null);
+  const config = useConfig();
+  const loginUrlConfig = config("loginUrlConfig");
```
spaceo, line 12: We are using the url hoc functionality for URLs in other places. Please use that

spaceo (changes requested):
url handling needs to be changed

src/apps/create-patron-user-info/CreatePatron.tsx
sinejespersen: I can, but then I would have to use the url, and get the href from that, which I thought was a little bit ugly. Should I do that?

src/apps/create-patron-user-info/UserInfo.tsx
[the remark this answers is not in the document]
sinejespersen: In something you (Reload) created, I am pretty sure the defaults are receivePostalMail: true and receiveEmail: true.

src/apps/create-patron-user-info/CreatePatron.tsx
spaceo: No you can just cast the URL object to string: `String(someUrlObject)`.

src/apps/create-patron-user-info/UserInfo.tsx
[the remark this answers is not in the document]
spaceo: My main point is that we should have an object with defaults somewhere central in `core`. I am not getting into what the value of the defaults should be.

spaceo (changes requested):
@sinejespersen The default patron values is a minor matter. However I think the url config handling needs to be changed

spaceo (approved):
LGTM 👍

PR #397, opened 2023-03-21
Fix/reservation list additions

jeppekroghitk:
#### Link to issue

N/A

#### Description

Based on test feedback, this PR adds and edits functionality related to the "reservation list" react component.
- Added boolean for allowing removing reservation when it is ready to pickup
- Added booleans for toggling predefined interest periods individually
- Generalized strtobool helper function to become applicable generally
- Swapped param value in reservation deletion call

#### Screenshot of the result

N/A

#### Checklist

### DIFF
--- a/src/apps/reservation-list/modal/reservation-details/reservation-details.tsx
+++ b/src/apps/reservation-list/modal/reservation-details/reservation-details.tsx
@@ -34,8 +34,9 @@ const ReservationDetails: FC<ReservationDetailsProps & MaterialProps> = ({
     "reservationDetailAllowRemoveReadyReservationsConfig"
   );
   const isDigital = !!reservation.identifier;
-  const userCanRemoveReadyReservations =
-    state === "readyForPickup" && strToBool(allowRemoveReadyReservation);
+  const allowUserRemoveReadyReservations =
+    (state === "readyForPickup" && strToBool(allowRemoveReadyReservation)) ||
+    state !== "readyForPickup";
 
   return (
     <div className="modal-details__container">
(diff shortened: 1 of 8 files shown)


src/apps/reservation-list/modal/reservation-details/reservation-details.tsx
```diff
@@ -23,11 +25,18 @@ const ReservationDetails: FC<ReservationDetailsProps & MaterialProps> = ({
...
   const { state, identifier, numberInQueue } = reservation;
   const { authors, pid, year, title, description, materialType } =
     material || {};
   const branches = useGetBranches("blacklistedPickupBranchesConfig");
+  const allowRemoveReadyReservation = config(
+    "reservationDetailAllowRemoveReadyReservationsConfig"
+  );
   const isDigital = !!reservation.identifier;
+  const allowUserRemoveReadyReservations =
+    (state === "readyForPickup" && strToBool(allowRemoveReadyReservation)) ||
```
spaceo, line 38: "readyForPickup" should be defined as an `object as const`or an `enum`.
Great with the `strToBool` function. I just think the name is too generic. Is it possible to give it a more descriptive name. What is the purpose of it?

src/components/reservation/helper.ts
```diff
@@ -17,28 +17,35 @@ import {
...
-export const hardcodedInterestPeriods = (t: UseTextFunction) => {
-  return {
-    "30": t("oneMonthText"),
-    "60": t("twoMonthsText"),
-    "90": t("threeMonthsText"),
-    "180": t("sixMonthsText"),
-    "360": t("oneYearText")
-  };
+  interestPeriods.forEach(([config, key, text]) => {
+    if (strToBool(t(config))) {
```
spaceo, line 38: This is really cryptic to me :) Why are you translating some config and then checking if it is `"1"`?
Could be a valid reason. But I get the impression that it could be done otherwise.

spaceo (changes requested):
Please look at my comments ;)

src/components/reservation/helper.ts
jeppekroghitk: The config is a bool, configurable for each interest period. The bool is set up as an "argtype", will be defined as a checkbox in drupal, either returning "1" or "0", ultimately making the visibility of each interest period configuarable in drupal

sinejespersen (approved):

spaceo (approved):
LGTM 👍

PR #399, opened 2023-03-22
Add 'FocusTrap'  in `ReservationSucces`

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFSOEG-447

#### Add 'FocusTrap' to improve accessibility in `ReservationSucces`

FocusTrap helps to trap the focus within the modal when it is open, preventing users from accidentally interacting with elements outside of the modal.

#### Checklist


kasperg (approved):
:+1: LGTM

jeppekroghitk (approved):

PR #401, opened 2023-03-22
Add 'aria-expanded' to `MaterialDetailsList`

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFSOEG-426

#### Add 'aria-expanded' to `MaterialDetailsList`
The aria-expanded attribute has been added to the MaterialDetailsList component to improve accessibility for users with disabilities. This attribute indicates the current state of the element, whether it is expanded or collapsed, and provides information to assistive technologies like screen readers.

#### Checklist

### DIFF
--- a/src/apps/material/material.tsx
+++ b/src/apps/material/material.tsx
@@ -239,7 +239,11 @@ const Material: React.FC<MaterialProps> = ({ wid }) => {
         disclosureIconExpandAltText=""
         dataCy="material-details-disclosure"
       >
-        <MaterialDetailsList className="pl-80 pb-48" data={detailsListData} />
+        <MaterialDetailsList
+          id="main-details"
+          className="pl-80 pb-48"
+          data={detailsListData}
+        />
       </Disclosure>
       {hasReview && hasReview.length > 0 && (
         <DisclosureControllable
--- a/src/apps/material/material.tsx
+++ b/src/apps/material/material.tsx
@@ -240,7 +240,7 @@ const Material: React.FC<MaterialProps> = ({ wid }) => {
         dataCy="material-details-disclosure"
       >
         <MaterialDetailsList
-          id="main-details"
+          id="material-details"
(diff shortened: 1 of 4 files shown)


src/apps/material/material.tsx
```diff
@@ -239,7 +239,11 @@ const Material: React.FC<MaterialProps> = ({ wid }) => {
         disclosureIconExpandAltText=""
         dataCy="material-details-disclosure"
       >
-        <MaterialDetailsList className="pl-80 pb-48" data={detailsListData} />
+        <MaterialDetailsList
+          id="main-details"
```
Adamik10, line 243: Okay here I go talking about naming things again. 🙈 But maybe I would consider aligning these id's - one ID is defined here, and the other one inside the `MaterialManifestationItem` on line 102. In this case the main details component would have an id of `main-details` and the other manifestation details would be called something like `material-details-870970-basis:52643414` and so on. Would it make more sense if it was `material-details` for the main one & `material-details-870970-basis:52643414` using IDs as it is for the other ones?
It's food for thought - but I am approving this PR. Good job 🙂

Adamik10 (approved):
Good job :) 
Approved with a comment. You can decide what you think about it.

src/apps/material/material.tsx
```diff
@@ -239,7 +239,11 @@ const Material: React.FC<MaterialProps> = ({ wid }) => {
         disclosureIconExpandAltText=""
         dataCy="material-details-disclosure"
       >
-        <MaterialDetailsList className="pl-80 pb-48" data={detailsListData} />
+        <MaterialDetailsList
+          id="material-details"
```
kasperg, line 243: Consider adding the `wid` (work id) to the `id` string.

This helps to ensure uniqueness even if different material apps are used on the same page.

jeppekroghitk (approved):

PR #407, opened 2023-03-27
improve 'blockStatus' check in `MaterialButtonsPhysical`

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFSOEG-486

#### This PR improves the 'blockStatus' check for the `MaterialButtonsPhysical` component
by importing and using the `isEmpty` function from lodash. This ensures a more accurate check for the `userData.patron.blockStatus` property.

#### Checklist


kasperg (approved):
:+1: Nice use of Lodash for simplicity.

PR #408, opened 2023-03-27
Improve Screen Reader Accessibility + Add TypeScript helper Function `invalidSwitchCase`

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFSOEG-435
https://reload.atlassian.net/browse/DDFSOEG-439
https://reload.atlassian.net/browse/DDFSOEG-443

#### Description
This pull request includes several improvements to enhance accessibility for users who rely on screen readers, as well as adding a TypeScript helper function named `invalidSwitchCase`. 

The primary changes are as follows:

- Added two translation functions, `getFacetFieldTranslation` and `getReservationModalTypeTranslation`, which utilize the `invalidSwitchCase` helper function to ensure all cases are covered in a switch function.

- Updated various elements with appropriate translations for their 'aria-label' attributes and other relevant places.

#### Checklist

### DIFF
--- a/src/components/facet-browser/helper.ts
+++ b/src/components/facet-browser/helper.ts
@@ -5,6 +5,7 @@ import {
 } from "../../core/dbc-gateway/generated/graphql";
 import useGetCleanBranches from "../../core/utils/branches";
 import { Filter, FilterItemTerm } from "../../core/filter.slice";
+import invalidSwitchCase from "../../core/utils/helpers/invalid-switch-case";
 
 export const allFacetFields = [
   FacetField.MainLanguages,
@@ -100,11 +101,7 @@ export function getAllFilterPathsAsString(filterObject: {
   return allFilterPathsAsString;
 }
 
-const handleFacetFieldTranslationError = (name: string): never => {
-  throw new Error(`Didn't expect "${name}" in getFacetFieldTranslation`);
-};
-
-export const getFacetFieldTranslation = (name: string): string => {
+export const getFacetFieldTranslation = (name: FacetField) => {
   switch (name) {
     case FacetField.AccessTypes:
       return "facetAccessTypesText";
@@ -133,7 +130,7 @@ export const getFacetFieldTranslation = (name: string): string => {
     case FacetField.WorkTypes:
       return "facetWorkTypesText";
     default:
-      return handleFacetFieldTranslationError(name);
+      return invalidSwitchCase<string>(name);
   }
 };
 
--- a/src/components/facet-line/FacetLineFilters.tsx
+++ b/src/components/facet-line/FacetLineFilters.tsx
@@ -1,6 +1,7 @@
 import React, { memo } from "react";
 import useFilterHandler from "../../apps/search-result/useFilterHandler";
 import {
+  FacetField,
   FacetResult,
   FacetValue
 } from "../../core/dbc-gateway/generated/graphql";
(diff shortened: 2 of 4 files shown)


src/components/facet-line/FacetLineFilters.tsx
```diff
@@ -60,7 +60,7 @@ const FacetLineFilters: React.FunctionComponent<FacetLineFiltersProps> = ({
                   value: ""
                 }}
                 options={formatValuesToDropdown(name, values)}
-                ariaLabel={name}
+                ariaLabel={t(`facet${upperFirst(name)}Text`)}
```
kasperg, line 63: This creates a tricky situation for managing translations as the code does not expose any information about what strings we can expect. This also shows by the fact that no facet translation controls are added in Storybook.

As I see it the possible facet names are limited by [the `FacetField` enumeration](https://github.com/reload/dpl-react/blob/main/src/core/dbc-gateway/generated/graphql.tsx#L306-L306). 


If you want to take a stab at it consider creating a function for generating translations for facet names and use TypeScript to ensure that all values are covered and translations are explicit but I do not think it is necessary to move this PR along.

src/components/reservation/forms/ModalReservationFormSelect.tsx
```diff
@@ -22,9 +23,9 @@ const modalProps = (
 ) => ({
   modalId: modalReservationFormId(type),
   screenReaderModalDescriptionText: t(
-    `screenReaderModalDescription${type.toUpperCase()}Text`
+    `screenReaderModalDescription${upperFirst(type)}Text`
```
kasperg, line 26: See comment about regarding dynamically generated translation strings. This case and the rest of this PR is not so bad though.

src/apps/material/material.entry.tsx
```diff
@@ -97,16 +101,16 @@ interface MaterialEntryTextProps {
...
-  orderDigitalCopyFeedbackOkText: string;
   orderDigitalCopyFeedbackErrorAgencyNotSubscribedText: string;
   orderDigitalCopyFeedbackErrorInvalidPickupBranchText: string;
   orderDigitalCopyFeedbackErrorMissingClientConfigurationText: string;
   orderDigitalCopyFeedbackErrorPidNotReservableText: string;
   orderDigitalCopyFeedbackErrorUnauthenticatedUserText: string;
+  orderDigitalCopyFeedbackOkText: string;
   orderDigitalCopyFeedbackTitleText: string;
+  orderDigitalCopyModalCloseModalAriaLabelText: string;
+  orderDigitalCopyModalScreenReaderModalDescriptionText: string;
```
kasperg, line 113: :+1: for cleanup.

kasperg (approved):
:+1: LGTM

src/components/facet-browser/helper.ts
```diff
@@ -100,4 +100,41 @@ export function getAllFilterPathsAsString(filterObject: {
   return allFilterPathsAsString;
 }
 
+const handleFacetFieldTranslationError = (name: string): never => {
+  throw new Error(`Didn't expect "${name}" in getFacetFieldTranslation`);
+};
+
+export const getFacetFieldTranslation = (name: string): string => {
```
kasperg, line 107: As far as I can tell this addition needs some more work and should not be merged as is.

kasperg (changes requested):

sinejespersen (approved):

kasperg (approved):
:+1: LGTM

PR #409, opened 2023-03-28
Add 'aria-labelledby' to `Link` + Change name 'description' to 'altText'

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFSOEG-421
https://reload.atlassian.net/browse/DDFSOEG-427

#### This pull request add 'aria-labelledby' to `Link` + Change name 'description' to 'altText'

When the 'coverSrc' property in the `Cover` component is invalid, the image won't be displayed, leading to an empty link. To address this, we use the 'aria-labelledby' attribute to associate the empty link with the title of the search result item. 

We generate a unique identifier for the title using the useId() hook, which is assigned to the id attribute of the title and used as the aria-labelledby value for the Link

Additionally, the description prop in the Cover component is renamed to altText.

This improves the overall experience for users who rely on screen readers.

#### Screenshot of the result
image

#### Checklist

### DIFF
--- a/src/components/cover/cover-image.tsx
+++ b/src/components/cover/cover-image.tsx
@@ -26,7 +26,7 @@ const CoverImage: FC<CoverImageType> = ({
       { "cover__img--shadow": shadow }
     )}
     src={src}
-    alt={altText}
+    alt={altText || ""}
   />
 );
 
--- a/src/components/cover/cover-image.tsx
+++ b/src/components/cover/cover-image.tsx
@@ -3,7 +3,7 @@ import clsx from "clsx";
 
 type CoverImageType = {
   src: string;
-  description?: string;
+  altText?: string;
   animate: boolean;
   setImageLoaded: () => void;
   shadow?: boolean;
@@ -11,7 +11,7 @@ type CoverImageType = {
 
 const CoverImage: FC<CoverImageType> = ({
   src,
-  description,
+  altText,
   animate,
   setImageLoaded,
   shadow
@@ -26,7 +26,7 @@ const CoverImage: FC<CoverImageType> = ({
       { "cover__img--shadow": shadow }
     )}
     src={src}
-    alt={description || ""}
+    alt={altText}
   />
 );
 
--- a/src/components/search-result-list/search-result-list-item/search-result-list-item-cover.tsx
+++ b/src/components/search-result-list/search-result-list-item/search-result-list-item-cover.tsx
@@ -4,7 +4,6 @@ import { Cover, CoverProps } from "../../cover/cover";
 type SearchResultListItemCoverProps = Omit<CoverProps, "animate" | "size">;
 const SearchResultListItemCover: React.FC<SearchResultListItemCoverProps> = ({
   id,
-  description,
   url,
   tint,
   linkAriaLabelledBy
@@ -14,7 +13,6 @@ const SearchResultListItemCover: React.FC<SearchResultListItemCoverProps> = ({
       animate
       id={id}
       size="small"
-      description={description}
       url={url}
       tint={tint}
       linkAriaLabelledBy={linkAriaLabelledBy}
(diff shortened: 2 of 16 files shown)


src/components/search-result-list/search-result-list-item/search-result-list-item.tsx
```diff
@@ -54,6 +54,7 @@ const SearchResultListItem: React.FC<SearchResultListItemProps> = ({
   resultNumber,
   dataCy = "search-result-list-item"
 }) => {
+  const searchResultListItemTitleId = useId();
```
kasperg, line 57: [Be aware that we need a `identifierPrefix` when we use `useId` and have multiple apps on the same page](https://react.dev/reference/react/useId).

You probably need to look into `mount.js` to address this.

kasperg: Consider a shorter variable name.

`searchResultListItemTitleId` gets kind of lengthy when we are already in the `SearchResultListItem` component. In my opinion `titleId` would provide the same information and be a lot shorter :smile:.

src/components/search-result-list/search-result-list-item/search-result-list-item-cover.tsx
```diff
@@ -6,16 +6,19 @@ const SearchResultListItemCover: React.FC<SearchResultListItemCoverProps> = ({
...
+
+  linkAriaLabelledBy
 }) => {
   return (
     <Cover
       animate
       id={id}
       size="small"
-      description={String(description)}
+      description={description}
```
kasperg, line 18: Please consider if both `description` and `linkAriaLabelledBy` is appropriate here.

The description (which will become an `alt` value if provided) *and* `linkAriaLabelledBy` serves the same purpose. I understand that the `Cover` component accepts both but as I see it for the `SearchResultListItemCover` we only want to provide one of them: `linkAriaLabelledBy`.

src/components/search-result-list/search-result-list-item/search-result-list-item-cover.tsx
```diff
@@ -6,16 +6,19 @@ const SearchResultListItemCover: React.FC<SearchResultListItemCoverProps> = ({
   id,
   description,
   url,
-  tint
+  tint,
+
+  linkAriaLabelledBy
```
kasperg, line 11: Please look into code formatting here.

src/components/cover/cover-image.tsx
```diff
@@ -26,7 +26,7 @@ const CoverImage: FC<CoverImageType> = ({
       { "cover__img--shadow": shadow }
     )}
     src={src}
-    alt={description || ""}
+    alt={altText}
```
kasperg, line 29: Consider what will happen if `altText` is not provided.

As I see it it was intentional that we used the empty string if no description was provided as [this marks the image as presentational](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#attributes). I fear if we remove the `alt` attribute entirely if no value is provided. That may be intentional but I do not see anywhere else where we switch to passing an empty array.

kasperg (changes requested):
:+1: A few things for your consideration.

src/core/mount.js
```diff
@@ -7,6 +7,10 @@ import { persistor, store } from "./store";
 import ErrorBoundaryAlert from "../components/error-boundary-alert/ErrorBoundaryAlert";
 import { closeLastModal } from "./modal.slice";
 
+// Todo: This should be updated because React.createElement() is deprecated.
+// After the update, ensure that prefixes (identifierPrefix) are specified for all apps.
+// This will guarantee unique IDs everywhere useID() is utilized.
+
```
kasperg, line 13: Please move and reconsider this.

1. This is in direct relation to the `render()` call. Please move the comment directly above this call. Otherwise chances are that it will be overlooked
2. It is `render()` that is deprecated - not `createElement()`

PR #410, opened 2023-03-29
Add aria-hidden attribute to autosuggest dividers

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFSOEG-434

#### This PR addresses the accessibility issue with the autosuggest
by adding the `aria-hidden` attribute to the dividers. This ensures that the dividers are not read by screen readers, providing a better user experience for those using assistive technology.

#### Checklist


spaceo (approved):
LGTM 👍

kasperg:
This should not be merged. It is superseeded by work being done in danskernesdigitalebibliotek/dpl-design-system#208 which will restructure the autocomplete with proper semantic HTML.

PR #411, opened 2023-03-29
Add 'lang' attribute to titles if they are not Danish

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFSOEG-436

#### This pull request adds new functionality to improve accessibility by assigning the 'lang' attribute to titles
A new function called `getManifestationLanguageIsoCode` has been implemented to accomplish this.

The `getManifestationLanguageIsoCode` function takes an array of LanguagesFragment objects and returns the ISO code of the main language. If multiple unique main languages are present, it defaults to the language defined in the document's HTML tag.

In addition, 'LanguagesFragment' fragments have been added to enhance code generation (Types)

#### Screenshot of the result
image

image

image

#### Checklist

### DIFF
--- a/src/apps/material/helper.ts
+++ b/src/apps/material/helper.ts
@@ -1,4 +1,4 @@
-import { compact, groupBy, uniqBy, uniq } from "lodash";
+import { compact, groupBy, uniqBy, uniq, head } from "lodash";
 import {
   constructModalId,
   getMaterialTypes,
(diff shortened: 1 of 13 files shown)


src/apps/material/helper.ts
```diff
@@ -107,6 +107,23 @@ export const getManifestationLanguages = (manifestation: Manifestation) => {
   );
 };
 
+export const getManifestationLanguageIsoCode = (
+  manifestation: Manifestation[]
+) => {
+  const uniqueLanguages = uniqBy(
+    manifestation.map((m) => m.languages),
+    "main[0].isoCode"
+  );
```
kasperg, line 116: Please reconsider this approach to determining unique languages.

As I see it is will not return unique `Language` object, it will return unique `Languages` objects based on the iso code of the first main language.

I think a more correct approach would be to return unique `Language` objects that are defined as main languages for the manifestations.

src/apps/material/helper.ts
```diff
@@ -107,6 +107,23 @@ export const getManifestationLanguages = (manifestation: Manifestation) => {
...
+  if (
+    uniqueLanguages.length === 1 &&
+    uniqueLanguages[0]?.main?.[0]?.isoCode &&
+    uniqueLanguages[0].main[0].isoCode !== "dan"
+  ) {
+    return uniqueLanguages[0].main[0].isoCode;
+  }
+  return undefined;
```
kasperg, line 124: Please reconsider if it is appropriate to return `undefined` if the language is Danish and if we only want to return a language if the manifestations only contain one language.

I understand what you are trying to do but I do not see it doing any harm specifying Danish as a language for titles in Danish. If the system is already in Danish it should be a no-op. If we get to translating the system to English or German then it will provide value to have Danish titles read out loud in Danish.

As I see it this will make the code easier to understand. This is especially important as the function stands without further documentation or reasoning.

src/apps/material/helper.ts
```diff
@@ -107,6 +107,23 @@ export const getManifestationLanguages = (manifestation: Manifestation) => {
   );
 };
 
+export const getManifestationLanguageIsoCode = (
+  manifestation: Manifestation[]
+) => {
+  const uniqueLanguages = uniqBy(
+    manifestation.map((m) => m.languages),
```
kasperg, line 114:
```suggestion
  manifestations: Manifestation[]
) => {
  const uniqueLanguages = uniqBy(
    manifestations.map((m) => m.languages),
```

It is a bit weird that we have one helper function which accepts an array of manifestations while all the others only accept a single one. Anyhow: Here it makes sense.

kasperg (changes requested):
:+1: good start but there are a few things that needs looking into.

Also I think there are more places where we show material titles that are part of the search project that we need to address. One place that comes to mind is the material suggestions in the autocomplete. There may be others. Please remember to update those as well.

spaceo:
LGTM 👍

sinejespersen (approved):

spaceo (approved):

kasperg (approved):

PR #412, opened 2023-03-31
Fix instant loan

kasperbirch1:
#### This PR are fixing 2 bugs
- Add array length check in `getInstantLoanBranchHoldingsAboveThreshold``
- Change 'matchString' to 'match_string' in `instantLoanConfig`

#### Checklist

### DIFF
--- a/src/components/reservation/helper.ts
+++ b/src/components/reservation/helper.ts
@@ -211,9 +211,14 @@ export const getInstantLoanBranchHoldings = (
 export const getInstantLoanBranchHoldingsAboveThreshold = (
   instantLoanBranchHoldings: HoldingsV3[],
   instantLoanThresholdConfig: string
-) =>
-  instantLoanBranchHoldings.filter(
-    ({ materials }) => materials.length >= Number(instantLoanThresholdConfig)
-  );
+) => {
+  const instantLoanBranchHoldingsAboveThreshold =
+    instantLoanBranchHoldings.filter(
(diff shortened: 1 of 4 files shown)


src/components/reservation/helper.ts
```diff
@@ -211,9 +211,14 @@ export const getInstantLoanBranchHoldings = (
...
-  );
+) => {
+  const instantLoanBranchHoldingsAboveThreshold =
+    instantLoanBranchHoldings.filter(
+      ({ materials }) => materials.length >= Number(instantLoanThresholdConfig)
+    );
+  return instantLoanBranchHoldingsAboveThreshold.length > 0
+    ? instantLoanBranchHoldingsAboveThreshold
+    : null;
+};
```
kasperg, line 222: I do not understand this change.

The function `getInstantLoanBranchHoldingsAboveThreshold()` returns an array. Why can this array not be empty?

src/core/utils/types/instant-loan.ts
```diff
@@ -1,5 +1,5 @@
 export type InstantLoanConfigType = {
-  matchString: string;
+  match_string: string;
```
kasperg, line 2: From a code style perspective we need to decide on what formatting we want here. Personally I am fine with camelCase.

This change needs to be aligned with danskernesdigitalebibliotek/dpl-cms#290.

kasperg (approved):
:+1: LGTM

sinejespersen (approved):

PR #413, opened 2023-03-31
ITKDEV - Feature/cmn 583 make urls work modal

sinejespersen:
#### Link to issue

n/a

#### Description

- This will update the url with appropriate ids
- remove unused code from intermediate list
- readd test
- skip test
- fix blocked patron bad setstate

#### Screenshot of the result

#### Checklist


yepzdk (approved):

src/apps/material/login-redirect.test.ts
```diff
@@ -22,7 +22,8 @@ describe("Material", () => {
     cy.getBySel("reservation-modal-parallel").should("be.visible");
   });
 
-  it("Shouldn't redirect logged in users", () => {
+  // TODO: this never worked as the modal-url-change never worked.
+  it.skip("Shouldn't redirect logged in users", () => {
```
spaceo, line 26: @sinejespersen We are a bit puzzled about this change because it has worked for us. Is it maybe connected to the changes that you have made to the modal slice in this PR?

spaceo: I'll have a talk with Adam about what the purpose of this test is and maybe it should be documented with some comments. I do not think it is self explanatory

spaceo: We have talked about it now :)
So I think indeed that we should document what is happening for users when they are anonymous and when they are authenticated and clicking on the reserve button.
But we dont think it makes sense to skip this test because it should work as is.

When the user is anonymous (in the preceding test) and they click the reserve button following happens:
* The user is redirected to the adgangsplatform with the reservation modal id in the return url
* The user gets logged in and the user is redirected to the return url
Because the modal id is in the redirect url, the user gets the reservation modal presented and everything is fine and dandy.

The idea of the test that you have skipped is:
* The user is authenticated
* The user clicks the reserve button
Because the user is logged in the user is not redirected and thus the url does not change AKA: the modal id is NOT present in the url.

spaceo (changes requested):
Look good. However we need to clarify the matter around the skipped test

spaceo (changes requested):
It is not the right approach to delete the tests because they break.
The result is that we get less code coverage.
If tests breaks when functionality is changed then the tests needs to be updated.
The tests are still relevant:
* Test that anonymous users gets redirected to adgansplatformen and back to material page with open reservationmodal
* Test that inlogged users can open up the reservation modal

So please reintroduce the test and change them so they work in the new setup. If you want we can help you with suggestion but the responsibility for updating the tests are on your side.

sinejespersen:
> It is not the right approach to delete the tests because they break. The result is that we get less code coverage. If tests breaks when functionality is changed then the tests needs to be updated. The tests are still relevant:
> 
> * Test that anonymous users gets redirected to adgansplatformen and back to material page with open reservationmodal
> * Test that inlogged users can open up the reservation modal
> 
> So please reintroduce the test and change them so they work in the new setup. If you want we can help you with suggestion but the responsibility for updating the tests are on your side.

I've talked to Adam who said I should just delete them. As I did. I do not understand the flow well enough to rewrite the tests, and seeing as they were never correctly written, I don't see this as my task.

spaceo:
> > It is not the right approach to delete the tests because they break. The result is that we get less code coverage. If tests breaks when functionality is changed then the tests needs to be updated. The tests are still relevant:
> > 
> > * Test that anonymous users gets redirected to adgansplatformen and back to material page with open reservationmodal
> > * Test that inlogged users can open up the reservation modal
> > 
> > So please reintroduce the test and change them so they work in the new setup. If you want we can help you with suggestion but the responsibility for updating the tests are on your side.
> 
> I've talked to Adam who said I should just delete them. As I did. I do not understand the flow well enough to rewrite the tests, and seeing as they were never correctly written, I don't see this as my task.

We have different oppinons here.
Sorry but I cannot accept this PR with deleted tests

spaceo (changes requested):
Sorry but I cannot accept this PR with deleted tests

Adamik10 (changes requested):
I'm sorry Sine, but we've talked about this in the team and since these changes alter the application in a way that directly affects the tests, do agree with Mikkel that it should be a part of the task to fix them so that we don't lose test coverage. When the two of us were having a conversation about this I didn't properly see all the ways this could be tackled. Thank you in advance for understanding.

spaceo (approved):
LGTM

Adamik10 (approved):
Haven't re-reviewed. I trust @spaceo. Just approving this so that it can be merged.

PR #416, opened 2023-04-13
Improve keyboard navigation in `SearchResultList`

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFSOEG-420

#### This pull request improves keyboard navigation in the `SearchResultList`. 

These changes ensure that the focus moves as expected when navigating through the list using the keyboard to move to the next pages.

The following changes have been made:

- Add 'page' and 'pageSize' props to the `SearchResultListProps` interface.
- Add a useEffect hook to focus on the first item of the new page when the page number changes.
- add a ref attribute to the first li element in the SearchResultList component.

#### Screenshot of the result
https://user-images.githubusercontent.com/49920322/231752771-0cf16655-666a-460a-a45d-8535d91fd2d1.mov

#### Checklist

### DIFF
--- a/src/components/search-result-list/SearchResultList.tsx
+++ b/src/components/search-result-list/SearchResultList.tsx
@@ -43,8 +43,9 @@ const SearchResultList: React.FC<SearchResultListProps> = ({
             <li
               key={item.workId}
               ref={isFirstNewItem ? lastItemRef : null}
-              // Because we're using a ref to focus the first item in the new page when pagination occurs.
-              // we need to remove focus ( set tabIndex -1), so that it can be set programmatically.
+              // We use a ref to focus the first item in the new page programmatically when pagination occurs.
+              // Set tabIndex -1 to support this without allowing keyboard focus. We have just as appropriate
+              // elements within the item suitable for keyboard focus.
               tabIndex={-1}
             >
               <SearchResultListItem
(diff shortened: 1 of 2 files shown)


src/components/search-result-list/SearchResultList.tsx
```diff
@@ -24,15 +37,24 @@ const SearchResultList: React.FC<SearchResultListProps> = ({ resultItems }) => {
...
-        ))}
+        resultItems.map((item, i) => {
+          const isFirstNewItem = i === page * pageSize;
+          return (
+            <li
+              key={item.workId}
+              ref={isFirstNewItem ? lastItemRef : null}
+              // Because we're using a ref to focus the first item in the new page when pagination occurs.
+              // we need to remove focus ( set tabIndex -1), so that it can be set programmatically.
+              tabIndex={-1}
```
kasperg, line 48:
```suggestion
              // We use a ref to focus the first item in the new page programmatically when pagination occurs.
              // Set tabIndex -1 to support this without allowing keyboard focus. We have just as appropriate 
              // elements within the item suitable for keyboard focus.
```

I tried to reword this a bit. Use it if it makes sense to you.

kasperg (approved):
:+1: A few items for your consideration.  Otherwise LGTM.

spaceo (approved):

PR #418, opened 2023-04-14
Remove unnecessary use of `filterCreators`

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFSOEG-503

#### Description

This Pull Request refactors the way we retrieve authors/creators from works and manifestations.

Previously, we looked at whether the type of Creators was a `person` or `Corporation`, but we also want to display the author if the material is, for example, a CD, such as `work-of:870970-basis:23157381`.

Therefore, I have removed the usage of filterCreators and deleted the function.

#### Screenshot of the result
image
image
image
image

#### Checklist


kasperg (approved):
:+1: LGTM

spaceo (approved):

PR #419, opened 2023-04-18
Check if the material has `subjects` before showing "Tags"

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFSOEG-509

#### Description - Check if the material has `subjects` before showing "Tags"

We only want to display "Tags" / "Emneord" / "subjectsList" if there are some items in the `subjects` array. 
This is achieved by checking subjectsList.length before rendering the HorizontalTermLine component.

#### Screenshot of the result
image

#### Checklist

### DIFF
--- a/src/components/material/MaterialDescription.tsx
+++ b/src/components/material/MaterialDescription.tsx
@@ -87,7 +87,7 @@ const MaterialDescription: React.FC<MaterialDescriptionProps> = ({ work }) => {
                 dataCy="material-description-series-members"
               />
             )}
-            {subjectsList && (
+            {subjectsList.length > 0 && (
               <HorizontalTermLine
                 title={t("identifierText")}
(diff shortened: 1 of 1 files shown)


src/components/material/MaterialDescription.tsx
```diff
@@ -87,7 +87,7 @@ const MaterialDescription: React.FC<MaterialDescriptionProps> = ({ work }) => {
                 dataCy="material-description-series-members"
               />
             )}
-            {subjectsList && (
+            {subjectsList.length > 0 && (
```
kasperg, line 90: This is the third time we have to implement this check. Have you considered moving the length check into the `HorizontalTermLine` component?

kasperg (approved):
:+1: LGTM

spaceo (approved):

PR #420, opened 2023-04-18
Add `allowOutsideClick` to `FocusTrap` in ReservationSucces/ Error modal

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFSOEG-506

####  This PR fixes the issue of not being able to close the ReservationSuccess or ReservationError modal by pressing the cross/close button.

The default behavior is to not allow clicking outside the `FocusTrap`. However, we still want to be able to close the modal with the mouse, so we set allowOutsideClick to `true`.

#### Checklist


kasperg (approved):
:+1: LGTM

spaceo (approved):

PR #422, opened 2023-04-18
Add `HorizontalTermLine` for `dk5MainEntry` ("emnetal") to nonFiction Works

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFSOEG-485

#### Description

This PR introduces the `HorizontalTermLine` component for `dk5MainEntry` ("emnetal")  specifically for nonFiction Works

#### Screenshot of the result

image

#### Checklist

### DIFF
--- a/src/components/material/MaterialDescription.tsx
+++ b/src/components/material/MaterialDescription.tsx
@@ -14,6 +14,7 @@ import { Work } from "../../core/utils/types/entities";
 import { Pid, WorkId } from "../../core/utils/types/ids";
 import { useUrls } from "../../core/utils/url";
 import HorizontalTermLine from "../horizontal-term-line/HorizontalTermLine";
+import { materialIsFiction } from "../../core/utils/helpers/general";
 
 export interface MaterialDescriptionProps {
   pid: Pid;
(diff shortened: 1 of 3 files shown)


src/components/material/MaterialDescription.tsx
```diff
@@ -65,6 +73,17 @@ const MaterialDescription: React.FC<MaterialDescriptionProps> = ({ work }) => {
             </p>
           )}
           <div className="material-description__links mt-32">
+            {!isFiction && dk5MainEntry && (
+              <HorizontalTermLine
+                title="Emnetal"
```
kasperg, line 78: Please ensure that static texts like this are translatable with `t()`.

src/components/material/MaterialDescription.tsx
```diff
@@ -65,6 +73,17 @@ const MaterialDescription: React.FC<MaterialDescriptionProps> = ({ work }) => {
             </p>
           )}
           <div className="material-description__links mt-32">
+            {!isFiction && dk5MainEntry && (
+              <HorizontalTermLine
```
kasperg, line 77: I wonder about the ordering of this element in relation to the order horizontal term lines.

This adds it as the first form line. I had expected expected it to be in relation to the `fictionNonfiction` one as it is relevant to non-fiction materials. If the current position has been cleared with the business then OK.

kasperg (approved):
:+1: but please remember to make texts translatable and add a corresponding control in the story before passing this along.

spaceo (approved):

PR #425, opened 2023-04-21
remove put and save from tests

sinejespersen:
#### Link to issue

n/a

#### Description

Remove save/put from tests and readd patron test

### DIFF
--- a/src/apps/reservation-list/modal/reservation-details/reservation-details-redirect.tsx
+++ b/src/apps/reservation-list/modal/reservation-details/reservation-details-redirect.tsx
@@ -29,7 +29,6 @@ const ReservationDetailsRedirect: FC<
     <div className={`modal-details__buttons ${className}`}>
       <button
         type="button"
-        data-cy="delete-reservation-button"
         onClick={() => openReservationDeleteModal(reservation)}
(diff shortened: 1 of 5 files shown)


src/apps/reservation-list/modal/reservation-details/reservation-details-redirect.tsx
```diff
@@ -29,6 +29,7 @@ const ReservationDetailsRedirect: FC<
     <div className={`modal-details__buttons ${className}`}>
       <button
         type="button"
+        data-cy="delete-reservation"
```
yepzdk, line 32: `dataCy` is used in [reservation-details-buttons.tsx](https://github.com/danskernesdigitalebibliotek/dpl-react/pull/425/files#diff-178291d6591a1bad453329a81156a69812e7e3acb9a685b2fa46a9a8a15656e3)
should `data-cy` be the same?

yepzdk (approved):
Approved with one minor comment.

spaceo (commented):
There is no context or explanation of what and why is happening here which makes reviewing a bit difficult :)
What I can see is that a lot of clicks have been removed from tests, which I consider being a degradation of the tests.
Why are the clicks removed from the tests?

PR #427, opened 2023-04-21
Use 'workYear' from 'work' instead of manifestation in `MaterialDetailsList`

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFSOEG-496

#### Description
This PR fixes the issue of missing the first edition year, which was caused by using 'workYear' on the selected / 'bestRepresentation' manifestation level where it was only provided on some of the manifestations. 

The following changes have been made to address this issue:

- Refactor `getManifestationFirstEditionYear` to `getWorkFirstEditionYear` as we should now use 'workYear' on the work level, rather than the manifestation level.
- Update the details list to display the first edition year using `getWorkFirstEditionYear` 

#### Screenshot of the result
Skærmbillede 2023-04-21 kl  15 14 11 (3)

#### Checklist

### DIFF
--- a/src/apps/material/helper.ts
+++ b/src/apps/material/helper.ts
@@ -135,12 +135,8 @@ export const getManifestationLanguageIsoCode = (
   return undefined;
 };
 
-export const getManifestationFirstEditionYear = (
-  manifestation: Manifestation
-) => {
-  return manifestation.workYear?.year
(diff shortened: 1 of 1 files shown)


Adamik10 (approved):
Perfect :) Approved

spaceo (approved):

PR #431, opened 2023-04-27
Prevent empty search on 'Enter' press in search input

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFSOEG-461

#### Description

This pull request implements a change that prevents users from executing an empty search when pressing the 'Enter' key in the search input field or the search icon is clicked. The update ensures that a search is only triggered when there is a meaningful input value, not just whitespace characters.

#### Screenshot of the result

https://user-images.githubusercontent.com/49920322/234840414-3fb42b6a-e8be-4d26-8793-faa524f477cd.mov

#### Checklist

### DIFF
--- a/src/components/search-bar/search-bar.tsx
+++ b/src/components/search-bar/search-bar.tsx
@@ -26,6 +26,9 @@ const SearchBar: React.FC<SearchBarProps> = ({
       </label>
       {/* eslint-disable react/jsx-props-no-spreading */}
       <input
+        required
+        pattern=".*\S+.*"
+        title={t("searchNoValidCharactersErrorText")}
         name="q"
         className="header__menu-search-input text-body-medium-regular"
         data-cy={dataCy}
@@ -41,6 +44,7 @@ const SearchBar: React.FC<SearchBarProps> = ({
       />
       {/* eslint-enable react/jsx-props-no-spreading */}
       <input
+        required
         type="image"
         src={searchIcon}
         alt={t("searchHeaderIconAltText")}
--- a/src/apps/search-header/search-header.dev.tsx
+++ b/src/apps/search-header/search-header.dev.tsx
@@ -105,6 +105,11 @@ export default {
       name: "Loading",
       defaultValue: "Loading",
       control: { type: "text" }
+    },
+    searchNoValidCharactersErrorText: {
+      name: "Search non-whitespace character error",
+      defaultValue: "Input must contain at least one non-whitespace character.",
+      control: { type: "text" }
     }
   }
 } as ComponentMeta<typeof SearchHeaderEntry>;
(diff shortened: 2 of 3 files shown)


src/components/search-bar/search-bar.tsx
```diff
@@ -24,6 +24,9 @@ const SearchBar: React.FC<SearchBarProps> = ({
       </label>
       {/* eslint-disable react/jsx-props-no-spreading */}
       <input
+        required
+        pattern=".*\S+.*"
+        title={t("searchNoValidCharactersErrorText")}
```
Adamik10, line 29: Would it make sense to show a different title if they already typed something in and it's correct? Just a consideration, I think this is also okay 😉

Adamik10 (approved):
Good stuff :) I left a single comment for consideration.

src/components/search-bar/search-bar.tsx
```diff
@@ -24,6 +24,9 @@ const SearchBar: React.FC<SearchBarProps> = ({
       </label>
       {/* eslint-disable react/jsx-props-no-spreading */}
       <input
+        required
+        pattern=".*\S+.*"
```
spaceo, line 28: Is the pattern necessary?

src/components/search-bar/search-bar.tsx
```diff
@@ -35,6 +38,9 @@ const SearchBar: React.FC<SearchBarProps> = ({
       />
       {/* eslint-enable react/jsx-props-no-spreading */}
       <input
+        required
+        pattern=".*\S+.*"
```
spaceo, line 42: Pretty sure it makes no sense to have a validation pattern on a button :)

spaceo (changes requested):
Needs to be looked at again.
I think you have to read about html5 input elements and validation

src/components/search-bar/search-bar.tsx
kasperbirch1: If it is not there, it is possible to make a search with whitespace characters.

src/components/search-bar/search-bar.tsx
kasperbirch1: "Use the [title](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#title) attribute to specify text that most browsers will display as a tooltip to explain what the requirements are to match the pattern."

https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/pattern

src/components/search-bar/search-bar.tsx
spaceo: 👍

spaceo (approved):
LGTM

sinejespersen (approved):

PR #433, opened 2023-05-01
Fixing parallel reservation issues with blacklisting 

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFSOEG-499

#### Description

This pull request aims to fix issues related to parallel reservations by addressing two specific problems with the `work-of:870970-basis:46143736`. Firstly, the branch `DK-775164` (Egå skolebibliotek) needs to be blacklisted, and secondly, we need to include the blacklist in the `UseReservableManifestations`.

To address these issues, this pull request introduces new hooks - `useGetAvailability`, `getAvailability`, and `useGetAvailability` - which serve as a wrapper around existing hooks. These new hooks handle cases where 'blacklistedConfig' needs to be considered while fetching data. By using the `getBlacklistedArgs` function, the wrapper ensures that blacklisted branches are appropriately excluded from the query results.

While working on this task, I realized that we are unable to manage situations where a parallel reservation includes an item that has already been reserved. As a result, I modified the `handleErrorText` function to handle an array of strings instead of just a single string."

#### Checklist

### DIFF
--- a/src/apps/material/helper.ts
+++ b/src/apps/material/helper.ts
@@ -9,7 +9,6 @@ import {
 import { ManifestationHoldings } from "../../components/find-on-shelf/types";
 import { ListData } from "../../components/material/MaterialDetailsList";
 import {
-  AvailabilityV3,
   HoldingsForBibliographicalRecordV3,
   HoldingsV3
 } from "../../core/fbs/model";
@@ -26,7 +25,12 @@ import {
   isArticle
 } from "../../components/material/material-buttons/helper";
 import { UseConfigFunction } from "../../core/utils/config";
-import { getAvailabilityV3, useGetAvailabilityV3 } from "../../core/fbs/fbs";
+import {
+  getAvailabilityV3,
+  getHoldingsV3,
+  useGetAvailabilityV3,
+  useGetHoldingsV3
+} from "../../core/fbs/fbs";
 
 export const getWorkManifestation = (
   work: Work,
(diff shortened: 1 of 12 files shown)


src/apps/material/helper.ts
```diff
@@ -402,3 +410,75 @@ export const isParallelReservation = (manifestations: Manifestation[]) =>
...
+  const blacklistBranches = config(blacklist, {
+    transformer: "stringToArray"
+  });
+  return {
+    recordid: faustIds,
+    ...(blacklistBranches ? { exclude: blacklistBranches } : {})
+  };
+};
+
+export const useGetAvailability = ({
```
Adamik10, line 431: I'm thinking - should we place these "use..xyz" hooks from here to be in separate files like UseReservableManifestations.tsx is?

Adamik10 (approved):
I see no glaring issues with this. Although, consider moving the new hooks into their separate files as we do with others. Good job.

src/apps/material/helper.ts
spaceo, line 431: Nice initiative. You could also consider moving the functionality around availability to a separate file in core somewhere. core/utils maybe?
Since it is more global than just the material app in my opinion.

src/apps/material/helper.ts
```diff
@@ -402,3 +410,75 @@ export const isParallelReservation = (manifestations: Manifestation[]) =>
   manifestations.length > 1 &&
   hasCorrectAccessType(AccessTypeCode.Physical, manifestations) &&
   !isArticle(manifestations);
+
+// Because we  need to exclude the branches that are blacklisted, we need to use a custom hook to prevent duplicate code
+const getBlacklistedArgs = (
```
spaceo, line 415: I think it should be called `getBlacklistedQueryArgs` to make it clear what it is being used for. I needed some extra seconds to decipher the meaning of it.

src/apps/material/helper.ts
```diff
@@ -402,3 +410,75 @@ export const isParallelReservation = (manifestations: Manifestation[]) =>
...
+  faustIds: FaustId[],
+  config: UseConfigFunction,
+  blacklist:
+    | "blacklistedAvailabilityBranchesConfig"
+    | "blacklistedPickupBranchesConfig"
```
spaceo, line 420: I dont think we need to use the full config prop names here. I think it could be more need to just make a union of some simple keys like: `"availability" | "pickupBranches" `.
Then you can construct the config names like they should be inside this function either with a map or a pattern. does it make sense?

spaceo (commented):
I have some suggestions for changes

spaceo (commented):
Looks good to me. But you have one warning:
image

sinejespersen (approved):

PR #423, opened 2023-04-19
  Don't open ReservationModal modal if the user is blocked.

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFSOEG-492

#### This pull request addresses the issue of the `ReservationModal` being opened for blocked users. 

Previously, when a user who was not logged in clicked on the Reserve button, they were sent to the login platform and then redirected back to the page with a URL containing `&modal=reservation-modal-********`. This opened the modal, even if the user was blocked.

To resolve this issue, I have implemented the following changes:

- Added a check for `isUserAllowed` before rendering the `ReservationModal` and `FindOnShelfModal`.
- Introduced `userIsBlocked` and `userIsAllowed` variables to improve code readability and maintainability.

These changes ensure that the `ReservationModal` is not opened for blocked users.

#### Screenshot of the result
**Blocked user:**

https://user-images.githubusercontent.com/49920322/233066134-ef5c558a-c76b-4f44-9f86-78c10b6730b6.mov

**user:**

https://user-images.githubusercontent.com/49920322/233066263-f1ee63cf-cb6d-4236-8b13-13814c936cca.mov

#### Checklist

- [ ] My complies with [our coding guidelines](../docs/code_guidelines.md).
- [ ] My code passes our static analysis suite. If not then I have added a comment explaining why this change should be exempt from the code standards and process.
- [ ] My code passes our continuous integration process. If not then I have added a comment explaining why this change should be exempt from the code standards and process.

#### Additional comments or questions

PR #435, opened 2023-05-02
Fix `MaterialAvailabilityText`

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFSOEG-511
https://reload.atlassian.net/browse/DDFSOEG-512

#### Description
This PR addresses the issues with the information text for online materials, ensuring that it displays the limits for the maximum number of e- and audiobook reservations and loans per user. 

Additionally, it adds a check so the text is not rendered when the user is not logged in.

#### Screenshot of the result
image
image

#### Checklist

I needed to split `MaterialType` because 'MaterialType' is specified as a string from the API, we need to have two different versions, as for example, ebook in the 'facet' case is `e-bog` and in 'manifestation' `ebog`.

#### Be aware of:
 `userIsAnonymous` will change to `isAnonymous`
https://github.com/danskernesdigitalebibliotek/dpl-react/pull/423


src/components/material/MaterialAvailabilityText/online/MaterialAvailabilityTextOnline.tsx
```diff
@@ -1,27 +1,60 @@
...
-    placeholders: { "@count": "X", "@limit": "Y" }
-  });
+  if (
+    materialType === ManifestationMaterialType.ebook &&
+    maxConcurrentEbookReservationsPerBorrower &&
+    maxConcurrentEbookLoansPerBorrower
+  ) {
+    availabilityText = t("onlineLimitMonthEbookInfoText", {
+      placeholders: {
+        "@count": maxConcurrentEbookReservationsPerBorrower,
```
Adamik10, line 40: Are we sure this is correct? here we want to show how many ebooks the user has already loaned this month, but according to this variable's name - maxConcurrentEbookReservationsPerBorrower - it is the maximum a singular person can reserve (per month I assume).

Adamik10: And I also missed this, but concurrent reservations would mean reservations happening at the same time. So it doesn't really seem to me like it's the monthly quota in either of these variables.

src/components/material/MaterialAvailabilityText/online/MaterialAvailabilityTextOnline.tsx
```diff
@@ -1,27 +1,60 @@
...
+  }
+
+  if (
+    materialType === ManifestationMaterialType.audioBook &&
+    maxConcurrentAudioReservationsPerBorrower &&
+    maxConcurrentAudioLoansPerBorrower
+  ) {
+    availabilityText = t("onlineLimitMonthAudiobookInfoText", {
+      placeholders: {
+        "@count": maxConcurrentAudioReservationsPerBorrower,
```
Adamik10, line 53: Same as the comment above.

src/components/material/MaterialHeader.tsx
```diff
@@ -153,9 +154,11 @@ const MaterialHeader: React.FC<MaterialHeaderProps> = ({
                     materialTitleId={materialTitleId}
                   />
                 </div>
-                <MaterialAvailabilityText
-                  manifestations={selectedManifestations}
-                />
+                {!userIsAnonymous() && (
```
Adamik10, line 157: This will have to be changed to the new "isUserAnonymous()" right?

Adamik10 (changes requested):
I am a bit unsure as to whether we have the right numbers set in the right places. Please read my comments and let's figure this out! :)

src/components/material/MaterialHeader.tsx
kasperbirch1: Yes when https://github.com/danskernesdigitalebibliotek/dpl-react/pull/423 is merged

src/components/material/MaterialAvailabilityText/online/MaterialAvailabilityTextOnline.tsx
```diff
@@ -1,27 +1,63 @@
...
+      }
+    });
+  }
+
+  if (
+    materialType === ManifestationMaterialType.audioBook &&
+    totalAudioLoans &&
+    maxConcurrentAudioLoansPerBorrower
+  ) {
+    availabilityText = t("onlineLimitMonthAudiobookInfoText", {
```
spaceo, line 54: Suggestion for a more function based and, in my opinion, more maintainable solution:
```typescript

type AvailabilityTextMap = {
  [key in ManifestationMaterialType]: {
    text: string;
    count: number;
    limit: number;
  };
};

const getAvaliabilityText = (
  type: ManifestationMaterialType,
  map: AvailabilityTextMap,
  t: UseTextFunction
) => {
  if (!map[type] || !map[type].text || !map[type].count || !map[type].limit) {
    return null;
  }

  const { text, count, limit } = map[type];

  return t(text, {
    placeholders: {
      '@count': count,
      '@limit': limit,
    },
  });
};

export default function App() {
  const t = useText();
  const availabilityTextMap: AvailabilityTextMap = {
    [ManifestationMaterialType.ebook]: {
      text: "onlineLimitMonthEbookInfoText",
      count: totalEbookLoans,
      limit: maxConcurrentEbookLoansPerBorrower,
    },
    [ManifestationMaterialType.audioBook]: {
      text: "onlineLimitMonthAudiobookInfoText",
      count: totalAudioLoans,
      limit: maxConcurrentAudioLoansPerBorrower,
    },
  };
  const text = getAvaliabilityText(
    ManifestationMaterialType.ebook,
    availabilityTextMap,
    t
  );
```

spaceo (commented):
Please look at my suggestion for getting the availability text

src/components/material/MaterialAvailabilityText/MaterialAvailabilityText.tsx
```diff
@@ -12,19 +16,28 @@ interface Props {
...
+
+  if (
+    hasCorrectAccessType(AccessTypeCode.Online, manifestations) &&
+    getAllIdentifiers(manifestations).length > 0 &&
+    materialType
+  ) {
+    const isbns = getAllIdentifiers(manifestations);
+    return (
+      <MaterialAvailabilityTextOnline
+        isbns={isbns}
```
spaceo, line 34: Since isbns is only used for this particular case you can also skip creating a variable that has no particular value other than occuping a tiny bit of memory and instead use the function directly:
```typescript
      <MaterialAvailabilityTextOnline
        isbns={getAllIdentifiers(manifestations)}

File: src/components/material/MaterialAvailabilityText/online/helper.ts [hunk 6]
```diff
@@ -0,0 +1,42 @@
...
+    return t(text, {
+      placeholders: {
+        "@count": count,
+        "@limit": limit
+      }
+    });
+  }
+
+  return null;
+};
```
spaceo, line 40: Look great. I have come up with a change that is a tad more extensible in the future:
```typescript
export const getAvailabilityText = ({
  type,
  map,
  t,
}: {
  type: ManifestationMaterialType | 'materialIsIncluded';
  map: AvailabilityTextMap;
  t: UseTextFunction;
}) => {
  if (!map || !map[type]) return null;

  const { text, count, limit } = map[type];

  if (count === undefined && limit === undefined && text) {
    return t(text);
  }

  if (count !== undefined && limit !== undefined && text !== undefined) {
    return t(text, {
      placeholders: {
        '@count': count,
        '@limit': limit,
      },
    });
  }

  return null;
};

spaceo (changes requested):
Looks great. I have a few more additions though

spaceo (approved):
LGTM

sinejespersen (approved):

Adamik10 (approved):
LGTM

PR #441, opened 2023-05-03
Fix infomedia access

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFSOEG-497

#### Description
This PR fixes an issue where anyone logged in could view infomedia articles. The desired functionality is that it is only available to be viewed for users that are residents in the library's municipality. After introducing these changes, some tests had to be adjusted.
As a part of this PR, various other code optimizations were included as well. The one prominent one would be the introduction of `ReservationFindOnShelfModals` - a component rendering a combination of a reservation modal and a find on shelf modal. The reason behind this addition is that all users should have access to find on shelf modal, but only logged in + not blocked users should have access to the reservation modal. This combination is seen twice on the material page and the access check wasn't the same. Now there exists only a singular place that decides viewing rights.

#### Screenshot of the result
image

#### Checklist

n/a


spaceo:
material, l41:
we should delete userCanReserve because it is the opposite of isBlocked
material, l61:
canUserReserve should be boolean (but it is going to be deleted anyway :))

material, l166-l180:
maybe we can make a component that wraps ReservationModal and FindOnShelfModal
so will not have to repeat logic in l198-211.
It will also solve the problem that in line 195 the block hides both modals for blocked users where it should only hide the reservation modal.

https://docs.google.com/spreadsheets/d/1zmcl0VIHTsiRmol2-IMthQnnuCsc3UKP0xvNPn5Dz2w/edit#gid=0

spaceo:
material, l41:
we should delete userCanReserve because it is the opposite of isBlocked
material, l61:
canUserReserve should be boolean (but it is going to be deleted anyway :))

material, l166-l180:
maybe we can make a component that wraps ReservationModal and FindOnShelfModal
so will not have to repeat logic in l198-211.
It will also solve the problem that in line 195 the block hides both modals for blocked users where it should only hide the reservation modal.

https://docs.google.com/spreadsheets/d/1zmcl0VIHTsiRmol2-IMthQnnuCsc3UKP0xvNPn5Dz2w/edit#gid=0

spaceo:
material, l41:
we should delete userCanReserve because it is the opposite of isBlocked
material, l61:
canUserReserve should be boolean (but it is going to be deleted anyway :))

material, l166-l180:
maybe we can make a component that wraps ReservationModal and FindOnShelfModal
so will not have to repeat logic in l198-211.
It will also solve the problem that in line 195 the block hides both modals for blocked users where it should only hide the reservation modal.

https://docs.google.com/spreadsheets/d/1zmcl0VIHTsiRmol2-IMthQnnuCsc3UKP0xvNPn5Dz2w/edit#gid=0

spaceo:
material, l41:
we should delete userCanReserve because it is the opposite of isBlocked
material, l61:
canUserReserve should be boolean (but it is going to be deleted anyway :))

material, l166-l180:
maybe we can make a component that wraps ReservationModal and FindOnShelfModal
so will not have to repeat logic in l198-211.
It will also solve the problem that in line 195 the block hides both modals for blocked users where it should only hide the reservation modal.

https://docs.google.com/spreadsheets/d/1zmcl0VIHTsiRmol2-IMthQnnuCsc3UKP0xvNPn5Dz2w/edit#gid=0

spaceo:
material, l41:
we should delete userCanReserve because it is the opposite of isBlocked
material, l61:
canUserReserve should be boolean (but it is going to be deleted anyway :))

material, l166-l180:
maybe we can make a component that wraps ReservationModal and FindOnShelfModal
so will not have to repeat logic in l198-211.
It will also solve the problem that in line 195 the block hides both modals for blocked users where it should only hide the reservation modal.

https://docs.google.com/spreadsheets/d/1zmcl0VIHTsiRmol2-IMthQnnuCsc3UKP0xvNPn5Dz2w/edit#gid=0

spaceo:
material, l41:
we should delete userCanReserve because it is the opposite of isBlocked
material, l61:
canUserReserve should be boolean (but it is going to be deleted anyway :))

material, l166-l180:
maybe we can make a component that wraps ReservationModal and FindOnShelfModal
so will not have to repeat logic in l198-211.
It will also solve the problem that in line 195 the block hides both modals for blocked users where it should only hide the reservation modal.

https://docs.google.com/spreadsheets/d/1zmcl0VIHTsiRmol2-IMthQnnuCsc3UKP0xvNPn5Dz2w/edit#gid=0

spaceo (approved):
LGTM 👍

PR #446, opened 2023-05-08
Add possibility to filter on `workTypes` in `AutosuggestCategory`

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFSOEG-326

#### Description
This pull request introduces the possibility to filter on `workTypes` in `AutosuggestCategory`, which was previously limited to only `materialTypes`. 

This change will provide users with a more flexible and powerful way to search and filter results.

#### Checklist

It also resolves the absent mouse functionality in `handleHighlightedIndexChange`, enabling users to use their mouse for navigating to the search page with filters within the `AutosuggestCategory`.

### DIFF
--- a/src/apps/search-header/search-header.tsx
+++ b/src/apps/search-header/search-header.tsx
@@ -17,7 +17,10 @@ import {
 } from "../../core/utils/helpers/url";
 import { WorkId } from "../../core/utils/types/ids";
 import { useText } from "../../core/utils/text";
-import { FacetMaterialType } from "../../core/utils/types/material-type";
+import {
+  AutosuggestCategoryList,
+  AutosuggestCategory
+} from "../../core/utils/types/material-type";
 import { findNonWorkSuggestion } from "./helpers";
 import { useStatistics } from "../../core/statistics/useStatistics";
 import { statistics } from "../../core/statistics/statistics";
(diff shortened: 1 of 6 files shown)


src/core/utils/types/material-type.ts
```diff
@@ -14,20 +14,22 @@ export const enum FacetMaterialType {
   magazine = "tidsskrift"
 }
 
-export const enum ManifestationMaterialType {
+export const enum AutosuggestCategoryType {
```
spaceo, line 17: Could also just be called: "AutosuggestCategory"?

src/core/utils/types/material-type.ts
```diff
@@ -14,20 +14,22 @@ export const enum FacetMaterialType {
...
-  magazine = "tidsskrift"
+  ebook = "e-bog",
+  movie = "Film",
+  audioBook = "lydbog (online)",
+  music = "Musik",
+  game = "Spil",
+  animatedSeries = "tegneserie"
 }
 
+export type AutosuggestCategoryFacetType = "materialTypes" | "workTypes";
```
spaceo, line 27: Could also just be called: "AutosuggestCategoryFacet"?

src/core/utils/types/material-type.ts
```diff
@@ -14,20 +14,22 @@ export const enum FacetMaterialType {
...
+
+export type AutosuggestCategoryListType = {
```
spaceo, line 29: Could also just be called: "AutosuggestCategoryItem"?

src/apps/search-result/useFilterHandler.tsx
```diff
@@ -41,7 +42,23 @@ const useFilterHandler = () => {
     [dispatch]
   );
 
-  return { filters, addToFilter, removeFromFilter, clearFilter };
+  const addFilterFromUrlParam = (facet: FacetField) => {
+    const urlFilter = getUrlQueryParam(facet);
+    if (urlFilter) {
+      addToFilter({
+        facet,
+        term: { key: urlFilter, term: urlFilter }
```
spaceo, line 50: Is it deliberate that you use `urlFilter`both for `key`and for `term`?

src/apps/search-result/search-result.tsx
```diff
@@ -87,13 +87,8 @@ const SearchResult: React.FC<SearchResultProps> = ({ q, pageSize }) => {
...
   // It could be reworked to support all filters and terms at a later point.
   useEffect(() => {
-    const materialTypeUrlFilter = getUrlQueryParam("materialType");
-    if (materialTypeUrlFilter) {
-      addToFilter({
-        facet: FacetField.MaterialTypes,
-        term: { key: materialTypeUrlFilter, term: materialTypeUrlFilter }
-      });
-    }
+    addFilterFromUrlParam(FacetField.MaterialTypes);
```
spaceo, line 90: I think is a bit misconceiving to call it `addFilterFromUrlParam` because I think it implies that you perform the action right away.
Maybe it could be called  something like `addFilterFromUrlParamListener` to emphasise that it only performs an action if an url param exist.

spaceo (commented):
I have some naming suggestions...

Adamik10 (approved):
Looks good to me :) Good job making sense with all the search-header stuff. It's pretty complex.

src/apps/search-header/search-header.tsx
```diff
@@ -161,18 +180,22 @@ const SearchHeader: React.FC = () => {
...
-      type === useCombobox.stateChangeTypes.ItemMouseMove ||
-      type === useCombobox.stateChangeTypes.MenuMouseLeave
-    ) {
+    // Don't do anything for mouse leave events + exit function.
+    if (type === useCombobox.stateChangeTypes.MenuMouseLeave) {
       return;
     }
-    // If this is a click/enter press we need to save the highlighted index
-    // before Downshift sets it to -1.
+    // Set highlighted index when hovering over with mouse + exit function.
```
sinejespersen, line 187: what does this mean?

kasperbirch1:
```jsx
    // This block handles the state change when the user hovers over an item with the mouse.
    // The type 'useCombobox.stateChangeTypes.ItemMouseMove' indicates that the mouse is moving over the items in the dropdown.
    if (type === useCombobox.stateChangeTypes.ItemMouseMove) {
      // If the highlightedIndex (the index of the item under the mouse cursor) is defined and greater than -1,
      // we update the state 'highlightedIndexAfterClick' to this index.
      // This state keeps track of the last highlighted index during a click or mouse move event.
      if (highlightedIndex !== undefined && highlightedIndex > -1) {
        setHighlightedIndexAfterClick(highlightedIndex);
      }
      // We exit the function early after handling a mouse move event,
      // ensuring that the rest of the function (which handles other event types) isn't executed.
      return;
    }

sinejespersen (approved):

PR #447, opened 2023-05-09
DDFSOEG-508 Check for logged-in user before fetching Patron Information

kasperbirch1:
#### Link to issue

#### Description

This pull request adds a check to ensure the user is logged in before attempting to fetch Patron Information. Previously, if there was no user token (i.e., the user was not logged in), the FBS fetcher would send the library token instead. 

This would cause issues when trying to fetch from `https://fbs-openplatform.dbc.dk/external/agencyid/patrons/patronid/v2`, resulting in an error after 3 failed attempts. To resolve this issue, we now check for a valid user before fetching Patron Information.

#### Checklist


kasperbirch1:
See this: https://github.com/danskernesdigitalebibliotek/dpl-react/pull/441

PR #451, opened 2023-05-15
Fix autosuggest middle section

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFSOEG-520

#### Description
This PR solves an issue where autosuggest shows fewer items in the middle section than there is in data. This happened because some of the suggestions doesn't have a work attached to them. We now filter these suggestions out of the material suggestion section.

#### Screenshot of the result
n/a

#### Checklist

n/a


spaceo (approved):
LGTM

PR #452, opened 2023-05-16
Switch to use `clsx` instead of template strings in `Disclosure`

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFSOEG-521

#### Description

This pull request improves the `Disclosure` component's className handling by replacing template strings with the `clsx` library. The previous approach resulted in undefined class names when no class was passed, leading to potential issues and inconsistencies.

#### Screenshot of the result
image

#### Checklist


spaceo (approved):
LGTM

PR #458, opened 2023-05-22
Fix e-resource availability

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFSOEG-517

#### Description
This PR fixes availability exposing for e-resources, and introduces loading state to availability labels. 
The e-resource availability is a bit tricky - they are per default always available, unless they can be accessed from Ereolen and they are not blue titles (unlimited loans) and there is a queue on them.

#### Screenshot of the result
image

#### Checklist

n/a

### DIFF
--- a/src/components/availability-label/AvailabilitySkeleton.tsx
+++ b/src/components/availability-label/AvailabilitySkeleton.tsx
@@ -1,0 +1,8 @@
+import * as React from "react";
+import { FC } from "react";
+
+const AvailabilitySkeleton: FC = () => {
+  return <div className="ssc-line w-100" style={{ width: "40px" }} />;
+};
+
+export default AvailabilitySkeleton;
--- a/src/components/availability-label/AvailabilitySkeleton.tsx
+++ b/src/components/availability-label/AvailabilitySkeleton.tsx
@@ -1,8 +1,8 @@
 import * as React from "react";
 import { FC } from "react";
 
-const AvailabilitySkeleton: FC = () => {
-  return <div className="ssc-line w-100" style={{ width: "40px" }} />;
-};
+const AvailabilitySkeleton: FC = () => (
+  <div className="ssc-line w-100" style={{ width: "40px" }} />
+);
 
 export default AvailabilitySkeleton;
--- a/src/components/availability-label/availability-labels.tsx
+++ b/src/components/availability-label/availability-labels.tsx
@@ -50,6 +50,12 @@ export const AvailabilityLabels: React.FC<AvailabilityLabelsProps> = ({
           })
           .flat();
 
+        const access = manifestationsOfMaterialType
+          .map((manifest) => {
+            return manifest.access.map((acc) => acc.__typename);
+          })
+          .flat();
+
         return (
           <AvailabilityLabel
             key={faustIds.join("-")}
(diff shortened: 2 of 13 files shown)


src/components/availability-label/AvailabilitySkeleton.tsx
```diff
@@ -0,0 +1,8 @@
+import * as React from "react";
+import { FC } from "react";
+
+const AvailabilitySkeleton: FC = () => {
+  return <div className="ssc-line w-100" style={{ width: "40px" }} />;
+};
```
spaceo, line 6:
```suggestion
const AvailabilitySkeleton: FC = () => (
  <div className="ssc-line w-100" style={{ width: "40px" }} />;
);
```

src/components/availability-label/availability-labels.tsx
```diff
@@ -50,6 +50,12 @@ export const AvailabilityLabels: React.FC<AvailabilityLabelsProps> = ({
           })
           .flat();
 
+        const access = manifestationsOfMaterialType
+          .map((manifest) => {
+            return manifest.access.map((acc) => acc.__typename);
+          })
+          .flat();
```
spaceo, line 57:
```suggestion
        const manifestationAccesses = manifestationsOfMaterialType
          .map(({ access }) => access.map((acc) => acc.__typename))
          .flat();
```

src/components/availability-label/helper.ts
```diff
@@ -37,24 +93,13 @@ export const useAvailabilityData = ({
...
-        //    2. If the library has a queue on the material
-        setIsAvailable(true);
-      }
-    }
-  });
+  useEffect(() => {
+    setIsLoading(
+      !!(isLoadingAvailability || isLoadingIdentifier || isLoadingProductInfo)
+    );
+  }, [isLoadingAvailability, isLoadingIdentifier, isLoadingProductInfo]);
```
spaceo, line 100: I like this construction but is it necessary to cast it to a boolean? Isn't the loading status always boolean from the different services?

src/components/availability-label/helper.ts
```diff
@@ -1,25 +1,81 @@
...
+  const [isCostFree, setIsCostFree] = useState<null | boolean>(null);
+  const [isLoading, setIsLoading] = useState<null | boolean>(null);
 
-  useGetAvailability({
+  useEffect(() => {
+    // An online material is by default always available.
+    if (isOnline && isCostFree !== false) {
+      setIsAvailable(true);
+    }
+  }, [isOnline, isCostFree]);
```
spaceo, line 38: Maybe I am wrong but the setup here with two requests and useage of `isCostFree` that is used to control `isAvailable` looks fragile and upredictable.

Does the setup work no matter which request resolves first?

Is it not possible to set availability directlty in the `useGetV1ProductsIdentifier()` onSuccess handler?

It would be nice with a unit test that proves that this mechanism works. But I know that it is extra time/scope

spaceo (commented):
Nice work ;)
I have some questions I would like to clarify.

src/components/availability-label/helper.ts
spaceo: Let's talk about it together. I have an idea that this could be more robust with a sequential handling of the requests with promises or await.

spaceo (approved):
LGTM 👍

sinejespersen (approved):

PR #459, opened 2023-05-23
Refactor `login-redirect.test`

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFSOEG-526

#### Description
This PR reimplements the `login-redirect.test` that was missing after changes in the Modal component.

The way we can test if an unauthorized user is redirected to the login platform is by checking if a current-path URL parameter has been added.

#### Checklist


Adamik10 (approved):
Perfect! :)) good job.

PR #460, opened 2023-05-24
Fix back button issue by adding `e.stopPropagation()` to `CardListItem`

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFSOEG-484

#### Description

This PR introduces a fix for the issue where the back button wasn't working correctly after clicking the title link in the `CardListItem` component.

Changes:
- Added `e.stopPropagation()` to the onClick handler of the title link in the `CardListItem` component.

This change ensures that only the specific onClick event for the title link is triggered, which effectively prevents unintended event propagation. As a result, the back button now functions correctly after using the title link.

#### Checklist


Adamik10 (approved):
Good stuff :)

sinejespersen (approved):

PR #461, opened 2023-05-24
Add `useGetPatronInformationByPatronId` hook to ensure the user is login

kasperbirch1:
#### Link to issue

#### Description
This PR introduces a new hook, `useGetPatronInformationByPatronId`, which is used to ensure the user is logged in before fetching patron data.

The introduction of this hook improves the maintainability of the code, as it centralizes and standardizes the user check logic across several components.

#### Checklist


Adamik10:
Follow in #456

PR #464, opened 2023-05-30
Itkdev - CMN-630 - add listview to url

sinejespersen:
#### Link to issue

https://jira.itkdev.dk/browse/CMN-630

#### Description

add listview to url (for statistics)
add RecommenderView type
save RecommenderView in url (for statistics)
remove dash from creation of modal-url (due date had two dashes, that annoyed me)
removed unused faustIdModalQueryParam

#### Screenshot of the result

n/a

#### Checklist

n/a


yepzdk (approved):

src/apps/something-similar/SomethingSimilarList.tsx
```diff
@@ -25,15 +26,23 @@ const SomethingSimilarList: FC<SomethingSimilarListProps & MaterialProps> = ({
...
+  const [recommendView, setRecommendView] =
+    useState<RecommenderView>("similar");
   const { data: somethingSimilarData } = useRecommendFromFaustQuery({
     faust: id,
     limit: somethingSimilarLimit as number
   });
 
+  const setViewHandler = useCallback(
+    (inputView: RecommenderView) => {
+      window.history.pushState("", "", `?recommender-view=${inputView}`);
```
spaceo, line 40: Are you sure yow want to use pushState and thereby add to the browser history. Or do you want to rewrite the history? In that case you should use `setQueryParametersInUrl`.
But if you want to add to the history I think you should do following:

* Create a function for it in the [url helper file ](https://github.com/reload/dpl-react/blob/604f25f4a85f9ed6b9e6e9e542291aa7a9f76f18/src/core/utils/helpers/url.ts).
* I think it is a bad practise to hardcode query params like that. What if another url param has already been set by some other functionality? It is move safe to use the [searchParam API](https://github.com/reload/dpl-react/blob/604f25f4a85f9ed6b9e6e9e542291aa7a9f76f18/src/core/utils/helpers/url.ts#L31)

src/core/utils/helpers/general.ts
```diff
@@ -415,7 +415,7 @@ export const patronAgeValid = (cpr: string, minAge: number) => {
 };
 
 export const constructModalId = (prefix: string, fragments: string[]) =>
-  `${prefix ? `${prefix}-` : ""}${fragments.join("-")}`;
+  `${prefix ? `${prefix}` : ""}${fragments.join("-")}`;
```
spaceo, line 418: I don't get this change. Why shouldn't the prefix be followed by a separator?
Actually I have come up with a more nice way to define this function:
```typescript
export const constructModalId = (prefix: string, fragments: string[]) =>
  `${(prefix ? [prefix, ...fragments] : fragments).join('-')}`;

File: src/components/reservation/ReservationModal.tsx [hunk 3]
```diff
@@ -31,7 +31,7 @@ const ReservationModal = ({
         "reservationModalScreenReaderModalDescriptionText"
       )}
       closeModalAriaLabelText={t("reservationModalCloseModalAriaLabelText")}
-      dataCy={dataCy || "reservation-modal"}
+      dataCy={dataCy || "-reservation-modal-"}
```
spaceo, line 34: Why is this necessary? It does not look right to me.

spaceo (changes requested):

src/components/reservation/ReservationModal.tsx
sinejespersen: No, it isnt necessary, i can revert the code with the dashes :)

src/apps/something-similar/SomethingSimilarList.tsx
sinejespersen: I'll try setQueryParametersInUrl :)

spaceo (approved):
LGTM 👍

PR #466, opened 2023-05-31
itkdev: 5786: make aria label more specific

sinejespersen:
#### Link to issue

https://platform.dandigbib.org/issues/5786

#### Description

Make aria-label more specific

### DIFF
--- a/src/components/simple-material/SimpleMaterial.tsx
+++ b/src/components/simple-material/SimpleMaterial.tsx
@@ -27,7 +27,7 @@ export interface SimpleMaterialProps {
 const SimpleMaterial: FC<SimpleMaterialProps> = ({
   bright,
   work: {
-    titles: { full: title },
+    titles: { full: fullTitle },
     creators,
     workId,
     manifestations: { all: manifestations }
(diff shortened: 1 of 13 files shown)


yepzdk (approved):

src/components/simple-material/SimpleMaterial.tsx
```diff
@@ -73,6 +73,7 @@ const SimpleMaterial: FC<SimpleMaterialProps> = ({
       </div>
       <div className="simple-material__favourite">
         <ButtonFavourite
+          title={title[0]}
```
spaceo, line 76: Maybe I miss something but why does it work to use `title` both as an array and [cast it to a string](https://github.com/danskernesdigitalebibliotek/dpl-react/blob/63682a68b7cd8d727ad170db7df2a52bfd8ce1d4/src/components/simple-material/SimpleMaterial.tsx#L84)

spaceo (commented):
I have one question

src/components/simple-material/SimpleMaterial.tsx
sinejespersen: I can try? is it always an array with only one entry?

spaceo (approved):

PR #469, opened 2023-05-31
Introduce OpenOrder Logic and Components for 'Overbygningmaterialer'

kasperbirch1:
#### Link to issue

#### Description

This PR introduces the logic and components for ordering a book from 'overbygningmaterialer' (openOrder).

To determine that the material is available as an 'overbygningmateriale', I have made a function called `getReservableOnAnotherLibrary`. This function returns a boolean and a list of PIDS that start with 'OVE' in otherCatalogues.

I have created a new graphql mutation called `openOrder` that is called instead of the normal reservation inside the `ReservationModalBody`.

Because the response is so different from the normal reservation, I have made a separate component called `OpenOrderResponse` that handles the messages that are returned.

I have added a unit test for the `getReservableOnAnotherLibrary` and a cypress e2e test that shows different messages depending on the response.

#### Checklist


src/apps/material/material.dev.tsx
```diff
@@ -92,6 +92,11 @@ export default {
       defaultValue: "Reserve",
       control: { type: "text" }
     },
+    reservableOnAnotherLibraryText: {
```
spaceo, line 95: Isn't it more correct tos say: reservable FROM another library?

src/components/material/material-buttons/physical/MaterialButtonReservableOnAnotherLibrary.tsx
```diff
@@ -0,0 +1,56 @@
+import React, { FC } from "react";
+import { useModalButtonHandler } from "../../../../core/utils/modal";
+import { reservationModalId } from "../../../../apps/material/helper";
+import { useText } from "../../../../core/utils/text";
+import { ButtonSize } from "../../../../core/utils/types/button";
+import { FaustId } from "../../../../core/utils/types/ids";
+import { useUrls } from "../../../../core/utils/url";
+import { Button } from "../../../Buttons/Button";
+
+export interface MaterialButtonReservableOnAnotherLibraryProps {
```
spaceo, line 10: FROM anotherLibrary :)

src/components/material/material-buttons/physical/MaterialButtonReservableOnAnotherLibrary.tsx
```diff
@@ -0,0 +1,56 @@
...
+  manifestationMaterialType: string;
+  size?: ButtonSize;
+  faustIds: FaustId[];
+  dataCy?: string;
+}
+
+const MaterialButtonReservableOnAnotherLibrary: FC<
```
spaceo, line 17: From...

src/components/material/material-buttons/physical/MaterialButtonReservableOnAnotherLibrary.tsx
```diff
@@ -0,0 +1,56 @@
...
+  MaterialButtonReservableOnAnotherLibraryProps
```
spaceo, line 18: From...

spaceo: And all other "On"'s...

src/components/reservation/OpenOrderResponse.tsx
```diff
@@ -0,0 +1,69 @@
...
+import { OpenOrderMutation } from "../../core/dbc-gateway/generated/graphql";
+import { translateOpenOrderStatus } from "./helper";
+
+type OpenOrderResponseProps = {
+  title: string;
+  modalId: string;
+  openOrderResponse: OpenOrderMutation;
+};
+
+const OpenOrderResponse: React.FC<OpenOrderResponseProps> = ({
```
spaceo, line 16: I made a general modal message component that could be used here. I think we should look at it together...

src/components/reservation/ReservationModalBody.tsx
```diff
@@ -128,12 +139,47 @@ export const ReservationModalBody = ({
     ? getFutureDateString(selectedInterest)
     : null;
 
+  const {
+    isReservable: isReservableOnAnotherLibrar,
```
spaceo, line 143: missing an y in the end (and of course the On should be replaced with From"

src/components/reservation/ReservationModalBody.tsx
```diff
@@ -128,12 +139,47 @@ export const ReservationModalBody = ({
...
+    pids: pidsOnAnotherLibrar
```
spaceo, line 144: Missing an y

src/components/reservation/ReservationModalBody.tsx
```diff
@@ -128,12 +139,47 @@ export const ReservationModalBody = ({
...
+
+    if (isReservableOnAnotherLibrar && patron) {
+      const { patronId, address, name, emailAddress } = patron;
+
+      mutateOpenOrder(
+        {
+          input: {
+            pids: [...pidsOnAnotherLibrar],
+            pickUpBranch: selectedBranch
+              ? removePrefixFromBranchId(selectedBranch)
```
spaceo, line 160: Great that you are using a function from our tool box!

src/core/utils/helpers/general.ts
```diff
@@ -562,12 +562,89 @@ export const getContributors = (short: boolean, creators: string[]) => {
...
+          }
+        },
+        {
+          pid: "870970-basis:111111111",
+          catalogueCodes: {
+            otherCatalogues: ["ABE789"],
+            nationalBibliography: ["ABE456"]
+          }
+        }
+      ] as unknown as Manifestation[]);
```
spaceo, line 641: Isn't possible to set the type where you define the `result? const? This looks weird 😅 
```typescript
const result:  Manifestation[] = getReservableOnAnotherLibrary([
```

src/core/utils/helpers/general.ts
```diff
@@ -562,12 +562,89 @@ export const getContributors = (short: boolean, creators: string[]) => {
...
+    it("should return false for isReservable and empty pids array if no manifestations are reservable on another library", () => {
+      const result = getReservableOnAnotherLibrary([
+        {
+          pid: "pid1",
+          catalogueCodes: {
+            otherCatalogues: ["ABE789"],
+            nationalBibliography: ["ABE123"]
+          }
+        }
+      ] as unknown as Manifestation[]);
```
spaceo, line 619: Please see comment at line 641

src/core/utils/helpers/general.ts
```diff
@@ -562,12 +562,89 @@ export const getContributors = (short: boolean, creators: string[]) => {
...
+          }
+        },
+        {
+          pid: "870970-basis:135721719",
+          catalogueCodes: {
+            otherCatalogues: ["OVE124"],
+            nationalBibliography: ["ABE456"]
+          }
+        }
+      ] as unknown as Manifestation[]);
```
spaceo, line 601: Please see comment at line 641

src/core/utils/helpers/general.ts
```diff
@@ -562,12 +562,89 @@ export const getContributors = (short: boolean, creators: string[]) => {
...
+export const getReservableOnAnotherLibrary = (
+  manifestations: Manifestation[]
+) => {
+  const matchingManifestations = manifestations.filter(({ catalogueCodes }) =>
+    catalogueCodes?.otherCatalogues.some((code) => code.startsWith("OVE"))
+  );
+
+  return {
+    isReservable: matchingManifestations.length > 0,
+    pids: matchingManifestations.map(({ pid }) => pid)
```
spaceo, line 574: does it make sense to return two values here, when they are both dealing with `matchingManifestations`?
If you call the function:
```typescript
getReservablePidsFromAnotherLibrary()
```
...you could just return the pids.
And then you know that if there are pids coming from the function, then there is reservable material from another library.

src/components/reservation/helper.ts
```diff
@@ -251,4 +253,56 @@ export const getInstantLoanBranchHoldingsAboveThreshold = (
...
+      return t("openOrderServiceUnavailableText");
+    case SubmitOrderStatus.UnknownError:
+      return t("openOrderUnknownErrorText");
+    case SubmitOrderStatus.UnknownPickupagency:
+      return t("openOrderUnknownPickupagencyText");
+    case SubmitOrderStatus.UnknownUser:
+      return t("openOrderUnknownUserText");
+    default:
+      return "";
+  }
```
spaceo, line 305: I think it makes more sense to use a map here, and the lookup if you have a definition in the map that matches the status.

src/components/reservation/helper.ts
```diff
@@ -251,4 +253,56 @@ export const getInstantLoanBranchHoldingsAboveThreshold = (
     ({ materials }) => materials.length >= Number(instantLoanThresholdConfig)
   );
 
+export const removePrefixFromBranchId = (branchId: string) => {
+  const splitBranchId = branchId.split("-");
+  return splitBranchId[1];
+};
+
+export const getPatronAddress = (address: AddressV2) =>
```
spaceo, line 261: This is very global. Do we maybe have a patron helper file in core where we could place this instead?

src/components/reservation/ReservationModalBody.tsx
```diff
@@ -128,12 +139,47 @@ export const ReservationModalBody = ({
...
+              : "",
+            expires: selectedInterest?.toString() || "90",
```
spaceo, line 162: "90" should not be hard coded like this. We need it in our global config scope. Please take a look in core how it has been done. It is quite a mess there though, because ITK and Reload has been dealing with config in slightly different manners. We can talk about it... that is easier.

spaceo (changes requested):
Great work. I have some comments though

spaceo (approved):
LGTM 👍

PR #470, opened 2023-06-01
Feature/order from another library

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFSOEG-529

#### Description
This PR expands the MaterialButtonPhysical to be able to distinguish between regular physical material reservations and fluid reservations where the material is ordered from a different library because the local library doesn't have it.
The button label only enters this state (see pic) when the manifestations to be reserved support being ordered from elsewhere + there are no reservable manifestation of the selected material type in the local library.

#### Screenshot of the result
image
image

#### Checklist

This will have to be developed further when the FBI API is ready with all the changes that support this solution.

### DIFF
--- a/src/apps/material/material.dev.tsx
+++ b/src/apps/material/material.dev.tsx
@@ -82,6 +82,11 @@ export default {
       defaultValue: "Reserve",
       control: { type: "text" }
     },
+    reserveFromAnotherLibraryText: {
+      name: "Reserve from another library",
+      defaultValue: "Reserve @materialType elsewhere",
+      control: { type: "text" }
+    },
     findOnBookshelfText: {
       name: "Find on bookshelf",
       defaultValue: "Find on shelf",
--- a/src/apps/material/material.dev.tsx
+++ b/src/apps/material/material.dev.tsx
@@ -87,6 +87,11 @@ export default {
       defaultValue: "Reserve @materialType elsewhere",
       control: { type: "text" }
     },
+    reserveFromAnotherLibraryWithoutMaterialTypeText: {
+      name: "Reserve from another library",
+      defaultValue: "Reserve elsewhere",
+      control: { type: "text" }
+    },
     findOnBookshelfText: {
       name: "Find on bookshelf",
       defaultValue: "Find on shelf",
--- a/src/apps/material/material.dev.tsx
+++ b/src/apps/material/material.dev.tsx
@@ -811,6 +811,12 @@ export default {
       defaultValue:
         '[\n   {\n      "value":"30",\n      "label":"1 month"\n   },\n   {\n      "value":"60",\n      "label":"2 months"\n   },\n   {\n      "value":"90",\n      "label":"3 months"\n   },\n   {\n      "value":"180",\n      "label":"6 months"\n   },\n   {\n      "value":"360",\n      "label":"1 year"\n   }\n]',
       control: { type: "text" }
+    },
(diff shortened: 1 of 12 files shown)


src/apps/material/material.dev.tsx
```diff
@@ -82,6 +82,16 @@ export default {
       defaultValue: "Reserve",
       control: { type: "text" }
     },
+    reserveFromAnotherLibraryText: {
+      name: "Reserve from another library",
+      defaultValue: "Reserve @materialType elsewhere",
+      control: { type: "text" }
+    },
+    reserveFromAnotherLibraryWithoutMaterialTypeText: {
+      name: "Reserve from another library",
```
spaceo, line 91: This is probably a detail. But the Storybook controls will have to labels with the same text now that you have to name properties with the same value, tight?

spaceo (changes requested):
Please look at my comment about naming the `SpecialManifestation` type.

src/components/material/material-buttons/helper.ts
```diff
@@ -51,4 +51,19 @@ export const areAnyAvailable = (availability: AvailabilityV3[]) => {
   return availability.some((item) => item.available);
 };
 
+// TODO: SpecialManifestationProps & SpecialManifestation type will need to be
+// deleted when the Manifestation type includes the danMARC2 property.
+type SpecialManifestationProps = {
+  danMARC2: string;
+};
+export type SpecialManifestation = Manifestation & SpecialManifestationProps;
```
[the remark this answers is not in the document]
Adamik10: Hihihi but `SpecialManifestation` is not gonna stay. It's there only because we can't get the TS type from the autogenerated code yet 😉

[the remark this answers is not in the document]
spaceo: Ha, ok ;)

spaceo (approved):
LGTM 👍

src/apps/material/material.test.ts
```diff
@@ -183,6 +183,10 @@ describe("Material", () => {
 
     cy.getBySel("material-description").scrollIntoView();
 
+    // We need to wait here because no other fixes work.
+    // eslint-disable-next-line
+    cy.wait(1000);
```
sinejespersen, line 188: In this case, and the one below, do you know what we are waiting for?

sinejespersen (approved):
The linter gives a warning, and I have a question regarding the wait's in the cypress files. not a blocking one though.

src/apps/material/material.test.ts
spaceo: I think Sines question is relevant. Could you please document why the wait is necessary?

src/components/reservation/ReservationModalBody.tsx
```diff
@@ -128,10 +132,71 @@ export const ReservationModalBody = ({
...
+        ]
+      })
+        .then((data) => {
+          // This state is used to show the success or error modal.
+          setReservationResponse(data as ReservationResponseV2);
+          // Because after a successful reservation the holdings (reservations) are updated.
+          queryClient.invalidateQueries(getGetHoldingsV3QueryKey());
+        })
+        .catch((error) => {
+          console.log(error);
```
spaceo, line 191: What is supposed to happen here in real life?

spaceo (changes requested):
Nice :). Please look at my questions

PR #475, opened 2023-06-06
Itkdev - 5624 - validate date on save

sinejespersen:
#### Link to issue

https://platform.dandigbib.org/issues/5624

#### Description

It now validates dates before it saves (basic browser validation, agreed on with fddf).

#### Screenshot of the result

n/a 

#### Checklist

no


yepzdk (approved):

src/apps/reservation-list/modal/pause-reservation/pause-reservation.tsx
```diff
@@ -31,41 +31,45 @@ const PauseReservation: FC<PauseReservationProps> = ({ id, user }) => {
...
-      const saveData = {
-        preferredPickupBranch: user.preferredPickupBranch,
-        receiveEmail: user.receiveEmail,
-        receivePostalMail: user.receivePostalMail,
-        receiveSms: user.receiveSms
-      } as Patron;
+  const save = useCallback(
+    (e: FormEvent<HTMLFormElement>) => {
+      e.preventDefault();
+      if (user) {
```
spaceo, line 37: I think it is prettier to make a guard, which also reduce indentation - in this case for quite a big block of code.
So instead of checking for user is true you can return quickly in the beginning if it is false.
```typescript 
  const save = useCallback(
    (e: FormEvent<HTMLFormElement>) => {
      e.preventDefault();
      if (!user) {
        return;
     }


```

spaceo: One other thing I thought of:
When are we naming an object `user` and when do we name it `patron`?

spaceo (approved):
Approved. But please read remarks

PR #479, opened 2023-06-07
itkdev: 5568: add et al text, and use style classes

sinejespersen:
#### Link to issue

https://platform.dandigbib.org/issues/5568

#### Description

Friends with: 
https://github.com/danskernesdigitalebibliotek/dpl-design-system/pull/234
https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/343

#### Screenshot of the result

n/a 

#### Checklist

n/a


yepzdk:
It looks like some tests still need some attention.

yepzdk (approved):

src/core/utils/helpers/general.ts
```diff
@@ -481,23 +482,54 @@ export const getReleaseYearSearchResult = (work: Work) => {
...
+    .join(", ")} ${andText} ${creators.slice(-1)}`;
+};
+
+// Creates a "by author, author and author"-string, or by author, author et al.
+export const getContributors = (short: boolean, creators: string[]) => {
+  // Todo this is sortof a hack, but using t: UseTextFunction as argument
+  // makes the components re-render.
+  const {
+    text: { data: texts }
+  } = store.getState();
```
spaceo, line 509: I really think we should avoid using the state directly. Either I think you should create it as a hook with `useSelector` to fetch text from the state or pass in the t function as a parameter.

spaceo (changes requested):
One thing I think should be changed...

src/core/utils/helpers/general.ts
sinejespersen: I've just moved it from one place to another. I did this a long time ago, because for reasons I could not figure out, if i sent the t function as a parameter endless loops would occur.

PR #480, opened 2023-06-08
itkdev: add showBadgeWithoutDueDate

sinejespersen:
#### Link to issue

n/a

#### Description

Loan list should only show status badge in connection to due date

#### Screenshot of the result

n/a

#### Checklist

n/a

### DIFF
--- a/src/apps/loan-list/materials/stackable-material/material-status.tsx
+++ b/src/apps/loan-list/materials/stackable-material/material-status.tsx
@@ -59,6 +59,7 @@ const MaterialStatus: FC<MaterialStatusProps> = ({
         <div className="list-reservation__deadline">
           {dueDate && (
             <StatusBadge
+              showBadgeWithoutDueDate={false}
               badgeDate={dueDate}
               dangerText={t("loanListStatusBadgeDangerText")}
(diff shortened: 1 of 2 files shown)


yepzdk (approved):

src/apps/loan-list/materials/stackable-material/material-status.tsx
```diff
@@ -59,6 +59,7 @@ const MaterialStatus: FC<MaterialStatusProps> = ({
         <div className="list-reservation__deadline">
           {dueDate && (
             <StatusBadge
+              showBadgeWithoutDueDate={false}
```
spaceo, line 62: Normally with boolean props you default to false and then add it if it should be true.
So you could flip the logic around and write:
```typescript
<StatusBadge showBadgeWithDueDate />
```

And then:
```typescript
const StatusBadge: FC<StatusBadgeProps> = ({
  badgeDate,
  warningText,
  showBadgeWithDueDate = false,
  dangerText,
  infoText,
```
and so forth...

spaceo (approved):
LGTM 👍 
Have one suggestion though...

PR #485, opened 2023-06-12
itkdev: switch logic on contact info section

sinejespersen:
#### Link to issue

n/a

#### Description

Correct error with switched logic 
added some missing props to menu
add styling classes to "something similar"

### DIFF
--- a/src/apps/dashboard/dashboard.dev.tsx
+++ b/src/apps/dashboard/dashboard.dev.tsx
@@ -134,6 +134,10 @@ export default {
       },
       defaultValue: "Number @count in line"
     },
+    reservationDetailsOthersInQueueText: {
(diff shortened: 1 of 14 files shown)


src/apps/dashboard/dashboard.dev.tsx
```diff
@@ -134,6 +134,10 @@ export default {
       },
       defaultValue: "Number @count in line"
     },
+    reservationDetailsOthersInQueueText: {
+      defaultValue: "Others are queueing for this material",
```
spaceo, line 138: Have you double checked our code for a similar text? Wouldn't it make sense to reuse translations in some places instead of having different context everywhere?

spaceo (approved):
Looks good. One comment

PR #488, opened 2023-06-14
Add Vitest in source testing

spaceo:
#### Description
While creating tests for some link component functionality I ran into an issue with testing/mocking a function that originally was an unexported function. There are other questionable hacks you can do
that is not a part of the Vitest recommended practices. However does Vitest have the [in source testing concept](https://vitest.dev/guide/in-source.html#setup) that solves the issue. I suggest we make use of that.

#### Checklist


sinejespersen (approved):

kasperbirch1 (approved):
Nice! I actually saw it in this video yesterday 👍

https://www.youtube.com/watch?v=7f-71kYhK00

PR #489, opened 2023-06-14
5389: make room for fixed buttons/pager

sinejespersen:
#### Link to issue

https://platform.dandigbib.org/issues/5389

#### Description

In some situations the pager and the fixed button in the bottom of the modals had a bit of an unfortunate overlap

friends with: https://github.com/danskernesdigitalebibliotek/dpl-design-system/pull/239

#### Screenshot of the result

image

#### Checklist

n/a

### DIFF
--- a/src/components/result-pager/result-pager.tsx
+++ b/src/components/result-pager/result-pager.tsx
@@ -5,17 +5,19 @@ export interface ResultPagerProps {
   setPageHandler: () => void;
   itemsShown: number;
   hitcount: number;
+  classNames: string;
   isLoading?: boolean;
 }
 function ResultPager({
   setPageHandler,
   itemsShown,
   hitcount,
-  isLoading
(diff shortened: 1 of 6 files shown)


yepzdk (approved):

src/components/result-pager/result-pager.tsx
```diff
@@ -5,17 +5,19 @@ export interface ResultPagerProps {
...
   itemsShown,
   hitcount,
-  isLoading
+  isLoading,
+  classNames
 }: ResultPagerProps) {
   const t = useText();
   return (
-    <div className="result-pager">
+    <div className={`result-pager ${classNames}`}>
```
kasperbirch1, line 20: I prefer to use clsx here

kasperbirch1 (approved):
LGTM 👍

PR #494, opened 2023-06-15
Add inline tests for facet-browser helper functions

kasperbirch1:
#### Description

This PR introduces in-source testing with `vitest`. 

The test cases use `toMatchSnapshot` to create the output, which is then monitored for changes through `git` whenever there is a refactoring of the functions.

The test cases are written to ensure the functionality of several utility functions.

#### Checklist

Related to https://github.com/danskernesdigitalebibliotek/dpl-react/pull/488


spaceo (changes requested):
Great addition! But I think the snapshots should be moved to separate snapshot files.
Could they reside either in a folder next to the file called "snapshots" or inside the unit test dir also in a folder called "snapshots"?

spaceo (changes requested):
There is a problem with the snapshot paths. They are not unique enough.

spaceo (approved):

sinejespersen (approved):

PR #495, opened 2023-06-20
Simplifying modal id creation

spaceo:
#### Description

Since prefix is required there is no need for checking it. Add test that proves it works.

#### Screenshot of the result

#### Checklist


sinejespersen (approved):

kasperbirch1 (approved):
LGTM 👍

PR #508, opened 2023-07-18
5882: streamline remove reservations

sinejespersen:
#### Link to issue

https://platform.dandigbib.org/issues/5882

#### Description

Consolidate some functionality

### DIFF
--- a/src/apps/dashboard/modal/ReservationsGroupModal.tsx
+++ b/src/apps/dashboard/modal/ReservationsGroupModal.tsx
@@ -112,6 +112,11 @@ const ReservationGroupModal: FC<ReservationGroupModalProps> = ({
     setMaterialsToDelete(materialIds);
   };
 
+  const deleteMaterialsModal = () => {
+    setReservationsToDelete(materialsToDelete);
+    setMaterialsToDelete([]);
+  };
+
   return (
     <Modal
       modalId={modalId}
@@ -146,7 +151,7 @@ const ReservationGroupModal: FC<ReservationGroupModalProps> = ({
                 collapsible={false}
                 size="small"
                 variant="filled"
-                onClick={() => setReservationsToDelete(materialsToDelete)}
+                onClick={() => deleteMaterialsModal()}
               />
             }
             amountOfSelectableMaterials={selectableReservations.length}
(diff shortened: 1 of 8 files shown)


jeppekroghitk (approved):

src/apps/dashboard/modal/ReservationsGroupModal.tsx
```diff
@@ -32,22 +22,20 @@ import { mapPublizonReservationToReservationType } from "../../../core/utils/hel
 interface ReservationGroupModalProps {
   pageSize: number;
   modalId: string;
+  setReservationsToDelete: (reservations: string[]) => void;
```
kasperg, line 25: Please consider if typing and or naming of this can be improved.

To avoid just passing strings around we have types for other ids. Could we create a new type for reservation ids and use this here?

Since we are not expecting `ReservationType` instances here I would also suggest renaming the value to `reservationIds`.

src/apps/reservation-list/modal/delete-reservation/delete-reservation-modal.tsx
```diff
@@ -7,65 +7,64 @@ import {
...
+import {
+  getGetV1UserReservationsQueryKey,
+  useDeleteV1UserReservationsIdentifier
+} from "../../../../core/publizon/publizon";
+import { isFaust, isIdentifier } from "../../../dashboard/util/helpers";
 
 interface DeleteReservationModalProps {
   modalId: string;
-  reservation: ReservationType;
+  reservations: string[];
```
kasperg, line 18: As mentioned above: Please look into stronger typing here.

If we can not pass an array of `ReservationType` then look into an array of reservation ids. In that case naming should also be clarified.

src/apps/reservation-list/modal/delete-reservation/delete-reservation-modal.tsx
```diff
@@ -7,65 +7,64 @@ import {
...
-    if (!isDigital(reservation) && reservation.reservationId) {
+
+  const removeSelectedReservations = () => {
+    if (reservations.length > 0) {
+      const reservationsToDelete = reservations
+        .map((id) => Number(isFaust(id)))
+        .filter((id) => id !== 0);
+      const digitalMaterialsToDelete = reservations
+        .map((id) => isIdentifier(id))
+        .filter((id) => id !== null);
```
kasperg, line 39: Please consider whether map and filter are both needed.

If the reservation id is a Faust then we know it is a number. If it is an identifier then we know it is not null.

jeppekroghitk: The faustid is normally handled as a string in the system, but the api endpoint for deletion only accepts integers. That is the reason why we cast it as we do here, altho it might seem redundant. I removed the null check, but it's probably there for a reason.

PR #509, opened 2023-07-19
5843: readd renew button, classes were set wrong

sinejespersen:
#### Link to issue

https://platform.dandigbib.org/issues/5843

#### Description

- classes were set on button instead of outer div, it is now fixed

### DIFF
--- a/src/apps/loan-list/modal/renew-button.tsx
+++ b/src/apps/loan-list/modal/renew-button.tsx
@@ -48,13 +48,12 @@ const RenewButton: FC<RenewButtonProps> = ({
   );
 
   return (
-    <div className="modal-details__buttons">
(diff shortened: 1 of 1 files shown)


jeppekroghitk (approved):

src/apps/loan-list/modal/renew-button.tsx
```diff
@@ -48,13 +48,12 @@ const RenewButton: FC<RenewButtonProps> = ({
   );
 
   return (
-    <div className="modal-details__buttons">
+    <div className={`modal-details__buttons ${classNames}`}>
```
kasperg, line 51: Consider using `clsx` for constructing `className` values. This is quite a simple case so here it is OK to not do so.

kasperg (approved):
:+1: LGTM

PR #511, opened 2023-07-26
Cleaning up getParams

spaceo:
#### Description

Got tired of ts warnings and thought [getParams](https://github.com/danskernesdigitalebibliotek/dpl-react/pull/511/files#diff-a09d47a1eb335411b7f4c67799d5a4d59bac0b55e8dcf3d3142c99f86d309378L169-L177) could get more lean.
Also wrote a test for it.

#### Checklist

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -12,7 +12,12 @@
   "settings": {
     "react": {
       "version": "16.11.0"
-    }
+    },
+    // Since we use vitest alongside our production code we have to instruct eslint
+    // not to throw the import/no-extraneous-dependencies error when doing so.
+    "import/core-modules": [
+      "vitest"
+    ]
   },
   "env": {
     "browser": true,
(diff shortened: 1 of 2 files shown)


sinejespersen (approved):

.eslintrc
```diff
@@ -12,7 +12,12 @@
   "settings": {
     "react": {
       "version": "16.11.0"
-    }
+    },
+    // Since we use vitest alongside our production code we have to instruct eslint
```
Adamik10, line 16: Love it that you wrote the reasoning in a comment.

src/core/utils/helpers/general.ts
```diff
@@ -545,4 +548,22 @@ if (import.meta.vitest) {
       );
     });
   });
+
+  describe("getParams", () => {
```
Adamik10, line 552: is there a reason why we don't test in a separate file? Aren't the file sizes just gonna blow up this way?

Adamik10 (approved):
Looks good, but I do have a question :)

PR #516, opened 2023-08-15
Feature/5885 accept modal

sinejespersen:
#### Link to issue

https://platform.dandigbib.org/issues/5885

#### Description

Added modal for accepting higher fees 

#### Screenshot of the result

image

#### Checklist

A pull request in dpl-cms with texts will follow...


src/core/utils/helpers/general.ts
```diff
@@ -329,6 +329,20 @@ export const getPhysicalQueuedReservations = (list: ReservationType[]) => {
   );
 };
 
+export const loansOverdue = (loans: LoanType[]): boolean => {
+  let areAllLoansOverdue = true;
```
spaceo, line 333: Just a suggestion:
Why not use `.every()` here as you do in line 343? Then you can skip the `let`.
I did not know of the `every()`function. It is quite nifty :)

src/components/accept-modal/AcceptModal.tsx
```diff
@@ -0,0 +1,59 @@
...
+import { getModalIds } from "../../core/utils/helpers/general";
+import Modal, { useModalButtonHandler } from "../../core/utils/modal";
+import { useText } from "../../core/utils/text";
+import { Button } from "../Buttons/Button";
+
+export interface AcceptModalProps {
+  accept: () => void;
+}
+
+const AcceptModal: FC<AcceptModalProps> = ({ accept }) => {
```
spaceo, line 12: Accept modal or /components/accept-modal is a quite generic namespace. Do you think the component is generic enough to be named like that or is it more tied to fees?

src/components/GroupModal/LoansGroupModal.tsx
```diff
@@ -54,14 +64,43 @@ const LoansGroupModal: FC<LoansGroupModalProps> = ({
...
     setMaterialsToRenew(materialIds);
   };
 
+  useEffect(() => {
+    if (acceptedButtonPressed) {
+      renew();
+      setAcceptedButtonPressed(false);
+    }
+  }, [acceptedButtonPressed, renew]);
+
```
spaceo, line 103: This seems funky to me. Could be that it makes sense. But having ìf acceptedButtonPressed then renew() and  setAcceptedButtonPressed(false)` instead of sorting that out in a clickhandler has a code smell to me. Could be wrong though :)

spaceo (commented):
Looks good. I have some comments I would like have clarified though :)

src/components/GroupModal/LoansGroupModal.tsx
sinejespersen: it was to keep the renew function in _one_ file.

src/components/accept-modal/AcceptModal.tsx
sinejespersen: can change the name to accept-fee-modal?

jeppekroghitk (approved):

spaceo (approved):
LGTM 👍

PR #517, opened 2023-08-22
Remove loginRequired check in MaterialButtonOnlineExternal

spaceo:
#### Link to issue

https://reload.atlassian.net/browse/DDFSOEG-490

#### Description

Since it is not working on DBC's side the decision has been made to remove the check for that property

#### Checklist


Adamik10 (approved):
LGTM :)

kasperg (approved):
:+1: LGTM

PR #518, opened 2023-08-23
Make availability label texts translatable

spaceo:
#### Link to issue

https://reload.atlassian.net/browse/DDFDPDEL-216

#### Description

The definition of the availability texts where defined halfways. This PR names them with a "availability" prefix that makes it clear what the purpose of the strings are. The [PR](https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/381) in the dpl-cms repository will make sure that these strings are added as well.

#### Checklist

### DIFF
--- a/cypress/support/index.ts
+++ b/cypress/support/index.ts
@@ -50,7 +50,7 @@ Cypress.Commands.add(
     fixtureFilePath,
     statusCode = 200
   }: InterceptGraphqlParams) => {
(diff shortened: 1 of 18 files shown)


Adamik10 (approved):
Looks gooddie. It's a lot of files that changed but I get why.

kasperg (approved):
:+1: LGTM

PR #544, opened 2023-09-19
Subject line on search results

spaceo:
#### Link to issue

https://reload.atlassian.net/browse/DDFSOEG-549

#### Description

Adds a subject line to search results that are non fictional.

#### Checklist

### DIFF
--- a/src/components/subject-number/SubjectNumber.tsx
+++ b/src/components/subject-number/SubjectNumber.tsx
@@ -1,10 +1,10 @@
 import * as React from "react";
 import { FC } from "react";
-import { Manifestation } from "../../core/utils/types/entities";
+import { Shelfmark } from "../../core/dbc-gateway/generated/graphql";
 
 export interface SubjectNumberProps {
   className?: string;
-  shelfmark: NonNullable<Manifestation["shelfmark"]>;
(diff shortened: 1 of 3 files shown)


src/components/subject-number/SubjectNumber.tsx
```diff
@@ -0,0 +1,15 @@
+import * as React from "react";
+import { FC } from "react";
+import { Manifestation } from "../../core/utils/types/entities";
+
+export interface SubjectNumberProps {
+  className?: string;
+  shelfmark: NonNullable<Manifestation["shelfmark"]>;
```
kasperg, line 7: I find it hard to decide when to use the underlying primitive type (in this case `string`?) and when to use the type of the property of a larger type 🤔 .

`NonNullable<Manifestation["shelfmark"]>` does not seem very readable compared to e.g. `string`. In this case we seem to assume when rendering the component that the value can be stringified anyway. 

Perhaps this is something we can talk about at some point.

kasperg (approved):
:+1: LGTM

Is it a business requirment that should be verified in a test?

Adamik10 (approved):
Looks good :)

PR #546, opened 2023-09-20
Skip release workflow for Dependabot

kasperg:
#### Description

Dependabot does not have permissions to create releases. This causes the workflow to fail when running ghr with the following error:

Failed to create GitHub release page: failed to create a release: POST https://api.github.com/repos/danskernesdigitalebibliotek/dpl-react/releases: 403 Resource not accessible by integration []

According to GitHub having Dependabot run with read-only scopes is intentional and the recommended approach is to skip running such  workflows based on branch naming:
https://docs.github.com/en/code-security/code-scanning/troubleshooting-code-scanning/resource-not-accessible-by-integration

#### Screenshot of the result

Before:

dpl-react@dca6122 2023-09-20 16-45-37

#### Checklist

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,7 +1,10 @@
 name: Upload Release Asset
 
 on:
-  push
+  push:
+    branches-ignore:
+      # Dependabot does not have permissions to create releases
+      - 'dependabot/**'
 
 jobs:
   build:


Adamik10 (approved):
Approved :)

spaceo (approved):
LGTM 👍

PR #548, opened 2023-09-22
Fix gh action cypress image to use chrome 116

spaceo:
Because of errors in the communication between Cypress and Chrome 117+ we need to make sure that we run Chrome 116.
This is a temporary solution and we should upgrade the cypress action in the future as a part of the work involved in upgrading Cypress.

#### Link to issue

#### Description

#### Screenshot of the result

#### Checklist


Adamik10 (approved):
I love me a good quick fix to unblock a workflow! :) Approved.

kasperg (approved):
:+1: LGTM

Consider adding references such as:

- The GitHub issue that lead us to the problem: https://github.com/cypress-io/cypress/issues/27804
- The Cypress documentation of how to run GitHub Actions with specific versions of browsers. Using Docker images with GitHub Actions is not something that we do elsewhere in the project so having a reason for why it is necessary is nice.

PR #552, opened 2023-09-27
ADR: Config handling

kasperg:
#### Link to issue

https://reload.atlassian.net/browse/DDFSOEG-558

#### Description

This documents our decisions regarding how to manage responsibilities for providing correct configuration between React and Drupal.

It also updates other ADRs to improve their presentation [on the documentation site](https://danskernesdigitalebibliotek.github.io/dpl-docs/dpl-react/). If you disagree about that part I can leave it out.


spaceo (approved):
Looks great.
Maybe it is ok, but I am in doubt if everybody understand what "the consuming application" is. Alternatively you could use another wording or maybe start the ADR with defining what the term means. Because in a way the react applications are also "consuming" the configuration from Drupal.

PR #556, opened 2023-09-27
Add possibility to define service urls in .env

spaceo:
#### Description

We will get tired of remembering to add it to storybook args and be careful not commit the mocked services. In the future we can handle other environment variables for other purposes as well.

#### Checklist


kasperg (approved):
:+1: LGTM. This is a nice improvement.

PR #557, opened 2023-09-28
Fix bug with subject number showing on all search results

spaceo:
#### Link to issue

https://reload.atlassian.net/browse/DDFSOEG-549

#### Description

The wrong object (work) was passed to materialIsFiction(). The bestRepresentation manifestation has the information.

#### Checklist


Adamik10 (approved):
Wonderful little fix :))

PR #558, opened 2023-09-29
Refactor `handleDropdownOnchange` in `FacetLineFilters`

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFSOEG-525

#### Description

We identified that the intelligentFacets query can sometimes return facets with a term intended for shortcut buttons. This same facet might also be present with other terms. To handle this efficiently and enhance our search capabilities, we've made the following updates:

- Added `createFacetsMap` Function:
This function is aimed to transform the facets array into an object. The main objective behind this transformation is to enhance search capabilities within the facets.

- Refactored `handleDropdownOnchange` Logic:
Incorporated the use of `createFacetsMap` within this function.

#### Screenshot of the result

https://github.com/danskernesdigitalebibliotek/dpl-react/assets/49920322/8ad57abf-0e00-4323-8f62-fd23a8323235

#### Checklist

### DIFF
--- a/src/components/Dropdown/Dropdown.tsx
+++ b/src/components/Dropdown/Dropdown.tsx
@@ -46,6 +46,7 @@ const Dropdown: React.FunctionComponent<DropdownProps> = ({
 
   const checkHandleOnChange = (e: React.ChangeEvent<HTMLSelectElement>) => {
     if (handleOnChange && e.target.value) handleOnChange(e);
+    console.log("No handleOnChange function passed to Dropdown");
     return undefined;
   };
 
--- a/src/components/Dropdown/Dropdown.tsx
+++ b/src/components/Dropdown/Dropdown.tsx
@@ -45,8 +45,9 @@ const Dropdown: React.FunctionComponent<DropdownProps> = ({
   };
 
   const checkHandleOnChange = (e: React.ChangeEvent<HTMLSelectElement>) => {
-    if (handleOnChange && e.target.value) handleOnChange(e);
-    console.log("No handleOnChange function passed to Dropdown");
+    if (handleOnChange && e.target.value) {
+      handleOnChange(e);
+    }
     return undefined;
   };
 
--- a/src/components/facet-browser/helper.ts
+++ b/src/components/facet-browser/helper.ts
@@ -134,26 +134,31 @@ export const getFacetFieldTranslation = (name: FacetField) => {
   }
 };
 
+type FacetMap = { [key: string]: FacetValue };
 // createFacetsMap generates a map for quick lookup and enhanced search
 // capabilities by combining facet name and term into a single string key.
 // Structure: { [facetName:termKey]: FacetValue }
 // Example Key: 'fictionalCharacters:Batman'
-export const createFacetsMap = (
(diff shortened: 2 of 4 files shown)


src/components/Dropdown/Dropdown.tsx
```diff
@@ -46,6 +46,7 @@ const Dropdown: React.FunctionComponent<DropdownProps> = ({
 
   const checkHandleOnChange = (e: React.ChangeEvent<HTMLSelectElement>) => {
     if (handleOnChange && e.target.value) handleOnChange(e);
+    console.log("No handleOnChange function passed to Dropdown");
```
spaceo, line 49: Please remove console statement

src/components/facet-browser/helper.ts
```diff
@@ -132,8 +134,29 @@ export const getFacetFieldTranslation = (name: FacetField) => {
   }
 };
 
+export const createFacetsMap = (
```
spaceo, line 137: I am bad at remembering this myself. But I think it is very valuable to write a small comment about what the purpose of a function is. Unless it is self-explanatory, eg.: getUrlQueryParams()
So in this case: "what is a facet map and why is it needed?"

src/components/facet-browser/helper.ts
```diff
@@ -132,8 +134,29 @@ export const getFacetFieldTranslation = (name: FacetField) => {
...
+): { [key: string]: FacetValue[] } => {
+  return facets.reduce(
+    (acc: { [key: string]: FacetValue[] }, facet: FacetResult) => {
+      const newAcc = { ...acc };
+
+      facet.values.forEach((value) => {
+        const combinedKey = `${facet.name}:${value.key}`;
+        newAcc[combinedKey] = newAcc[combinedKey]
+          ? [...newAcc[combinedKey], value]
+          : [value];
```
spaceo, line 148: Isn't it possible just to write:
```typescript
newAcc[combinedKey] = value;
```
If you overwrite it does not matter because the new object should be the same as the former.

src/components/facet-line/FacetLineFilters.tsx
```diff
@@ -38,9 +40,7 @@ const FacetLineFilters: React.FunctionComponent<FacetLineFiltersProps> = ({
     e: React.ChangeEvent<HTMLSelectElement>,
     facet: string
   ) => {
-    const term = facets
-      .find((item) => item.name === facet)
-      ?.values.find((item) => item.key === e.target.value);
+    const term = facetMap[`${facet}:${e.target.value}`][0];
```
spaceo, line 43: Two things:
* Why are the values in the map arrays? Isn't it just necessary to have one object?
* Since you made this nice facet map I think it would make sense to make a function that finds the value/object from the given facet/termName. Eg.: `findTermInFacetMap(facetName, termName)`

spaceo (changes requested):
Great work.
I have a few comments I would like you to look at.

spaceo:
Btw: Great with the screencast! I am really a fan of that.

spaceo (approved):
LGTM 👍

PR #572, opened 2023-10-02
"Feedback på fornyelser - samlesag", Part I

spaceo:
#### Link to issue

https://github.com/danskernesdigitalebibliotek/dpl-design-system/blob/develop/.nvmrc

#### Description

This is part I of the "Feedback på fornyelser - samlesag".
It makes sure that there is a modal feedback when renewing materials.
The next part will contain some Cypress tests and look into if the cache is properly cleared when material has been renewed.

#### Screenshot of the result

https://github.com/danskernesdigitalebibliotek/dpl-react/assets/998889/f2232857-0db4-4815-9c83-5a116b1247b7

#### Checklist


src/apps/dashboard/dashboard-notification-list/dashboard-notification-list.tsx
```diff
@@ -322,11 +324,20 @@ const DashboardNotificationList: FC<DashboardNotificationListProps> = ({
         />
       )}
 
-      <MaterialDetailsModal modalId={`${loanDetails}${modalLoanDetailsId}`}>
+      <MaterialDetailsModal
+        modalId={constructMaterialDetailsModalId(
```
kasperbirch1, line 328: Maybe we should declare modalId as a const to avoid multiple function calls and duplicate code? 😊

kasperbirch1 (approved):

PR #577, opened 2023-10-06
Get rid of node in docker and adjust task cmds

spaceo:
We want to be able to run node/yarn in our hosts but at the same time keep the wiremock functionality which is helpful especially in connection working with patron functionality/data.


kasperg (changes requested):
Remember to update the documentation accordingly. Our list of required software and howtos should be updated if there are usage that we no longer want to support.

A few words in the commit comment about *why* we prefer to run node on the host would be good as well.

docs/README.md
```diff
@@ -24,18 +26,16 @@ npm login --registry=https://npm.pkg.github.com
...
-5. If you need to log in through Adgangsplatformen, you need to change
-   your url to <http://dpl-react.docker/> instead of <http://localhost>. This
-   avoids getting log in errors
+1. Ensure that your Node version matches what is specified in `.nvmrc`.
+2. Run `task dev:start`
+   - Storybook will open automatically in a browser
+   - The console will contain build and lint information
+3. If you need to log in through Adgangsplatformen:
+   - Add `127.0.0.1 dpl-react.docker` to your `/etc/hosts` file
+   - [Ensure that Node can bind to port 80](https://gist.github.com/guifromrio/6389682)
```
JacobArrow, line 35: Good to have the info about port 80 binding here

JacobArrow (approved):
I think this looks great! 🔥

PR #578, opened 2023-10-06
Replaced node-sass with sass

JacobArrow:
#### Description

As it turned out the node-sass version we were using, was causing trouble when compiling on Windows, we decided to replace it with sass instead.

To keep the integration with postcss we are using a postcss-sass package to compile scss correctly

#### Checklist


kasperg (approved):
:+1: LGTM

Consider squashing your extra commit before merging.

PR #584, opened 2023-10-09
Fix screen reader feedback when changing user password

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFLSBP-85

#### Description

We utilize the role="alert" attribute on the messages. This way, if the user is using a screen reader, they will be notified about the feedback messages as they appear.

#### Screenshot of the result
n/a

#### Checklist

n/a


spaceo (approved):
LGTM 👍

PR #586, opened 2023-10-10
Group dependabot updates for certain vendors

kasperg:
#### Description

We use many packages from Storybook, GraphQL Codegen and Babel. These are often updated in parallel and may have dependencies between them. Thus is makes sense to group these updates of these packages in Dependabot to ensure that we get an update to e.g. Storybook 7, GraphQL Codegen 5, Babel 8 in the same pull request


spaceo (approved):
LGTM 👍

PR #591, opened 2023-10-11
Fix/user name in header

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFLSBP-122

#### Description

#### Screenshot of the result

#### Checklist

### DIFF
--- a/src/components/search-bar/story-header.dev.inc.tsx
+++ b/src/components/search-bar/story-header.dev.inc.tsx
@@ -1,15 +1,18 @@
 import * as React from "react";
 import menuIcon from "@danskernesdigitalebibliotek/dpl-design-system/build/icons/basic/icon-menu.svg";
+import searchIcon from "@danskernesdigitalebibliotek/dpl-design-system/build/icons/basic/icon-search.svg";
 import profileIcon from "@danskernesdigitalebibliotek/dpl-design-system/build/icons/basic/icon-profile.svg";
 import heartIcon from "@danskernesdigitalebibliotek/dpl-design-system/build/icons/basic/icon-heart.svg";
 import watchIcon from "@danskernesdigitalebibliotek/dpl-design-system/build/icons/basic/icon-watch-static.svg";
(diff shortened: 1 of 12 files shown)


src/components/search-bar/story-header.dev.inc.tsx
```diff
@@ -94,18 +97,39 @@ const StoryHeader: React.FC<StoryHeaderProps> = ({ children }) => {
                 </li>
               </ul>
             </div>
-            <div className="header__menu-profile header__button">
-              <a href="/" className="hide-linkstyle">
+            {userProfile || (
+              <div className="header__menu-profile header__button">
                 <img src={profileIcon} alt="Profile" />
```
JacobArrow, line 102: Do we not want to show the newly added user name to the story?

src/components/skeletons/TextLineSkeleton.tsx
```diff
@@ -0,0 +1,10 @@
+import * as React from "react";
+
+export interface TextLineSkeletonProps {
+  width: number;
+}
+const TextLineSkeleton: React.FC<TextLineSkeletonProps> = ({ width }) => (
+  <div className="ssc-line w-100" style={{ width: `${width}px` }} />
```
JacobArrow, line 7: Is the `w-100` a class for setting width?
Isn't that a bit redundant if the width prop is required to be set?

JacobArrow (approved):
LGTM 🔥

src/components/search-bar/story-header.dev.inc.tsx
Adamik10: We want to show it if `userProfile` is defined, which it is in the "header -> user menu" story. But we don't want that in the "header -> search" story. :))

src/components/skeletons/TextLineSkeleton.tsx
Adamik10: Good thinking!

PR #592, opened 2023-10-12
Update/ghr in release.yml GH action

Adamik10:
#### Link to issue
n/a

#### Description

We experienced some issues where GHR sometimes marks a GH release as a draft, and sometimes as pre-release without a clear reasoning. We hope that this update will fix this inconsistancy.

#### Screenshot of the result
n/a

#### Checklist

n/a


kasperg (approved):
:+1: LGTM

PR #593, opened 2023-10-12
Mail should not be required in user info

spaceo:
#### Link to issue

https://reload.atlassian.net/browse/DDFLSBP-208
https://reload.atlassian.net/browse/DDFLSBP-110

#### Description

There has been some confusion about mail and phone no., wether they should be required or not.
Apparently the rule is:
None of them are required, except mail on the user registration form.

### DIFF
--- a/src/components/contact-info-section/ContactInfoSection.tsx
+++ b/src/components/contact-info-section/ContactInfoSection.tsx
@@ -51,7 +51,6 @@ const ContactInfoSection: FC<ContactInfoSectionProps> = ({
           changePatron={changePatron}
           patron={patron}
           showCheckboxes={showCheckboxes}
-          isRequired
         />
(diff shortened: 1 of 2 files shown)


src/components/contact-info-section/ContactInfoSection.tsx
```diff
@@ -13,13 +13,15 @@ interface ContactInfoSectionProps {
   inLine: boolean;
   changePatron: ChangePatronProps;
   showCheckboxes: boolean;
+  requiredFields?: ("email" | "phone")[];
```
kasperg, line 16: Personally I prefer pulling this into a separate type - even if it just a simple string literal.

kasperg (approved):
:+1: LGTM

PR #594, opened 2023-10-13
DDFLSBP 202 brugeroprettelse fortryd knappen virker ikke

spaceo:
#### Link to issue

https://reload.atlassian.net/browse/DDFLSBP-202

#### Description
When a user clicks "Cancel" on the user registration form the user is logged out and redirected to the frontpage of the mounting application.

- Refactor menuLogOutUrl to logoutUrl
- Make the cancel button go to the logout path


Adamik10 (approved):
Good stuff :))
One question though - why don't we add the logout url to the urls configuration file?
Or is it already there?

PR #603, opened 2023-10-16
Fix/card clicks

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFLSBP-168

#### Description
Makes reservation material & stackable material card clickable. Before, the user would have to click on the card title, or an arrow that shows up on hover.

#### Screenshot of the result
n/a

#### Checklist

n/a


spaceo (approved):
LGTM 👍

PR #606, opened 2023-10-17
adds the `PromoBar` component and a new custom hook `useNotificationMessage` to the `PatronPage`

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFLSBP-84

#### Description

This Pull Request adds the `PromoBar` component and a new custom hook `useNotificationMessage` to the `PatronPage`. 

The `PromoBar` provides visual notifications to users, while the `useNotificationMessage` hook is designed to manage these notifications by handling auto-scrolling and setting removal timeouts.

#### Screenshot of the result

https://github.com/danskernesdigitalebibliotek/dpl-react/assets/49920322/a7fe635e-51c6-480e-b680-a1bd2121b735

### DIFF
--- a/src/apps/patron-page/PatronPage.dev.tsx
+++ b/src/apps/patron-page/PatronPage.dev.tsx
@@ -240,17 +240,17 @@ export default {
       defaultValue: "You used @this ebooks out of you quota of @that ebooks",
       control: { type: "text" }
     },
-    phoneInputMessageText: {
+    patronPagePhoneInputMessageText: {
       name: "Phone input validation message",
       defaultValue:
         "The phone number must be 6 to 15 characters in length and should be comprised solely of numbers or begin with a +",
       control: { type: "text" }
     },
-    handleResponseInformationText: {
+    patronPageHandleResponseInformationText: {
       defaultValue: "Your changes are saved.",
       control: { type: "text" }
     },
-    loadingText: {
+    patronPageLoadingText: {
       name: "Loading",
       defaultValue: "Loading..",
       control: { type: "text" }
--- a/src/apps/patron-page/PatronPage.dev.tsx
+++ b/src/apps/patron-page/PatronPage.dev.tsx
@@ -247,7 +247,7 @@ export default {
       control: { type: "text" }
     },
     handleResponseInformationText: {
-      defaultValue: "Dine ændringer er gemt.",
+      defaultValue: "Your changes are saved.",
       control: { type: "text" }
     },
     loadingText: {
--- a/src/apps/patron-page/PatronPage.dev.tsx
+++ b/src/apps/patron-page/PatronPage.dev.tsx
@@ -245,6 +245,15 @@ export default {
       defaultValue:
         "The phone number must be 6 to 15 characters in length and should be comprised solely of numbers or begin with a +",
       control: { type: "text" }
+    },
+    handleResponseInformationText: {
+      defaultValue: "Dine ændringer er gemt.",
+      control: { type: "text" }
+    },
+    loadingText: {
+      name: "Loading",
+      defaultValue: "Loading..",
+      control: { type: "text" }
     }
   }
 } as ComponentMeta<typeof PatronPage>;
--- a/src/core/utils/useNotificationMessage.ts
+++ b/src/core/utils/useNotificationMessage.ts
@@ -1,0 +1,32 @@
+import { useState } from "react";
+
+type UseNotificationOptionsType = {
+  timeout?: number;
+  scrollToTop?: boolean;
+};
+type UseNotificationReturnType = [string | null, (text: string) => void];
+
+export const useNotificationMessage = ({
+  timeout,
+  scrollToTop
+}: UseNotificationOptionsType): UseNotificationReturnType => {
+  const [notificationMessage, setNotificationMessage] = useState<string | null>(
+    null
+  );
+
+  const handleNotificationMessage = (text: string) => {
+    setNotificationMessage(text);
+    if (scrollToTop) {
+      window.scrollTo(0, 0);
+    }
+    if (timeout) {
+      setTimeout(() => {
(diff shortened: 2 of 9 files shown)


src/apps/patron-page/PatronPage.dev.tsx
```diff
@@ -245,6 +245,15 @@ export default {
...
+    },
+    handleResponseInformationText: {
+      defaultValue: "Dine ændringer er gemt.",
+      control: { type: "text" }
+    },
+    loadingText: {
+      name: "Loading",
+      defaultValue: "Loading..",
+      control: { type: "text" }
     }
```
spaceo, line 257: These keys should be prefixed with "patron-page-" or similar in order to prevent text key collisons

src/core/utils/useNotificationMessage.ts
```diff
@@ -0,0 +1,32 @@
+import { useState } from "react";
+
+type UseNotificationOptionsType = {
+  timeout?: number;
+  scrollToTop?: boolean;
+};
+type UseNotificationReturnType = [string | null, (text: string) => void];
+
+export const useNotificationMessage = ({
+  timeout,
```
spaceo, line 10: Great with the timeout. I like it!
But I think we should have a default so we do not need to think about what the timeout should be everytime we use the hook

spaceo: Same goes for scrollToTop. I think most of the time we want it to scroll to top? Or what do you think?

src/core/utils/useNotificationMessage.ts
```diff
@@ -0,0 +1,32 @@
...
+      window.scrollTo(0, 0);
+    }
+    if (timeout) {
+      setTimeout(() => {
+        setNotificationMessage(null);
+      }, timeout);
+    }
+  };
+
+  return [notificationMessage, handleNotificationMessage];
```
spaceo, line 29: I think you should return the PromoBar instead of the notificationMessage. Then we make the whole thing more declarative and nizzle.

spaceo (changes requested):
Great work!
I have a few comments

spaceo (approved):
LGTM 👍

src/core/utils/useNotificationMessage.tsx
```diff
@@ -0,0 +1,38 @@
...
+      }, timeout);
+    }
+  };
+
+  const NotificationComponent: React.FC = () => {
+    if (!notificationMessage) return null;
+    return <PromoBar text={notificationMessage} type="info" />;
+  };
+
+  return [NotificationComponent, handleNotificationMessage];
```
spaceo, line 35: I think it is nice to have components isolated in separate files. That also makes it easier for possible unittests/snapshots.
Also it should be possible to make the message a prop for NotificationComponent to make it pure and, again, easier to test :):
```typescript
interface NotificationComponentProps {
  message: string;
};

const const NotificationComponent: React.FC<NotificationComponentProps> = ({message}) => <PromoBar text={message} type="info" />;
```

And then here:
```typescript
  return [(<NotificationComponent message={notificationMessage} />), handleNotificationMessage];

Comment (spaceo) [reply src/core/utils/useNotificationMessage.tsx:35]: Did not test it 😄

**Review by spaceo — CHANGES_REQUESTED** [review]
Greatness. And the annoying reviewer has some more remarks

File: src/apps/patron-page/PatronPage.dev.tsx [hunk 5]
```diff
@@ -240,11 +240,20 @@ export default {
       defaultValue: "You used @this ebooks out of you quota of @that ebooks",
       control: { type: "text" }
     },
-    phoneInputMessageText: {
+    patronPagePhoneInputMessageText: {
```
spaceo, line 243: If you have refactored this data prop, what about dpl-cms then?

src/tests/unit/notification-message.test.tsx
```diff
@@ -26,22 +21,22 @@ const ComponentWithNotificationMessage = ({
...
+
+  it("should not display a message before the button is clicked", async () => {
+    const { getByTestId } = render(<ComponentWithNotificationMessage />);
+    const wrapper = getByTestId("wrapper");
 
+    // Assert that the wrapper does not contain the message initially
     expect(wrapper).toMatchInlineSnapshot(`
       <div
-        title="test-1"
+        data-testid="wrapper"
```
spaceo, line 48: It's fine to use data test id's to target the tested elements that is how it is done in other of the similar tests. But don't you get into the problem we saw with everything rendered into the same virtual dom and duplicate id's?

src/tests/unit/notification-message.test.tsx
```diff
@@ -30,6 +30,9 @@ describe("useNotificationMessage hook", () => {
     const { getByTestId } = render(<ComponentWithNotificationMessage />);
     const wrapper = getByTestId("wrapper");
 
+    // Expectations before the button is clicked
+    expect(screen.queryByText(/Some message/)).toBeNull(); // Expect the message to not be displayed
```
spaceo, line 43: Great that you start with asserting that there is nothing present! 👍

spaceo (changes requested):
Great additions!
Some comments need to be clarified though before approval

src/apps/patron-page/PatronPage.dev.tsx
kasperbirch1: it is fixed here: https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/427/files

src/tests/unit/notification-message.test.tsx
kasperbirch1: I don't think so after I added::
```
 afterEach(() => {
    cleanup();
  });
```

spaceo (approved):
LGTM 👍

PR #608, opened 2023-10-17
Revert "Update Install GHR command in release.yml so it uses the newest version

kasperbirch1:
#### Description

This reverts commit 20416d2a8e7d3deb93ad35e3fdc1d8327fbe8ea4.

This is because I encountered issues while creating tags/releases.


Adamik10 (approved):
Great!

PR #609, opened 2023-10-17
Disable background scrolling while any modal is open

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFLSBP-81

#### Description
This PR disables scrolling on the body element as long as any modals are open.

#### Screenshot of the result
n/a

#### Checklist

n/a


Adamik10:
Follow in #610

PR #610, opened 2023-10-17
Fix/scroll behind modals

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFLSBP-81

#### Description
This PR disables scrolling on the body element as long as any modals are open.
We also make sure that modals' z-index is higher than header's so that modals don't get obstructed.

#### Screenshot of the result
n/a

#### Checklist

n/a


spaceo (approved):
LGTM 👍

PR #632, opened 2023-10-23
Group parallel reservations

kasperg:
#### Link to issue

https://reload.atlassian.net/browse/DDFLSBP-79

#### Description

This change ensures that parallel reservations in FBS are displayed as a single reservation in the UI:

- The `reservationId` attribute for `ListType` has been replaced by a `reservationIds` attribute
- A new hook for retrieving reservation groups has been implemented. This supersedes the default `useGetReservationsV2()`. Instead of returning individual reservations it returns reservation groups which contain an array of reservation ids. Single reservations will have a single entry, parallel reservations will have multiple. The return value is mapped to the current `ListType` and `ReservationType` types.
- In general passing around strings as different forms of identifiers have been replaced by passing around business objects. As a mininum `ListType` but preferably `ReservationType` or `LoanType` where applicable.
- Components which previously took nullable `faust` (FBS) and `identifier` (Publizon) props now take a single `ListType` `item` prop
- Higher-order components `fetchDigitalMaterial()` and `fetchMaterial()` now work with `item` props
- States which previously used strings for ids, now use business objects

This change has lead to a number of related changes along the way:
- To have a stable approach for generating string ids for business objects I have created `itemId()`, `loanId()` and `reservationId()` functions. They can be used for `key` props and potentially `id` attributes.
- Test cases and fixtures have been updated with parallel reservations
- I have removed `it.only()` from a few test cases to be able to run all tests locally. I have `skip`'ed or fixed other cases in the same test which is now running.
- Helpers related to modals have been moved from `helpers/general` to `helpers/modal-helpers`
- Construction of modal id strings are delegated to functions related to the individual modal component which take relevant domain objects as arguments e.g. `deleteReservationModalId(ReservationType)`
- A few components have been updated to use the new domain hooks for loans and reservations
- Equality checks for business objects instead of primitives means that we have to use `isEqual()` from Lodash and `useDeepCompareEffect()` from `react-use`
- In an attempt to make Cypress tests less flaky I have added aliases and `wait()` in some tests. This is especially relevant for test cases which rely on other e.g. get the first list of reverations which can either be ready, physical or digital depending on what loads first.
- I merged in #577 to ease my development proces. We should consider if the comments there should be considered before merging this or these changes should be pulled from this PR.

- I have tried to add comments throughout the PR to explain what is going on
- I tried to enable Cypress Cloud during this PR ([0510d98](https://github.com/danskernesdigitalebibliotek/dpl-react/pull/632/commits/0510d98d784baeb5f70b3a5c0ef0ef7a0ffcd944)). I disabled it again to avoid getting to much in. You can see it in action in this PR.
- Codecov fails even though it should not. Sigh.


Taskfile.yml
kasperg, line 1: Changes here are a result of merging in #577.

src/apps/fee-list/list.tsx
```diff
@@ -29,6 +29,7 @@ const List: FC<ListProps> = ({
           {fees.map((itemData) => (
             <StackableFees
               amountOfMaterialsWithDueDate={itemData.materials.length}
+              item={{ faust: itemData.materials[0].recordId as FaustId }}
               faust={itemData.materials[0].recordId as FaustId}
```
kasperg, line 33: I did not updated fees.

They are FBS specific so still need a `faust` property. I have added `item` which is now needed for higher-order components to work.

src/apps/loan-list/list/loan-list.test.ts
```diff
@@ -615,7 +616,8 @@ describe("Loan list", () => {
     }).as("physical_loans");
   });
 
-  it("It opens details modal (digital loans)", () => {
+  // TODO This test hangs for whatever reason.
+  it.skip("It opens details modal (digital loans)", () => {
```
kasperg, line 620: I could not get this test to run.

It has previously been deactivated because this test uses `it.only` for another case. I opted for disabling it specifically to get the rest of them runnning.

src/apps/loan-list/list/loan-list.test.ts
```diff
@@ -481,7 +482,7 @@ describe("Loan list", () => {
       .find(".list-reservation")
       .eq(0)
       .find(".list-reservation__deadline p")
-      .should("have.text", "Due date 24-10-2022");
+      .should("have.text", "Due date 24-10-2022 08:32");
```
kasperg, line 485: This change it not related to parallel reservations but I removed `it.only` form another case in this test in order to run all tests.

To make tests pass here adding a time was necessary.

src/apps/menu/menu.test.ts
```diff
@@ -1,7 +1,6 @@
 import { TOKEN_LIBRARY_KEY } from "../../core/token";
 
-// todo will be readded in a future pr
-describe.skip("Menu (authenticated))", () => {
```
kasperg, line 4: This test seems to be working so I (re)enabled it.

src/apps/recommender/Recommender.tsx
```diff
@@ -1,62 +1,32 @@
...
-      );
-    }
-    if (fbsErrorReservations && !isSuccessFbsReservations) {
-      setPhysicalReservations([]);
+    fbs: {
+      reservations: physicalReservations,
+      isLoading: fbsReservationsLoading
     }
-  }, [fbsErrorReservations, fbsReservations, isSuccessFbsReservations]);
+  } = useReservations();
```
kasperg, line 29: This is basically a followup on #655 which uses the new hooks in the recommender component.

src/apps/reservation-list/list/reservation-list.tsx
```diff
@@ -45,273 +40,99 @@ const ReservationList: FC<ReservationListProps> = ({ pageSize }) => {
...
 
   const {
-    isSuccess: isSuccessPublizon,
-    data: publizonData,
-    isLoading: isLoadingPublizon
-  } = useGetV1UserReservations();
-
-  // State
-  const [readyForPickupReservationsFBS, setReadyForPickupReservationsFBS] =
-    useState<ReservationType[] | ReservationDetailsV2[] | null>(null);
```
kasperg, line 69: All this deletion is a result of using `useReservations()` and business objects leading to much fewer states and effects.

src/apps/reservation-list/modal/delete-reservation/delete-reservation-modal.tsx
```diff
@@ -32,12 +45,13 @@ const DeleteReservationModal: FC<DeleteReservationModalProps> = ({
   const removeSelectedReservations = () => {
     if (reservations.length > 0) {
       const reservationsToDelete = reservations
-        .map((id) => Number(isFaust(id)))
-        .filter((id) => id !== 0);
+        .filter(isPhysicalReservation)
+        .map(({ reservationIds }) => reservationIds)
+        .flat();
```
kasperg, line 50: This is they key moment of the actual change: Instead of assuming that there is one reservation id per reservation object we now have several per item. They still combine into a single array of ids to delete.

src/apps/reservation-list/modal/reservation-details/physical-list-details.tsx
```diff
@@ -94,16 +94,18 @@ const PhysicalListDetails: FC<PhysicalListDetailsProps & MaterialProps> = ({
...
       {
         data: {
-          reservations: [
-            {
-              expiryDate: selectedExpiryDate,
-              pickupBranch: selectedBranch,
-              reservationId
-            }
-          ]
+          reservations: reservationsChanges
```
kasperg, line 108: This is the second key change. Instead of applying changes to a single (physical) reservation id we now create an array of multiple reservations to update if this is a parallel reservation.

src/apps/reservation-list/modal/reservation-details/reservation-details.test.ts
```diff
@@ -466,6 +483,23 @@ describe("Reservation details modal test", () => {
...
+        JSON.stringify({
+          reservations: [
+            {
+              expiryDate: "2022-09-21",
+              // TODO: This should be DK-775120 as this is what the user
+              //  selected in previous steps but that does not seem to work
+              //  right now.
+              pickupBranch: "DK-775160",
+              reservationId: 4698559133
+            }
```
kasperg, line 497: As I see it i have identified a bug here. The reservation sent to FBS does not actually use the updated value. That should be handled in a separate issue though.

src/apps/reservation-list/utils/helpers.ts
```diff
@@ -13,24 +13,12 @@ export const sortByOldestPickupDeadline = (
...
-    .sort(
-      (objA, objB) =>
-        new Date(objA.pickupDeadline || new Date()).getTime() -
-          new Date(objB.pickupDeadline || new Date()).getTime() ||
-        (objA.title || "").localeCompare(objB.title || "")
-    );
-};
-
-export const getReservedPhysical = (list: ReservationType[]) => {
-  return list
```
kasperg, line 29: These functions were not used and did too many things filtering and sorting which are non-obvious from the function name.

src/core/modal.slice.ts
```diff
@@ -57,9 +61,9 @@ const modalSlice = createSlice({
           window.history.pushState(
             "",
             "",
-            `?modal=${alreadyOpenModals === null ? "" : alreadyOpenModals}${
-              action.payload.modalId
-            }`
+            appendQueryParametersToUrl(new URL(getCurrentLocation()), {
+              modal: `${alreadyOpenModals ?? ""}${action.payload.modalId}`
+            })
```
kasperg, line 66: This is an attempt at not replacing the entire url when opening a modal but instead only adding the new modal.

src/core/utils/types/reservation-type.ts
```diff
@@ -16,3 +16,43 @@ interface Reservation extends ListType {
...
+  }
+  return String(reservation.identifier);
+}
+
+export function isReservationType(item: ListType): item is ReservationType {
+  return (
+    !!item.identifier ||
+    (!!item.reservationIds && item.reservationIds.length > 0)
+  );
+}
```
kasperg, line 42: Look, a [type guard](https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards)!

src/core/utils/useGetReservationGroups.ts
```diff
@@ -0,0 +1,141 @@
...
+              [reservation.recordId]: reservation.reservationId
+            };
+          },
+          {}
+        )
+      };
+    }
+  );
+  return processedReserations;
+}
```
kasperg, line 57: This is also another key point of the PR: Reservations are grouped by transaction id to identify and create parallel reservations.

src/core/utils/useGetReservationGroups.ts
```diff
@@ -0,0 +1,141 @@
...
+        ...data
+      };
+    }
+
+    it("groups reservations by transaction id", () => {
+      const reservations = groupReservations([
+        generateReservation({ transactionId: "t1" }),
+        generateReservation({ transactionId: "t2" }),
+        generateReservation({ transactionId: "t2" }),
+        generateReservation({ transactionId: "t3" })
```
kasperg, line 106: This is an example of what I think we could achieve with something like fishery reload/dpl-react#47 - a generator/factory which ensures we have valid business objects and test cases which can focus on the attributes which are relevant to the case at hand.

src/core/utils/types/reservation-type.ts
```diff
@@ -16,3 +16,29 @@ interface Reservation extends ListType {
...
+
+export type PhysicalReservationType = ReservationType &
+  NonNullableFields<Required<Pick<ListType, "faust" | "reservationIds">>>;
+
+export type DigitalReservationType = ReservationType &
+  NonNullableFields<Required<Pick<ListType, "identifier" | "reservationId">>>;
+
+export function isPhysicalReservation(
+  reservation: ReservationType
+): reservation is PhysicalReservationType {
```
spaceo, line 46: Awesome! :)

src/core/utils/types/loan-type.ts
```diff
@@ -13,3 +13,7 @@ export interface LoanType extends ListType {
 export function isLoanType(item: ListType): item is LoanType {
   return !!item.loanId;
 }
+
+export function loanId(loan: LoanType): string {
```
spaceo, line 17: Wouldn't be an idea to create a `LoanId`type?

src/apps/reservation-list/modal/reservation-details/reservation-details.tsx
```diff
@@ -58,39 +61,43 @@ const ReservationDetails: FC<ReservationDetailsProps & MaterialProps> = ({
...
-          {reservation.reservationId && allowUserRemoveReadyReservations && (
-            <ReservationDetailsButton
-              classNames="modal-details__buttons--hide-on-mobile"
-              openReservationDeleteModal={openReservationDeleteModal}
-              reservation={reservation}
-              numberInQueue={numberInQueue}
-            />
-          )}
-          {isDigital && reservation.identifier && (
+          {isPhysicalReservation(reservation) &&
```
spaceo, line 64: I like!

src/core/utils/types/reservation-type.ts
```diff
@@ -27,6 +27,21 @@ export type PhysicalReservationType = ReservationType &
 export type DigitalReservationType = ReservationType &
   NonNullableFields<Required<Pick<ListType, "identifier" | "reservationId">>>;
 
+export function reservationId(reservation: ReservationType): string {
```
spaceo, line 30: A `ReservationId`type as return type?

src/apps/reservation-list/modal/delete-reservation/delete-reservation-modal.tsx
```diff
@@ -13,15 +13,23 @@ import {
...
   ReservationType
 } from "../../../../core/utils/types/reservation-type";
+import { getModalIds } from "../../../../core/utils/helpers/modal-helpers";
 
 interface DeleteReservationModalProps {
   modalId: string;
   reservations: ReservationType[];
 }
 
+export function deleteReservationModalId(reservation: ReservationType): string {
```
spaceo, line 27: A `ModalId`as return type?

src/apps/reservation-list/list/DisplayedReservations.tsx
```diff
@@ -18,49 +18,39 @@ const DisplayedReservations: FC<DisplayedReservationsProps> = ({
...
-        <List
-          openReservationDetailsModal={openReservationDetailsModal}
-          pageSize={pageSize}
-          header={t("reservationListDigitalReservationsHeaderText")}
-          emptyListDataCy="reservation-list-digital-reservations-empty-list"
-          reservations={reservedReservationsPublizon}
-          emptyListLabel={t("reservationListDigitalReservationsEmptyText")}
-        />
-      )}
+      <List
```
spaceo, line 29: This is such a thrill!

src/apps/reservation-list/modal/reservation-details/reservation-details.test.ts
spaceo: Great. Did you already create a task for it?

spaceo (approved):
Great. What a load of work you have done here.
I have some minor comments, mostly about types.
It's great to see that several places in the application have been simplified because of the new domain objects and hooks!

src/core/utils/useGetReservationGroups.ts
```diff
@@ -0,0 +1,158 @@
...
+          setSuccess(true);
+        }
+      })
+      .catch((errorResponse) => {
+        setError(errorResponse);
+        setLoading(false);
+      });
+  }, [setLoading, setError, setReservations]);
+
+  return { isLoading: loading, isSuccess: success, error, data: reservations };
```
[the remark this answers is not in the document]
kasperg: I think this can be resolved now that the new hook implementation is merely a wrapper around the existing hook.

src/apps/loan-list/modal/material-details-modal.tsx
```diff
@@ -1,12 +1,32 @@
...
 
+export function reservationDetailsModalId(
+  reservation: ReservationType
+): string {
+  const prefix = String(getModalIds().reservationDetails);
+  const fragment = String(reservation.identifier || reservation.faust);
+  return constructModalId(prefix, [fragment]);
+}
+
+export function loanDetailsModalId(loan: LoanType) {
```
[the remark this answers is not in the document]
kasperg: My original intent was to add a `*ModalId` function for each modal, but `MaterialDetailsModal` is used with child components: Currently `MaterialDetails` and `ReservationDetails`. As the two child components do not have anything that rhymes with `Modal` in their component name I opted for adding these functions with the parent modal.

I can see why this is confusing and I am open to suggestions for how to address this better. One approach might to be add a comment explaining the situation above.

PR #633, opened 2023-10-23
Defaults in dpl-react configuration

spaceo:
#### Link to issue

https://reload.atlassian.net/browse/DDFLSBP-246

#### Description

This PR is created as a port of the work being done on this [PR](https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/436) in dpl-cms.
Initiatives in this project:
* Changed allowRemoveReadyReservations to be boolean property. I think we should handle booleans as a part of a json structure config. In that way it is "natively boolean" and we do not have to juggle around with tranforming "1" and "0" or "true" and "false".
* Instant loan config. The place it is used there was no checks if the various config properties were set. I tried to accomodate the situation by let the handling functions accept null parameters where they were used. Not optional though would like to handle it earlier.
* In the work with the allowRemoveReadyReservations boolean I wanted to be sure of how the config handler works with the different input. So I wrote a pair of unit tests to document it.

This PR has precedence over https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/436.

### DIFF
--- a/src/components/reservation/helper.ts
+++ b/src/components/reservation/helper.ts
@@ -216,7 +216,7 @@ export const consolidatedHoldings = (branchHoldings: HoldingsV3[]) => {
 export const getInstantLoanBranchHoldings = (
   branchHoldings: HoldingsV3[],
   whitelist: AgencyBranch[],
-  instantLoanStrings: string[] | null
+  instantLoanStrings: string[]
 ) => {
   const whitelistBranchIds = whitelist.map(({ branchId }) => branchId);
   // 1. Filter holdings by branch on whitelist
@@ -230,7 +230,7 @@ export const getInstantLoanBranchHoldings = (
         // if a material group description contains any of the instant loan strings
         // and is available, it is an instant loan.
         return (
-          (instantLoanStrings ?? []).some((instantLoanString) => {
+          instantLoanStrings.some((instantLoanString) => {
             return materialGroup.description?.includes(instantLoanString);
           }) && available
         );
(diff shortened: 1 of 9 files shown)


src/tests/unit/config.test.tsx
```diff
@@ -0,0 +1,72 @@
+import { configureStore, combineReducers } from "@reduxjs/toolkit";
```
kasperg, line 1: Good stuff adding tests.

I wonder if there is a reason why this does not use in file testing?

src/components/reservation/helper.ts
```diff
@@ -216,7 +216,7 @@ export const consolidatedHoldings = (branchHoldings: HoldingsV3[]) => {
 export const getInstantLoanBranchHoldings = (
   branchHoldings: HoldingsV3[],
   whitelist: AgencyBranch[],
-  instantLoanStrings: string[]
```
kasperg, line 219: I am not too keen on this change. As I see it it goes against [our decision](https://github.com/danskernesdigitalebibliotek/dpl-react/blob/develop/docs/architecture/adr-007-configuration.md) to make the mouting application provide configuration in the expected format.

If no strings are provided then we should at least be able to set a `{enabled: false}`. I guess this goes against the design in danskernesdigitalebibliotek/dpl-cms#436 where modules can register their own settings. 🤔 

If we have become wiser can we do something do update the ADR accordingly?

kasperg (approved):
:+1: but with a few considerations for you.

src/tests/unit/config.test.tsx
spaceo: Becase "@testing-library/react-hooks" is a dev dependency and the config.ts file is a production file

PR #649, opened 2023-10-31
Set the patron `preferredPickupBranch` instead of the empty string in `openOrder`

kasperbirch1:
#### Link to issue

#### Description

The previous implementation set an empty string for the `pickUpBranch` field in `openOrder` if `selectedBranch` was not provided. However, an empty string is still considered a valid value when sent to the API, which triggered a `UNKNOWN_USER` error due to the user not being found.

Additionally, the `getPatronAddress` function has been removed as it's no longer necessary to send the address. There was a concern regarding potential mismatches in address formatting, which could result in errors since `openOrder` requires exact matches for input values.

### DIFF
--- a/src/components/reservation/ReservationModalBody.tsx
+++ b/src/components/reservation/ReservationModalBody.tsx
@@ -7,8 +7,7 @@ import {
   getMaterialTypes,
   getManifestationType,
   materialIsFiction,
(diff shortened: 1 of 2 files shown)


spaceo (approved):
LGTM 👍

PR #650, opened 2023-11-01
Add disabled state for delete materials button

JacobArrow:
#### Link to issue
https://reload.atlassian.net/browse/DDFLSBP-171

#### Description
A user should no be able to trigger the delete materials modal, if no materials have been selected. This adds a simple check for that.

#### Screenshot of the result
image
image


kasperg (approved):
:+1: nice and simple.

PR #651, opened 2023-11-01
Add aria label to link

JacobArrow:
#### Link to issue
https://reload.atlassian.net/browse/DDFLSBP-180

#### Description
We want screen readers to read the number of reserved items. Because the number is not a semantic part of the a tag, the screen reader will ignore it.
An aria-label can solve this, but a more correct solution would be to move the span inside the a tag - which would require changes in design system aswell.

### DIFF
--- a/src/apps/menu/menu-navigation-list/MenuNavigationItem.tsx
+++ b/src/apps/menu/menu-navigation-list/MenuNavigationItem.tsx
@@ -37,7 +37,11 @@ const MenuNavigationItem: FC<MenuNavigationItemProps> = ({
   return (
     <li className="link-filters mb-16">
       <div className="link-filters__tag-wrapper">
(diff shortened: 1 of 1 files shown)


kasperg (changes requested):
I do not understand the implementation here. I appreciated your comment

> An aria-label can solve this, but a more correct solution would be to move the span inside the a tag.

I agree but I cannot tell from the comment why you did not update the implementation to move the number already there into the tag.

If this is very complicated then I would suggest adding `aria-hidden="true"` to the following span. This way we avoid having screen readers read the number twice in some situations.

In general there might be an issue around the actual wording of the labels. Is it "Reservations 5" or "5 reservations"... and what about "1 reservation" etc. I guess that is outside the scope of this PR.

spaceo (approved):
If this makes the screen reader read the output properly, I'm happy. LGTM 👍

PR #652, opened 2023-11-01
Close reservations modal when deleting reservation

JacobArrow:
#### Link to issue
https://reload.atlassian.net/browse/DDFLSBP-102

#### Description
Upon deleting a reserved material, the user can be stuck in an infinite loop, because the underlying reservations modal is not closed after deleting a reservation. This change always closes the reservationsQueued modal when deleting a reserved material.

### DIFF
--- a/src/apps/reservation-list/modal/delete-reservation/delete-reservation-modal.tsx
+++ b/src/apps/reservation-list/modal/delete-reservation/delete-reservation-modal.tsx
@@ -12,7 +12,6 @@ import {
   useDeleteV1UserReservationsIdentifier
 } from "../../../../core/publizon/publizon";
 import { isFaust, isIdentifier } from "../../../dashboard/util/helpers";
-import { getModalIds } from "../../../../core/utils/helpers/general";
 
 interface DeleteReservationModalProps {
   modalId: string;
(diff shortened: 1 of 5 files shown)


src/apps/reservation-list/modal/delete-reservation/delete-reservation-modal.tsx
```diff
@@ -65,6 +67,7 @@ const DeleteReservationModal: FC<DeleteReservationModalProps> = ({
         )
       );
 
+      close(reservationsQueued as string);
```
JacobArrow, line 70: I am a bit unsure if this is too closely coupled?
Another option would be to pass in the underlying `modalId` we want closed with itself. 
Thoughts?

spaceo:
We agreed on waiting with this until some global handling of closing existing modals have been implemented

src/tests/unit/modal.test.tsx
JacobArrow, line 1: This test is a bit redundant - maybe a bit overkill
Thoughts?

src/core/utils/modal.tsx
```diff
@@ -66,7 +66,12 @@ function Modal({
   };
 
   return (
-    <FocusTrap>
+    <FocusTrap
+      focusTrapOptions={{
+        // set fallbackFocus when running vitest to avoid focus trap errors
```
spaceo, line 71: This is a minor, but we usually format our comments with the first letter in uppercase `// Set fallbackFocus` and end the comment with a period `.`

spaceo (approved):
One comment about the formatting of a comment but otherwise LGTM 👍

PR #655, opened 2023-11-02
Introduce and implement `useReservations` and `useLoans` hooks

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFLSBP-197
https://reload.atlassian.net/browse/DDFLSBP-198
https://reload.atlassian.net/browse/DDFLSBP-233
https://reload.atlassian.net/browse/DDFLSBP-157

#### Description

This Pull Request addresses issues pertaining to the display of reservations and loans from both FBS and Publizon systems. It introduces two new hooks: `useReservations` and `useLoans` and the implement of them

These new hooks simplify interactions with both FBS and Publizon systems by consolidating reservations and loans from each provider, respectively. Additionally, they return other specific lists of loans and reservations as required for the UI.

The changes have been integrated in sections relevant to loans and reservations.

All changes have been consolidated into a single commit to mitigate potential merge conflicts, especially with Kasper Garnes who has been working on the same files.

#### Screenshot of the result

https://github.com/danskernesdigitalebibliotek/dpl-react/assets/49920322/a86c5aac-da47-415e-ba2f-a4952f7bc051

image

https://github.com/danskernesdigitalebibliotek/dpl-react/assets/49920322/cc2db61c-8ddd-4534-8fe8-e16b5b19156b

### DIFF
--- a/src/apps/loan-list/list/loan-list.tsx
+++ b/src/apps/loan-list/list/loan-list.tsx
@@ -57,8 +57,8 @@ const LoanList: FC<LoanListProps> = ({ pageSize }) => {
   const [accepted, setAccepted] = useState<boolean>(false);
   const [modalDetailsId, setModalDetailsId] = useState<string | null>(null);
   const {
-    sortedByLoanDateFbs,
-    sortedByLoanDatePublizon,
+    loansSortedByDateFbs,
+    loansSortedByDatePublizon,
     stackedMaterialsDueDatesFbs
   } = useLoans();
   useEffect(() => {
@@ -67,20 +67,20 @@ const LoanList: FC<LoanListProps> = ({ pageSize }) => {
       return;
     }
     let loanForModal = null;
-    if (sortedByLoanDateFbs && modalDetailsId) {
+    if (loansSortedByDateFbs && modalDetailsId) {
       loanForModal = getFromListByKey(
-        sortedByLoanDateFbs,
+        loansSortedByDateFbs,
         "loanId",
         modalDetailsId
       );
     }
     if (
       loanForModal?.length === 0 &&
-      sortedByLoanDatePublizon &&
+      loansSortedByDatePublizon &&
       modalDetailsId
     ) {
       loanForModal = getFromListByKey(
-        sortedByLoanDatePublizon,
+        loansSortedByDatePublizon,
         "identifier",
         modalDetailsId
       );
@@ -91,8 +91,8 @@ const LoanList: FC<LoanListProps> = ({ pageSize }) => {
   }, [
     modalDetailsId,
     modalLoan,
-    sortedByLoanDateFbs,
-    sortedByLoanDatePublizon
+    loansSortedByDateFbs,
+    loansSortedByDatePublizon
   ]);
 
   const openAcceptModal = useCallback(() => {
@@ -142,9 +142,9 @@ const LoanList: FC<LoanListProps> = ({ pageSize }) => {
   }, [allLoansId, loanDetails, openDueDateModal]);
 
   const listContainsLoans =
-    (Array.isArray(sortedByLoanDateFbs) && sortedByLoanDateFbs.length > 0) ||
-    (Array.isArray(sortedByLoanDatePublizon) &&
-      sortedByLoanDatePublizon.length > 0);
+    (Array.isArray(loansSortedByDateFbs) && loansSortedByDateFbs.length > 0) ||
+    (Array.isArray(loansSortedByDatePublizon) &&
+      loansSortedByDatePublizon.length > 0);
 
   const resetAccepted = () => {
     setAccepted(false);
@@ -156,11 +156,11 @@ const LoanList: FC<LoanListProps> = ({ pageSize }) => {
         <h1 className="text-header-h1 my-32">{t("loanListTitleText")}</h1>
         {listContainsLoans && (
           <>
-            {sortedByLoanDateFbs && (
+            {loansSortedByDateFbs && (
               <List
                 pageSize={pageSize}
                 emptyListLabel={t("loanListPhysicalLoansEmptyListText")}
-                loans={sortedByLoanDateFbs}
+                loans={loansSortedByDateFbs}
                 dueDates={stackedMaterialsDueDatesFbs}
                 view={view}
                 openLoanDetailsModal={openLoanDetailsModal}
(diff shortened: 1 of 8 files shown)


src/apps/loan-list/list/loan-list.tsx
```diff
@@ -63,59 +56,44 @@ const LoanList: FC<LoanListProps> = ({ pageSize }) => {
...
-  const [physicalLoansDueDates, setPhysicalLoansDueDates] = useState<string[]>(
-    []
-  );
-  const { isSuccess, data } = useGetLoansV2();
-  const { data: publizonData } = useGetV1UserLoans();
+  const {
+    sortedByLoanDateFbs,
+    sortedByLoanDatePublizon,
+    stackedMaterialsDueDatesFbs
+  } = useLoans();
```
spaceo, line 63: I think it would be more nice with a return type that is something like: 
```typescript
type UseLoanProvider = {
  loansSortedByDate: Loan[];
  stackedMaterialsDueDates: Something[];
};

type UseLoans = {
  fbs: UseLoanProvider;
  publizon: UseLoanProvider;
};
```
Does it make sense?

src/apps/menu/menu-logged-in/MenuLoggedInContent.tsx
```diff
@@ -117,9 +52,9 @@ const MenuLoggedInContent: FC<MenuLoggedInContentProps> = ({ pageSize }) => {
   }, [fbsFees]);
 
   const showNotifications =
-    loansOverdue !== 0 ||
-    loansSoonOverdue !== 0 ||
-    reservationsReadyForPickup !== 0;
+    allOverdueLoans.length !== 0 ||
```
spaceo, line 55: all... :)

src/core/utils/useLoans.tsx
```diff
@@ -0,0 +1,88 @@
...
+  const stackedMaterialsDueDatesFbs = getDueDatesLoan(mappedLoansFbs);
+  return {
+    allLoans,
+    sortedByLoanDateFbs,
+    sortedByLoanDatePublizon,
+    allOverdueLoans,
+    allSoonOverdueLoans,
+    allFarFromOverdueLoans,
+    stackedMaterialsDueDatesFbs
+  };
```
spaceo, line 85: Again I think the return shape could:
* Skip the all prefixes
* Group provider related loans like described earlier

spaceo: How about loading state and error handling? If the hook does not handle those matters (which it looks like it doesn't) at least it should unify the error and loading states for the request and expose them in the product of the function.

src/core/utils/useReservations.tsx
```diff
@@ -0,0 +1,59 @@
...
+  ];
+
+  return {
+    allReservations,
+    readyToLoanReservationsFBS,
+    readyToLoanReservationsPublizon,
+    allReadyToLoanReservations,
+    allQueuedReservations,
+    queuedFBSReservations,
+    queuedPublizonReservations
```
spaceo, line 55: I have the same comments here as in the other hook.

src/core/utils/useReservations.tsx
```diff
@@ -0,0 +1,59 @@
...
+  };
```
spaceo, line 56: I really think we should have at least a snapshot test for both hooks where we mock the response from fbs/publizon.

spaceo (changes requested):
Really nice improvement.
I have some comments I would like you to look at.

src/core/utils/useLoans.tsx
```diff
@@ -0,0 +1,138 @@
...
+  farFromOverdue: LoanType[];
+  isLoading: boolean;
+  isError: boolean;
+  sortedByDate?: LoanType[];
+  stackedMaterialsDueDates?: string[];
+};
+
+type UseLoansType = {
+  all: Loans & {
+    loans: LoanType[];
```
kasperg, line 30: Consider moving `loans` to the `Loans` type. 

I wonder why we do not think it is relevant to return all loans from either `fbs` or `publizon` going forward.

src/core/utils/useLoans.tsx
```diff
@@ -0,0 +1,138 @@
...
+  // The stacked materials view shows materials stacked by
+  // due date, and for this we need a unique list of due dates
+  const loansSortedByDateFbs = sortByDueDate(mappedLoansFbs);
+  const loansSortedByDatePublizon = sortByDueDate(mappedLoansPublizon);
+
+  // list of all due dates used for the stacked materials
+  const stackedMaterialsDueDatesFbs = getDueDatesLoan(mappedLoansFbs);
+  return {
+    all: {
+      loans,
```
kasperg, line 111: Consider dropping the `loans` property, make the `sortedByDueDate` property required and then assign `loans` to `all.sortedByDueDate`.

1. As far as I can tell `loans` is already sorted on L69.
2. Required attributes should be preferred to optional
3. It removed the special case of of the `loans` property on `all`.

src/apps/loan-list/list/loan-list.tsx
```diff
@@ -63,59 +56,46 @@ const LoanList: FC<LoanListProps> = ({ pageSize }) => {
...
-  );
-  const { isSuccess, data } = useGetLoansV2();
-  const { data: publizonData } = useGetV1UserLoans();
+  const {
+    fbs: {
+      sortedByDate: loansSortedByDateFbs,
+      stackedMaterialsDueDates: stackedMaterialsDueDatesFbs
+    },
+    publizon: { sortedByDate: loansSortedByDatePublizon }
+  } = useLoans();
```
spaceo, line 65: I think you should make aliases like this:
```
const {
  fbs: loansPhysical,
  publizon: loansDigital
}

File: src/core/utils/useReservations.tsx [hunk 9]
```diff
@@ -0,0 +1,100 @@
...
+type Reservations = {
+  readyToLoan: ReservationType[];
+  queued: ReservationType[];
+  isLoading: boolean;
+  isError: boolean;
+};
+
+type UseReservationsType = {
+  all: Reservations & {
+    reservations: ReservationType[];
```
kasperg, line 20: Consider adding `reservations` to the `Reservations` type.

1. This makes the types consistent.
2. As far as I can tell there is no way to retrieve all reservations from `fbs` or `publizon`. I cannot tell why we would not want that.

src/core/utils/useReservations.tsx
```diff
@@ -0,0 +1,100 @@
+import { useGetReservationsV2 } from "../fbs/fbs";
+import { useGetV1UserReservations } from "../publizon/publizon";
+import { getQueuedReservations } from "./helpers/general";
```
kasperg, line 3: Please move this function from the general helpers to this file.

I do not see a reason why developers should use it directly.

spaceo (approved):
LGTM 👍

PR #664, opened 2023-11-06
Fix/ereol pickup indication

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFLSBP-213

#### Description
The pickup indication wasn't present for online materials on the user reservations list. A physical address was shown for physical materials, but there was nothing at all for the digital ones. Now we have a generic translatable text for these.
 
#### Screenshot of the result
image

n/a

### DIFF
--- a/src/apps/reservation-list/reservation-material/reservation-info.tsx
+++ b/src/apps/reservation-list/reservation-material/reservation-info.tsx
@@ -81,7 +81,10 @@ const ReservationInfo: FC<ReservationInfoProps> = ({
         color={success as string}
         percent={100}
         info={getInfo()}
-        label={[pickupLibrary, pickupNumber || ""]}
+        label={[
+          pickupLibrary,
+          pickupNumber || t("reservationListDigitalPickupText")
+        ]}
         reservationInfo={reservationInfo}
         openReservationDetailsModal={openReservationDetailsModal}
         empty={!showStatusCircleIcon}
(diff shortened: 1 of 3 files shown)


src/apps/reservation-list/reservation-material/reservation-info.tsx
```diff
@@ -81,7 +81,10 @@ const ReservationInfo: FC<ReservationInfoProps> = ({
         color={success as string}
         percent={100}
         info={getInfo()}
-        label={[pickupLibrary, pickupNumber || ""]}
+        label={[
+          pickupLibrary,
+          pickupNumber || t("reservationListDigitalPickupText")
+        ]}
```
JacobArrow, line 87: Why is it pickupNumber that is replaced with the `reservationListDigitalPickupText`?
Do we expect pickupLibrary to be `undefined`?
What if its not?
Could we end up in a case where it displays som something like:
**Hovedbibliotek
Online access**

We should have an identifier for when a material is digital and replace the whole array with `reservationListDigitalPickupText`

Adamik10: Inside the `reservation-status.tsx` component Sine already relied on it not being defined for one of the use cases. So for the sake of minimizing the work necessary to get this out, I ran with that too. 
I have no reason to believe it's wrong. And although I totally see your point, if the scenario you're describing is true, we would need to change some more code connected to this other than my munniscule addition :)

JacobArrow: Thank you for the clarification. I just needed to make sure that it was not an oversight :)

JacobArrow (approved):
Great work 🔥

PR #666, opened 2023-11-08
Move logic for display `MaterialButtonReservableFromAnotherLibrary`

kasperbirch1:
#### Description

Move logic for display `MaterialButtonReservableFromAnotherLibrary`

If the material can only be reserved on another library, we don't want to show other buttons like `MaterialButtonFindOnShelf`

#### Screenshot of the result
image


JacobArrow (approved):
Nice addition! 🔥

PR #669, opened 2023-11-09
Refactoring thresholdConfig => expirationWarningDaysBeforeConfig

spaceo:
#### Link to issue

https://reload.atlassian.net/browse/DDFSOEG-583

#### Description

Refactoring thresholdConfig => expirationWarningDaysBeforeConfig
Because it has been changed in the cms.
Also adding test and a hook for getting the loan thresholds.

#### Screenshot of the result

image

### DIFF
--- a/src/apps/dashboard/dashboard.dev.tsx
+++ b/src/apps/dashboard/dashboard.dev.tsx
@@ -156,10 +156,8 @@ export default {
       defaultValue: "Borrow before @date",
       control: { type: "text" }
     },
-    // Config
-    thresholdConfig: {
-      defaultValue:
-        '{\n      "colorThresholds":{\n      "danger":"0",\n      "warning":"6"\n   }\n   }',
+    expirationWarningDaysBeforeConfig: {
+      defaultValue: "6",
       control: { type: "text" }
     }
   },
--- a/src/tests/unit/status-badge.test.tsx
+++ b/src/tests/unit/status-badge.test.tsx
@@ -1,0 +1,170 @@
+import { configureStore, combineReducers } from "@reduxjs/toolkit";
+import React, { ReactNode } from "react";
+import { Provider } from "react-redux";
+import { afterEach, beforeEach, describe, expect, test, vi } from "vitest";
+import { cleanup, render } from "@testing-library/react";
+import configReducer from "../../core/config.slice";
+import StatusBadge from "../../apps/loan-list/materials/utils/status-badge";
+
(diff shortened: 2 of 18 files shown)


src/tests/unit/status-badge.test.tsx
```diff
@@ -0,0 +1,170 @@
...
+
+  test("Expiring soon warning if within threshold", () => {
+    const { getByTestId } = render(
+      <Wrapper>
+        <div data-testid="badge-wrapper">
+          <StatusBadge
+            showBadgeWithDueDate
+            // Since the due date is tomorrow and the warning threshold is set to within 6 days
+            // we should get a warning.
+            badgeDate="2023-11-10 12:00:00+0100"
```
kasperg, line 63: :+1: I appreciate this comment.

One might argue that you could make relationship in code calculate the date as `date` - 1 day but all these programmatically set values can also make it harder to read the test case. Having the comment helps me understand the relationship in text.

src/apps/dashboard/dashboard.dev.tsx
```diff
@@ -156,10 +156,8 @@ export default {
       defaultValue: "Borrow before @date",
       control: { type: "text" }
     },
-    // Config
-    thresholdConfig: {
-      defaultValue:
-        '{\n      "colorThresholds":{\n      "danger":"0",\n      "warning":"6"\n   }\n   }',
```
kasperg, line 162: [Some places it seems as if we move from primitive values to JSON objects even for single keys](https://github.com/danskernesdigitalebibliotek/dpl-react/pull/633/files#diff-0c0b0afee3b4a39016462d76c44101e19d81c3e17506f2e629e366c88cfc7b3fL31).

This takes a JSON object which now only has a single key and reduces it to a primitive value. 

Do you have any thoughts about in which situations one should use one approach over the other?

kasperg (approved):
:+1: LGTM and appreciate the test.

I have a couple of things for you to consider before you merge.

src/tests/unit/status-badge.test.tsx
spaceo: 👍

src/apps/dashboard/dashboard.dev.tsx
spaceo: I agree. It is inconsistent. In the future we should always handle config as JSON objects. In that way we can deal with the types as boolean and number natively.
I'lla add a comment about it.

PR #670, opened 2023-11-10
Fix `ReferenceError: process is not defined`

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFSOEG-585

#### Description

Added checks to ensure safe access to `process.env`, addressing environments where it might be undefined.

### DIFF
--- a/src/core/utils/helpers/vitest.ts
+++ b/src/core/utils/helpers/vitest.ts
@@ -1,0 +1,4 @@
+export const isVitestEnvironment =
+  typeof process !== "undefined" && process.env && process.env.VITEST;
+
+export default {};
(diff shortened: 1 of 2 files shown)


src/core/utils/helpers/vitest.ts
```diff
@@ -0,0 +1,4 @@
+export const isVitestEnvironment =
+  typeof process !== "undefined" && process.env && process.env.VITEST;
```
spaceo, line 2: @kasperbirch1 We need to do it differently and add some extra functionality to the webpack configuration. But let's talk about it, that's easier :)

spaceo (changes requested):
We need to talk :)

spaceo (approved):
LGTM 👍

PR #671, opened 2023-11-10
Fix/advanced search cql state

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DSC-77
https://reload.atlassian.net/browse/DSC-78

#### Description
This PR fixes the root problem of two bugs, where after performing an initial search in the advanced search app, the consequent searches would always use the initial cql query for searching without adjusting based on user input.

During this work, a new bug was discovered, that this PR also addresses - when removing multiselect dropdown items the external state wasn't being updated. 
Example:
- user wrote "Harry potter" into the text field on Advanced search page
- user chose "book" and "e-book" as filters from multiselect dropdown
- user then decided to remove "book" from filters
- -> this change wasn't being reflected in the "translated CQL" preview field, or upon consequent search

#### Screenshot of the result
n/a

n/a

### DIFF
--- a/src/apps/advanced-search/CqlSearchHeader.tsx
+++ b/src/apps/advanced-search/CqlSearchHeader.tsx
@@ -18,7 +18,7 @@ const CqlSearchHeader: React.FC<CqlSearchHeaderProps> = ({
     if (initialCql.trim() !== "") {
       setCql(initialCql);
     }
-  });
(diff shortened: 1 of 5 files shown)


spaceo (approved):
LGTM 👍 
Great with the test 🎉

PR #680, opened 2023-11-15
Refactor reservation pause banner

JacobArrow:
#### Link to issue
https://reload.atlassian.net/browse/DDFLSBP-70

#### Description
Previously it was not clear if a pause was active or not, and a toggle button was used to open a modal for changing settings.
This PR adds some descriptive text about the state of the current pause and a "Settings" button that behaves as expected.

#### Screenshot of the result
image
image

PR #681, opened 2023-11-15
Added ability to cancel pause period

JacobArrow:
#### Link to issue
https://reload.atlassian.net/browse/DDFLSBP-209

#### Description
This adds the ability to cancel an ongoing pause period after it has been set. 
The issue suggests, that the user should be prompted with a success message, but when #680 is introduced, the banner will update according to the current pause settings.

#### Screenshot of the result when #680 is also merged in
https://github.com/danskernesdigitalebibliotek/dpl-react/assets/14014012/ca7908a8-b2c0-4368-bd5b-1527a9685bf7

### DIFF
--- a/src/apps/patron-page/PatronPage.dev.tsx
+++ b/src/apps/patron-page/PatronPage.dev.tsx
@@ -85,6 +85,10 @@ export default {
       defaultValue: "Save",
       control: { type: "text" }
     },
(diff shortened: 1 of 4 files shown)


kasperbirch1 (approved):
Nice that the code is more readable now. LGTM 👍

PR #682, opened 2023-11-15
Show "online access" label for all digital materials

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFLSBP-213

#### Description
This PR introduces a fix so that all digital reservations get the correct label, instead of the physical reservation address.

#### Screenshot of the result
n/a

n/a

### DIFF
--- a/src/apps/reservation-list/reservation-material/reservation-info.tsx
+++ b/src/apps/reservation-list/reservation-material/reservation-info.tsx
@@ -82,9 +82,9 @@ const ReservationInfo: FC<ReservationInfoProps> = ({
         percent={100}
         info={getInfo()}
         label={
-          !isDigital
-            ? [pickupLibrary, pickupNumber || ""]
-            : [t("reservationListDigitalPickupText")]
+          isDigital
+            ? [t("reservationListDigitalPickupText")]
+            : [pickupLibrary, pickupNumber || ""]
         }
         reservationInfo={reservationInfo}
         openReservationDetailsModal={openReservationDetailsModal}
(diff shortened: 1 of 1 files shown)


src/apps/reservation-list/reservation-material/reservation-info.tsx
```diff
@@ -81,10 +81,11 @@ const ReservationInfo: FC<ReservationInfoProps> = ({
...
         info={getInfo()}
-        label={[
-          pickupLibrary,
-          pickupNumber || t("reservationListDigitalPickupText")
-        ]}
+        label={
+          !isDigital
+            ? [pickupLibrary, pickupNumber || ""]
+            : [t("reservationListDigitalPickupText")]
+        }
```
JacobArrow, line 88: I guess my speculations [here](https://github.com/danskernesdigitalebibliotek/dpl-react/pull/664/files#r1387828611) turned out to be true 😅

src/apps/reservation-list/reservation-material/reservation-info.tsx
```diff
@@ -81,10 +81,11 @@ const ReservationInfo: FC<ReservationInfoProps> = ({
         color={success as string}
         percent={100}
         info={getInfo()}
-        label={[
-          pickupLibrary,
-          pickupNumber || t("reservationListDigitalPickupText")
-        ]}
+        label={
+          !isDigital
```
JacobArrow, line 85: I know it is a small thing, but I would prefer if we removed the `not` check and just switched the two arrays around.

JacobArrow (approved):
This is a very nice solution and it is much easyer to understand what is happening.
Great job 🔥

PR #686, opened 2023-11-16
Set default start pause date as today

JacobArrow:
#### Link to issue
https://reload.atlassian.net/browse/DDFLSBP-210

#### Description
We no longer want to control the pause start date from CMS, but always fallback to the current date


src/apps/patron-page/PatronPage.tsx
JacobArrow, line 1: I'm not quite sure how this commit made its way in here

Adamik10 (approved):
Very nice 👏

PR #687, opened 2023-11-17
Only enable body scrolling after closing the last modal, not after every

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFLSBP-283

#### Description
This PR fixes an issue with scrolling behind modals. The user shouldn't be able to scroll in the content behind a modal until the very last modal is closed. Before, the scrolling was made possible after any modal was closed.

#### Screenshot of the result
n/a

n/a

### DIFF
--- a/src/core/modal.slice.ts
+++ b/src/core/modal.slice.ts
@@ -70,8 +70,6 @@ const modalSlice = createSlice({
       }
     },
     closeModal(state: StateProps, action: PayloadProps) {
-      // Enables background scrolling to use when Modal is closed
-      document.body.style.overflow = "";
       const modalId = state.modalIds.pop();
       if (state.modalIds.indexOf(action.payload.modalId) > -1) {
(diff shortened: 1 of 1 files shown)


src/core/modal.slice.ts
```diff
@@ -83,6 +81,12 @@ const modalSlice = createSlice({
         removeModalIdFromUrl(state);
         returnFocusElement();
       }
+      // Enables background scrolling to use when last modal is closed
+      // console.log(getConf("modalIds", configuration));
```
kasperbirch1, line 85: Do we want to have uncommented console.log's in the code?

kasperbirch1 (approved):
LGTM 👍

src/core/modal.slice.ts
Adamik10: Well spotted! I am removing it :)))

PR #688, opened 2023-11-17
Do not allow only() in tests

kasperg:
#### Description

While it may be best practice to use this when developing tests it is not a good idea to have it commit to VCS.

Doing so will prevent our CI (and future developers) from running other tests.

This is especially severe when running all tests locally as this means that only tests using only() will be used.

Using eslint-plugin-no-only-tests seems to be the suggested solution: https://github.com/cypress-io/eslint-plugin-cypress/issues/28

This will fail at the moment. It should be merged once we have removed all the onlys. This occurs as a part of #612.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -9,6 +9,9 @@
     "plugin:prettier/recommended",
     "plugin:cypress/recommended"
   ],
+  "plugins": [
+    "no-only-tests"
+  ],
   "settings": {
     "react": {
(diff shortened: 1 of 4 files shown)


kasperg:
@spaceo All of them should be solved as a part of #612. I see a new one popped up meanwhile in Advanced Search. This only underscores the usefulness here :smile:.

PR #701, opened 2023-11-21
Move webpack loaders for CSS files

kasperg:
#### Description

For whatever reason this comes up as a problem when using dapple. It does not seem to make a difference in other situations so we might as well add it.


spaceo (approved):
LGTM 👍

PR #703, opened 2023-11-22
Fixed bug where a digital reservation could not be opened

JacobArrow:
#### Link to issue
https://reload.atlassian.net/browse/DDFLSBP-311

#### Description
Digital materials does not have a faust or reservationId, so we need to check for the identifier also.


spaceo (approved):
LGTM 👍

PR #704, opened 2023-11-22
Feedback when deleting reservations (don't merge b4 dpl-cms)

spaceo:
#### Link to issue

https://reload.atlassian.net/browse/DDFLSBP-99

#### Description

Modal feedback when deleting reservation.

Work also implied:
* Creating a generic hook for doing a request and get status back that until now have been used in two modals with feedback
* Tame multiple requests with promises to ensure that all requests have been performed before we can give feedback
* Change some tests to handle the feedback in the flow of deleting reservations

#### Screenshot of the result

image

Merge [this](https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/491) first...


src/apps/reservation-list/modal/delete-reservation/delete-reservation-modal.tsx
```diff
@@ -74,16 +74,46 @@ const DeleteReservationModal: FC<DeleteReservationModalProps> = ({
   return (
     <Modal
       modalId={modalId}
-      classNames="modal-cta"
+      classNames="modal-cta modal-padding"
```
JacobArrow, line 77: `modal-padding` is a weird class...

src/apps/reservation-list/modal/delete-reservation/delete-reservation-modal.tsx
```diff
@@ -74,16 +74,46 @@ const DeleteReservationModal: FC<DeleteReservationModalProps> = ({
...
+          subTitle={t("deleteReservationModalErrorsStatusText")}
+          ctaButton={{
+            text: t("deleteReservationModalButtonText"),
+            closeAllModals: true,
+            callback: () => {
+              setRequestStatus("idle");
+            }
+          }}
+        />
+      )}
```
JacobArrow, line 116: Seems a bit redundant...
Would prefer an if statement and ternary operator inside title and subTitle, but that is just me I guess

src/apps/reservation-list/modal/delete-reservation/helper.ts
```diff
@@ -0,0 +1,61 @@
+import { UseTextFunction } from "../../../../core/utils/text";
+import { RequestStatus } from "../../../../core/utils/types/request";
+import { isFaust, isIdentifier } from "../../../dashboard/util/helpers";
+
+export const getReservationsToDelete = (reservations: string[]) => {
+  if (!reservations) {
+    return { physical: [], digital: [] };
```
JacobArrow, line 7: Do we not want to check for at reservation return type here?

src/apps/reservation-list/modal/delete-reservation/delete-reservation-modal.tsx
```diff
@@ -74,16 +74,46 @@ const DeleteReservationModal: FC<DeleteReservationModalProps> = ({
...
+        <DeleteReservationContent
+          deleteReservation={() => removeSelectedReservationsHandler()}
+          reservationsCount={reservations.length}
+          deletionStatus={requestStatus}
+        />
+      )}
+      {requestStatus === "success" && (
+        <ModalMessage
+          title={t("deleteReservationModalSuccessTitleText")}
+          subTitle={t("deleteReservationModalSuccessStatusText")}
```
JacobArrow, line 93: Can we tell the user how many reservations were deleted?

JacobArrow (approved):
Nice job!
I would really like that we could show the number of deleted reservations :)
Something "The reservation was deleted" or "4 reservations were deleted"

src/apps/reservation-list/modal/delete-reservation/delete-reservation-modal.tsx
spaceo: Yes ;) I skipped it to keep spent coding time down. But you are right, we need it.

src/apps/reservation-list/modal/delete-reservation/delete-reservation-modal.tsx
spaceo: I havde defined the ctaButton params once to remove redundant code. 👍

PR #717, opened 2023-11-27
Remove typescript-graphql-typed-files-modules

kasperg:
#### Description

The package does not seem to be used.

According to the documentation the package is used by adding typescript-graphql-typed-files-modules as a plugin in the YAML file for graphql-codegen.  https://github.com/agendrix/typescript-graphql-typed-files-modules

We do not have it declared in dbc-gateway.codegen.yml. Consequenty it does not seem to be used.

It is a bit tricky to remove packages which are only used for codegen and are not tested during standard CI. I think we should be good though.

The reason I add it now is that we have a Dependabot update for the package #713.

With that in mind I only assigned you @spaceo since you are the one who set up the code generation initially.

### DIFF
--- a/package.json
+++ b/package.json
@@ -132,7 +132,6 @@
     "svg-url-loader": "^8.0.0",
     "ts-node": "^10.7.0",
     "typescript": "^4.6.4",
-    "typescript-graphql-typed-files-modules": "^0.1.2",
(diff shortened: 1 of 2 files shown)


spaceo (approved):
LGTM 👍

PR #720, opened 2023-11-28
More precise types on requests and reservations

spaceo:
#### Link to issue

https://reload.atlassian.net/browse/DDFLSBP-99

#### Description

Set more precise types on requestsAndReservations
`requests`was not inferred correctly.

### DIFF
--- a/src/apps/reservation-list/modal/delete-reservation/helper.ts
+++ b/src/apps/reservation-list/modal/delete-reservation/helper.ts
@@ -1,3 +1,6 @@
+import { useDeleteReservations } from "../../../../core/fbs/fbs";
+import { DeleteReservationsParams } from "../../../../core/fbs/model/deleteReservationsParams";
+import { useDeleteV1UserReservationsIdentifier } from "../../../../core/publizon/publizon";
 import { UseTextFunction } from "../../../../core/utils/text";
 import { RequestStatus } from "../../../../core/utils/types/request";
 import {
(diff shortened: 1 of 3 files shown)


src/apps/reservation-list/modal/delete-reservation/helper.ts
```diff
@@ -1,10 +1,36 @@
...
+export type OperationPhysical = UseMutateFunction<
+  void | null,
+  unknown,
+  ParamsPhysical,
+  unknown
+>;
+type Request = {
+  params: { params: { reservationid: number[] } } | { identifier: string };
+  operation: OperationPhysical | OperationDigital;
+};
```
kasperg, line 33:
```suggestion
type Request = {
  params: { params: { reservationid: number[] } } 
  operation: OperationPhysical
} | {
  params: { identifier: string }
  operation: OperationDigital
};
```

If we want to make types more precise consider matching parameter with operation type.

src/apps/reservation-list/modal/delete-reservation/helper.ts
```diff
@@ -22,28 +48,6 @@ export const getReservationsToDelete = (reservations: ReservationType[]) => {
...
-  if (physical === "error" || digital === "error") {
-    return "error";
-  }
-
-  if (physical === "success" || digital === "success") {
-    return "success";
-  }
-
-  return "idle";
-};
```
kasperg, line 45: Nice to get rid of this.

kasperg (approved):
:+1: I am not sure what the problem is by looking at the code and reading the comment but it seems reasonable.

If we are talking precise types I have one suggestion for you.

src/apps/reservation-list/modal/delete-reservation/helper.ts
spaceo: Brilliant! Of course. Thx 👍

PR #721, opened 2023-11-28
Fix Markdown syntax for link.

achton:
#### Description

The software which builds documentation (MkDocs) is picky about validating URLs, and will break if they are not valid. This should fix that problem.

As far as I know, we cannot determine whether this change will fix the [broken documentation builds](https://github.com/danskernesdigitalebibliotek/dpl-docs/actions/runs/6970384342/job/18968175754) until we merge it to `develop` (or `main`?).


kasperg (approved):
:+1: Good catch

PR #722, opened 2023-11-28
Set correct fees url

JacobArrow:
#### Link to issue
https://reload.atlassian.net/browse/DDFLSBP-313

#### Description
In an attempt to streamline the fee url to a single field from Drupal, we wan to use the feesAndCompensationRatesUrl instead. materialOverdueUrl has been depricated and removed from Drupal.


Dresse (approved):
LGTM 👍

JacobArrow:
Fixed in #632

PR #723, opened 2023-11-28
Align fees url

JacobArrow:
#### Link to issue
https://reload.atlassian.net/browse/DDFLSBP-274

#### Description
In an attempt to streamline the fee url to a single field from Drupal, we wan to use the feesAndCompensationRatesUrl instead. feespageURL is used for the user fee page /user/me/fees


Dresse (approved):
LGTM 👍

PR #725, opened 2023-12-01
Tame those urls!!

spaceo:
AND importantly: fail if it is not present.
We had to many incidents where this happened, that is the story behind this initiative.

We should still consider having a validation step of the incoming props in the apps. The would highly improve the integrity of the apps.


Adamik10 (approved):
I am Adam Antal and I approve of this PR. Very good that we'll get notified about what's wrong when missing a URL now 👏

PR #753, opened 2023-12-06
Fix/pay button

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFLSBP-334

#### Description
This PR regards the "pay" button in a warning bar on the user dashboard page. This warning bar poppes up whenever a user has any fees to pay. The button text shouldn't be underlined, the translatable text in the button was updated, and an aria label has been added to it. 

#### Screenshot of the result
image

Siblings: https://github.com/danskernesdigitalebibliotek/dpl-design-system/pull/371
https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/531

### DIFF
--- a/src/apps/dashboard/dashboard.dev.tsx
+++ b/src/apps/dashboard/dashboard.dev.tsx
@@ -50,8 +50,12 @@ export default {
       defaultValue: "You owe in total",
       control: { type: "text" }
     },
-    payOwedText: {
-      defaultValue: "Pay",
+    seeMoreText: {
+      defaultValue: "See more",
+      control: { type: "text" }
+    },
+    seeMoreAriaLabelText: {
+      defaultValue: "See your fees and how to pay",
       control: { type: "text" }
     },
     totalAmountFeeText: {
--- a/src/apps/dashboard/dashboard.dev.tsx
+++ b/src/apps/dashboard/dashboard.dev.tsx
@@ -50,11 +50,11 @@ export default {
       defaultValue: "You owe in total",
       control: { type: "text" }
     },
-    seeMoreText: {
+    dashboardSeeMoreFeesText: {
       defaultValue: "See more",
       control: { type: "text" }
     },
-    seeMoreAriaLabelText: {
+    dashboardSeeMoreFeesAriaLabelText: {
       defaultValue: "See your fees and how to pay",
       control: { type: "text" }
     },
--- a/src/apps/loan-list/materials/utils/warning-bar.tsx
+++ b/src/apps/loan-list/materials/utils/warning-bar.tsx
@@ -1,4 +1,4 @@
-import React, { FC } from "react";
+import React, { FC, useId } from "react";
 import IconWarning from "@danskernesdigitalebibliotek/dpl-design-system/build/icons/basic/icon-warning.svg";
 import Link from "../../../../components/atoms/links/Link";
 
@@ -6,6 +6,7 @@ interface WarningBarProps {
   linkText?: string;
   overdueText?: string;
   rightButtonText?: string;
+  rightButtonAriaLabelText?: string;
   rightText?: string;
   leftLink?: URL;
   rightLink?: URL;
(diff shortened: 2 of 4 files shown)


src/apps/dashboard/dashboard.dev.tsx
```diff
@@ -50,8 +50,12 @@ export default {
       defaultValue: "You owe in total",
       control: { type: "text" }
     },
-    payOwedText: {
-      defaultValue: "Pay",
+    seeMoreText: {
+      defaultValue: "See more",
+      control: { type: "text" }
+    },
+    seeMoreAriaLabelText: {
```
JacobArrow, line 57: Align the naming to the one above, if you do decide to rename it

src/apps/loan-list/materials/utils/warning-bar.tsx
```diff
@@ -32,7 +36,7 @@ const WarningBar: FC<WarningBarProps> = ({
             {leftLink && (
               <Link
                 href={leftLink}
-                className="link-tag color-secondary-gray ml-8"
+                className="link-tag color-secondary-gray ml-8 hide-linkstyle"
```
JacobArrow, line 39: Can't we override the link styling in the design system?
We know we have multiple places where a link button is rendered with an underline, and adding this utility class everywhere is not the solution. There should not be a case where we want an underline.

src/apps/loan-list/materials/utils/warning-bar.tsx
```diff
@@ -48,11 +52,15 @@ const WarningBar: FC<WarningBarProps> = ({
...
           </p>
+          <span className="hide-visually" id={labelId}>
+            {rightButtonAriaLabelText}
+          </span>
           {rightLink && (
             <Link
               dataCy="warning-bar-right-link"
               href={rightLink}
-              className="btn-primary btn-filled btn-small arrow__hover--right-small"
+              className="btn-primary btn-filled btn-small arrow__hover--right-small hide-linkstyle"
```
JacobArrow, line 62: Can't we override the link styling in the design system?
We know we have multiple places where a link button is rendered with an underline, and adding this utility class everywhere is not the solution. There should not be a case where we want an underline.

src/apps/loan-list/materials/utils/warning-bar.tsx
Adamik10: @JacobArrow no, we just don't want the text in buttons to be underlined ever.

src/apps/loan-list/materials/utils/warning-bar.tsx
Adamik10: https://github.com/danskernesdigitalebibliotek/dpl-design-system/pull/371

src/apps/loan-list/materials/utils/warning-bar.tsx
JacobArrow: > we just don't want the text in buttons to be underlined ever
But this is a regular link - why remove the underline then?

src/apps/loan-list/materials/utils/warning-bar.tsx
```diff
@@ -48,11 +52,15 @@ const WarningBar: FC<WarningBarProps> = ({
           >
             {rightText}
           </p>
+          <span className="hide-visually" id={labelId}>
+            {rightButtonAriaLabelText}
+          </span>
           {rightLink && (
             <Link
```
JacobArrow, line 59: Wouldn't it be better to use the LinkButton component here?

JacobArrow (changes requested):
Nice to have the LinkButton styles controlled by the design system.
I had a few more thoughts - sorry :3

src/apps/loan-list/materials/utils/warning-bar.tsx
Adamik10: Great point!

src/apps/loan-list/materials/utils/warning-bar.tsx
Adamik10: @JacobArrow does it matter? It is only a label. And it's not visible at all anyways.

JacobArrow (approved):
Nice!
Great job 💪🏼

PR #755, opened 2023-12-07
Don't hide fee list page

JacobArrow:
#### Link to issue
https://reload.atlassian.net/browse/DDFLSBP-310

#### Description
At some point it was decided to hide the whole page when any modal was open.
This gives some weird behavior when opening non-full screen modals like the menu. We don't want to hide the page.


Dresse (approved):
LGTM 👍

PR #769, opened 2023-12-11
Add Dependabot group for Redux

kasperg:
#### Description

There are often dependencies between core Redux and packages building on Redux like Redux Toolkit, React Redux, Redux Persist e.g. #760 and #765.

Group them to make them move in parallel.


spaceo (approved):
LGTM 👍

PR #772, opened 2023-12-11
Update interest period treshold with drupal

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFLSBP-150

#### Description
This PR updates the usage of the "interest period" after which a loaner is no longer interested in a material when reserving it. There were [some updates in the backend](https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/475) to the structure of the data we use, so here we update the frontend part.
We also add new wiremock stubs for the material page (so that it doesn't crash when running it locally).

#### Screenshot of the result
n/a
n/a

### DIFF
--- a/src/components/reservation/UserListItems.tsx
+++ b/src/components/reservation/UserListItems.tsx
@@ -51,7 +51,7 @@ const UserListItems: FC<UserListItemsProps> = ({
 }) => {
   const t = useText();
   const config = useConfig();
-  const interstPeriods = config<InterestPeriods>("interestPeriodsConfig", {
+  const interestPeriods = config<InterestPeriods>("interestPeriodsConfig", {
     transformer: "jsonParse"
   });
 
@@ -61,10 +61,10 @@ const UserListItems: FC<UserListItemsProps> = ({
   };
 
   const interestPeriod = selectedInterest
-    ? getNoInterestAfter(selectedInterest, interstPeriods, t)
+    ? getNoInterestAfter(selectedInterest, interestPeriods, t)
     : getNoInterestAfter(
-        Number(interstPeriods.defaultInterestPeriod.value),
-        interstPeriods,
+        Number(interestPeriods.defaultInterestPeriod.value),
+        interestPeriods,
         t
       );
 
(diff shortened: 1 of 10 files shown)


src/components/reservation/UserListItems.tsx
```diff
@@ -51,7 +51,7 @@ const UserListItems: FC<UserListItemsProps> = ({
 }) => {
   const t = useText();
   const config = useConfig();
-  const interstPeriods = config<Option[]>("interestPeriodsConfig", {
+  const interstPeriods = config<InterestPeriods>("interestPeriodsConfig", {
```
JacobArrow, line 54: Fix spelling

src/components/reservation/types.ts
```diff
@@ -0,0 +1,6 @@
+import { Option } from "../Dropdown/Dropdown";
+
+export type InterestPeriods = {
+  interestPeriods: Option[];
+  defaultInterestPeriod: Option;
+};
```
JacobArrow, line 6: This feels weird for me.
I would suggest naming the type `Periods`

src/components/reservation/helper.ts
```diff
@@ -31,10 +31,10 @@ export const getPreferredBranch = (id: string, array: AgencyBranch[]) => {
 
 export const getNoInterestAfter = (
   days: number,
-  interestPeriod: Option[],
+  interestPeriod: InterestPeriods,
```
JacobArrow, line 34: Rename to period

src/components/reservation/forms/NoInterestAfterModal.tsx
```diff
@@ -22,7 +22,7 @@ const NoInterestAfterModal = ({
 }: PickupModalProps) => {
   const t = useText();
   const config = useConfig();
-  const interstPeriods = config<Option[]>("interestPeriodsConfig", {
+  const interstPeriods = config<InterestPeriods>("interestPeriodsConfig", {
```
JacobArrow, line 25: Rename to periods

JacobArrow (approved):
I'm gonna approve this, but I really think you should rethink the implementation of the `interestPeriods` object. It seems like don't really need the whole object anywhere and are using either `defaultInterestPeriod` or `interestPeriods`, and are therefore making unnecessary property accessors. 

I know its a minor thing, but `interestPeriod.interestPeriods.find()` is just not very pretty.

If you just want this out in a hurry, then doing something like this, will be the quick (but not perfect) solution: `periods.interestPeriods.find()`

PR #787, opened 2023-12-16
Show fees after payment change date on the fees page

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFLSBP-349

#### Description
Fees after 27th of October were not showing (payment change happened on that date) on the fees page.

#### Screenshot of the result
image

n/a

### DIFF
--- a/src/apps/fee-list/FeeList.tsx
+++ b/src/apps/fee-list/FeeList.tsx
@@ -63,7 +63,7 @@ const FeeList: FC = () => {
       }
       const feesPostPaymentChange = getFeesInRelationToPaymentChangeDate(
         fbsFees,
-        true
+        false
       ).length;
       if (feesPostPaymentChange > 0) {
         setItemsPostPaymentChange(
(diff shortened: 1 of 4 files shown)


src/apps/fee-list/FeeList.tsx
```diff
@@ -63,7 +63,7 @@ const FeeList: FC = () => {
       }
       const feesPostPaymentChange = getFeesInRelationToPaymentChangeDate(
         fbsFees,
-        true
+        false
```
spaceo, line 66: Cray detail 😅 
Although this is sufficient I think we should also channge the "hanging" hardcoded [change-date](https://github.com/danskernesdigitalebibliotek/dpl-react/blob/103b4b10892a90180d48138093d70bf04dc5a3f2/src/apps/fee-list/utils/helper.ts#L8).
We should put it in the config area in core.
Because of the urgency of getting this out as soon as possible, I think it is ok to create a task for it and move along.

spaceo (approved):
Approved. But please look at my comment.

src/apps/fee-list/FeeList.tsx
Adamik10: But is it really something they should be able to configure? I mean the date is never gonna change. It's something that happened in the past. Why do you think it should be configurable?

PR #796, opened 2023-12-19
Expanding fetching system for FBI to handle multiple urls

spaceo:
#### Link to issue
https://reload.atlassian.net/browse/DDFLSBP-318

#### Description
In that way we can change FBI graphql url depending on which query that is being executed.
We need that because DDF wants to change profile name depending on which query that is being run.

#### Additional comments
Is connected to [this PR](https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/581).

### DIFF
--- a/src/core/dbc-gateway/helper.ts
+++ b/src/core/dbc-gateway/helper.ts
@@ -1,40 +1,17 @@
 import { QueryFunctionContext } from "react-query";
 import { beforeAll, vi } from "vitest";
-import {
-  getServiceBaseUrl,
-  serviceUrlKeys
-} from "../utils/reduxMiddleware/extractServiceBaseUrls";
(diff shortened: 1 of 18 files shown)


src/core/dbc-gateway/helper.ts
```diff
@@ -0,0 +1,102 @@
...
+  }
+
+  const { queryKey } = context;
+  const [queryName] = queryKey;
+  return resolveBaseUrl(queryName as string);
+};
+
+export default {};
+
+/* ********************************* Vitest Section  ********************************* */
```
Adamik10, line 33: Very nice with the test :)))

Adamik10 (approved):
I have a comment/question, but very good job! :))
There's something weird going on with the one test that hasn't run, but you got that I'm sure 😉

PR #797, opened 2023-12-20
Refactor button reservable from another library

kasperbirch1:
#### Link to issue

#### Description

This pull request refactors the order buttons by moving the logic for displaying the `MaterialButtonReservableFromAnotherLibrary` component and refactoring the `getReservablePidsFromAnotherLibrary` function into a `useReservableFromAnotherLibrary` hook. That also checks if the material is not reservable and should be ordered via openOrder.

#### Screenshot of the result

### DIFF
--- a/src/core/utils/useReservableFromAnotherLibrary.tsx
+++ b/src/core/utils/useReservableFromAnotherLibrary.tsx
@@ -1,0 +1,28 @@
+import { useConfig } from "./config";
+import { getAllFaustIds } from "./helpers/general";
+import { useGetHoldings } from "../../apps/material/helper";
+import { Manifestation } from "./types/entities";
+
+const useReservableFromAnotherLibrary = (manifestations: Manifestation[]) => {
+  const config = useConfig();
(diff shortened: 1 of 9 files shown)


src/core/utils/useReservableFromAnotherLibrary.tsx
```diff
@@ -0,0 +1,28 @@
+import { useConfig } from "./config";
+import { getAllFaustIds } from "./helpers/general";
+import { useGetHoldings } from "../../apps/material/helper";
+import { Manifestation } from "./types/entities";
+
+const useReservableFromAnotherLibrary = (manifestations: Manifestation[]) => {
```
spaceo, line 6: Great with the hook.
I think tou should add some tests testing the hook now that you have deleted the old tests.

spaceo (changes requested):
Looks good. But I think it is a bad idea to delete tests without adding new

spaceo (approved):
LGTM 👍

kasperg:
@spaceo already gave this a :+1: :smile:

PR #799, opened 2023-12-20
Use correct semantics for header

JacobArrow:
#### Link to issue
https://reload.atlassian.net/browse/DDFLSBP-177

#### Description
A div is not semantically correct to use inside a h-tag. Instead we should use a span and tell the screen reader to percieve it as a text.

### DIFF
--- a/src/apps/loan-list/list/ListHeader.tsx
+++ b/src/apps/loan-list/list/ListHeader.tsx
@@ -11,9 +11,7 @@ const ListHeader: FC<ListHeaderProps> = ({ header, amount, children }) => {
     <div className="dpl-list-buttons">
       <h2 className="dpl-list-buttons__header">
         {header}
-        <span role="text" className="dpl-list-buttons__power">
-          {amount}
-        </span>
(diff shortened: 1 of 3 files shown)


src/apps/loan-list/list/ListHeader.tsx
```diff
@@ -11,7 +11,9 @@ const ListHeader: FC<ListHeaderProps> = ({ header, amount, children }) => {
     <div className="dpl-list-buttons">
       <h2 className="dpl-list-buttons__header">
         {header}
-        <div className="dpl-list-buttons__power">{amount}</div>
+        <span role="text" className="dpl-list-buttons__power">
```
spaceo, line 14: Apparently text is not a valid role?
From what I can read it should be a WAI-ARIA role as described in 5.4 definition of roles:
[5.4 Definition of Roles](https://www.w3.org/TR/wai-aria-1.1/#role_definitions)

spaceo (changes requested):
Our static accessibility check is complaining...

spaceo (approved):
LGTM 👍

PR #809, opened 2023-12-27
Fix/interest period

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFLSBP-356

#### Description
This PR fixes various bugs experienced in the reservation flow when changing interest period. 
- default value was inconsistent 
- switching between interest periods wasn't mirrored properly

#### Screenshot of the result
n/a

n/a

### DIFF
--- a/src/components/reservation/UserListItems.tsx
+++ b/src/components/reservation/UserListItems.tsx
@@ -36,12 +36,7 @@ export interface UserListItemsProps {
 
 const UserListItems: FC<UserListItemsProps> = ({
   patron,
-  patron: {
-    defaultInterestPeriod,
-    preferredPickupBranch,
-    phoneNumber,
-    emailAddress
-  },
+  patron: { preferredPickupBranch, phoneNumber, emailAddress },
   branches,
   selectedBranch,
   selectBranchHandler,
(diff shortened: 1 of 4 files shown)


src/components/reservation/UserListItems.tsx
```diff
@@ -84,7 +79,10 @@ const UserListItems: FC<UserListItemsProps> = ({
             buttonAriaLabel={t("changeInterestPeriodText")}
           />
           <NoInterestAfterModal
-            selectedInterest={selectedInterest ?? defaultInterestPeriod}
+            selectedInterest={
+              selectedInterest ??
+              Number(interestPeriods.defaultInterestPeriod.value)
```
spaceo, line 84: Isn't there a way to make sure that `interestPeriods.defaultInterestPeriod.value` is a number in an earlier stage. I think we should avoid casting values if possible.

spaceo (commented):
II have one question

src/components/reservation/UserListItems.tsx
Adamik10: Unfortunately the `Option` type is defined with a `value: string`, and we use this type in multiple dropdowns in the reservation workflow, so changing it there wouldn't be smart, I'd say. But I could alter the <NoInterestAfterModal> component! That would be a solution I think :)

PR #812, opened 2023-12-28
Update advanced search filters

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DSC-81

#### Description
The FBI API came out with some changes to their specs, so we now need to update how we contact it. As a part of this, most keys of advancedSearchIndexTranslations also needed to be updated to match.
As a part of this PR, a minor bug was also fixed where the number 0 was showing on the page as a part of the loading results text headline

#### Screenshot of the result
n/a

n/a

### DIFF
--- a/src/apps/advanced-search/AdvancedSearchResults.tsx
+++ b/src/apps/advanced-search/AdvancedSearchResults.tsx
@@ -88,7 +88,7 @@ const AdvancedSearchResult: React.FC<AdvancedSearchResultProps> = ({
   }, [data, page]);
 
   const shouldShowSearchResults = isLoading || (!isLoading && hitcount > 0);
-  const shouldShowResultHeadline = hitcount && !isLoading;
+  const shouldShowResultHeadline = !!(hitcount && !isLoading);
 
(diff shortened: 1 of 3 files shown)


src/apps/advanced-search/AdvancedSearchResults.tsx
```diff
@@ -88,7 +88,7 @@ const AdvancedSearchResult: React.FC<AdvancedSearchResultProps> = ({
   }, [data, page]);
 
   const shouldShowSearchResults = isLoading || (!isLoading && hitcount > 0);
-  const shouldShowResultHeadline = hitcount && !isLoading;
+  const shouldShowResultHeadline = !!(hitcount && !isLoading);
```
spaceo, line 91: I prefer casting to Boolean. I think it is easier to read. But that is a matter of taste.
```typescript
  const shouldShowResultHeadline = Boolean(hitcount && !isLoading);

spaceo (approved):
LGTM 👍 
I have one suggestion...

PR #813, opened 2023-12-29
Fix single vs group loan click events

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFLSBP-338

#### Description
Some of the business logic needed to be updated for loan list cards. When clicking on the title of a grouped loan, it is the group modal that should open, not a modal with details. 

#### Screenshot of the result
n/a

n/a


spaceo (approved):
LGTM 👍

PR #814, opened 2023-12-29
Fix/arrows

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFLSBP-319

#### Description
This PR makes the usage of arrows on cards consistent throughout.

#### Screenshot of the result
n/a

[Sibling PR in the design system](https://github.com/danskernesdigitalebibliotek/dpl-design-system/pull/412)

### DIFF
--- a/src/apps/dashboard/dashboard-notification/dashboard-notification.tsx
+++ b/src/apps/dashboard/dashboard-notification/dashboard-notification.tsx
@@ -31,7 +31,7 @@ const DashboardNotification: FC<DashboardNotificationProps> = ({
       onClick={notificationClickEvent}
       className="mb-16"
     >
-      <div className="list-dashboard shadow-medium-hover arrow__hover--right-small">
+      <div className="list-dashboard shadow-medium-hover arrow arrow__hover--right-small">
         <div
           className={`color-secondary-gray number number--${notificationColor}`}
         >
--- a/src/apps/dashboard/dashboard-notification/dashboard-notification.tsx
+++ b/src/apps/dashboard/dashboard-notification/dashboard-notification.tsx
@@ -31,7 +31,7 @@ const DashboardNotification: FC<DashboardNotificationProps> = ({
       onClick={notificationClickEvent}
       className="mb-16"
     >
-      <div className="list-dashboard shadow-medium-hover arrow arrow__hover--right-small">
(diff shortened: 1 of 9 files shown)


JacobArrow:
I'm waiting for the implementation in design system before approving this :)

src/apps/dashboard/dashboard-notification/dashboard-notification.tsx
```diff
@@ -31,7 +31,7 @@ const DashboardNotification: FC<DashboardNotificationProps> = ({
       onClick={notificationClickEvent}
       className="mb-16"
     >
-      <div className="list-dashboard shadow-medium-hover arrow__hover--right-small">
+      <div className="list-dashboard shadow-medium-hover arrow arrow__hover--right-small">
```
JacobArrow, line 34: Where is the `arrow` class in design system? Cant find it😬

src/components/Buttons/ArrowButton.tsx
```diff
@@ -18,6 +18,7 @@ const ArrowButton: React.FC<ArrowButtonProps> = ({
   return (
     <button
       aria-labelledby={arrowLabelledBy}
+      className="arrow-button"
```
JacobArrow, line 21: Why does this not have `arrow__hover--right-small` like the rest?

JacobArrow (dismissed):
I had some thoughts

JacobArrow (changes requested):
I had some thoughts

src/apps/dashboard/dashboard-notification/dashboard-notification.tsx
Adamik10: I'm not sure whether it's there, but I believer we use the arrow class together whenever we use arrow__hover--right-small

src/components/Buttons/ArrowButton.tsx
Adamik10: What do you mean like the rest? the arrow__hover--right-small is from my understanding only used on the card element containing the arrow inside it.

JacobArrow: Just sharing my thoughts, and wanted to make sure it was not an oversight. Carry on good sir 😁

JacobArrow (approved):
Let's gooooo 🚀

PR #828, opened 2024-01-02
Fix/translation error throwing

Adamik10:
#### Link to issue
-

#### Description
Errors were being thrown left and right for translations defined as empty strings. That shouldn't be the case.

#### Screenshot of the result
n/a

n/a


spaceo (approved):
LGTM 👍

PR #854, opened 2024-01-15
Change the 'Reserve' button text to the standard and move the notice about material ordered from other libraries to the `MaterialAvailabilityTextParagraph`.

kasperbirch1:
#### Link to issue

#### Description
This PR enhances the user experience by providing a more streamlined interface, while still effectively notifying the user when the material is ordered from an another library.

In this revision, I have incorporated a useEffect hook within the `UserListItems` component. This modification ensures that the default branch and interest period, as specified in the `ModalReservationFormSelect`, are consistently used as the values transmitted with the request.

#### Screenshot of the result

https://github.com/danskernesdigitalebibliotek/dpl-react/assets/49920322/d1823d27-6faa-4cad-9370-eaa84d7260c8

### DIFF
--- a/src/components/material/material-buttons/physical/MaterialButtonReservableFromAnotherLibrary.tsx
+++ b/src/components/material/material-buttons/physical/MaterialButtonReservableFromAnotherLibrary.tsx
@@ -41,7 +41,9 @@ const MaterialButtonReservableFromAnotherLibrary: FC<
       label={
         size === "small"
           ? t("reserveText")
-          : `${t("reserveText")} ${manifestationMaterialType}`
+          : `${t("reserveWithMaterialTypeText", {
+              placeholders: { "@materialType": manifestationMaterialType }
+            })}`
       }
       buttonType="none"
       variant="filled"
--- a/src/components/material/material-buttons/physical/MaterialButtonReservableFromAnotherLibrary.tsx
+++ b/src/components/material/material-buttons/physical/MaterialButtonReservableFromAnotherLibrary.tsx
@@ -40,10 +40,8 @@ const MaterialButtonReservableFromAnotherLibrary: FC<
       dataCy={dataCy}
       label={
         size === "small"
-          ? t("reservableFromAnotherLibraryText")
-          : `${t(
-              "reservableFromAnotherLibraryText"
-            )} ${manifestationMaterialType}`
+          ? t("reserveText")
+          : `${t("reserveText")} ${manifestationMaterialType}`
       }
       buttonType="none"
       variant="filled"
--- a/src/components/reservation/ReservationModalBody.tsx
+++ b/src/components/reservation/ReservationModalBody.tsx
@@ -129,9 +129,10 @@ export const ReservationModalBody = ({
     !!selectedPeriodical
   );
 
-  const reservablePidsFromAnotherLibrary = useReservableFromAnotherLibrary(
-    selectedManifestations
-  );
+  const {
+    reservablePidsFromAnotherLibrary,
+    materialIsReservableFromAnotherLibrary
+  } = useReservableFromAnotherLibrary(selectedManifestations);
 
   // If we don't have all data for displaying the view render nothing.
   if (!userResponse.data || !holdingsResponse.data) {
@@ -178,14 +179,14 @@ export const ReservationModalBody = ({
       );
     }
 
-    if (reservablePidsFromAnotherLibrary?.length > 0 && patron) {
+    if (materialIsReservableFromAnotherLibrary && patron) {
       const { patronId, name, emailAddress, preferredPickupBranch } = patron;
 
       // Save reservation to open order.
       mutateOpenOrder(
         {
           input: {
-            pids: [...reservablePidsFromAnotherLibrary],
+            pids: reservablePidsFromAnotherLibrary,
             pickUpBranch: selectedBranch
               ? removePrefixFromBranchId(selectedBranch)
               : removePrefixFromBranchId(preferredPickupBranch),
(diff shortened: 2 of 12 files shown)


src/components/reservation/ReservationModalBody.tsx
```diff
@@ -177,19 +172,20 @@ export const ReservationModalBody = ({
       );
     }
 
-    if (reservablePidsFromAnotherLibrary?.length && patron) {
-      const { patronId, name, emailAddress, preferredPickupBranch } = patron;
+    if (
+      reservablePidsFromAnotherLibrary?.length &&
+      patron &&
```
spaceo, line 177: We have a function that is more descriptive/made for the purpose of checking if a user is logged in that is called somthing like: `userIsLoggedIn` I think we should use that instead and I think that should be the first check in this logic.

src/components/reservation/ReservationModalBody.tsx
```diff
@@ -177,19 +172,20 @@ export const ReservationModalBody = ({
       );
     }
 
-    if (reservablePidsFromAnotherLibrary?.length && patron) {
-      const { patronId, name, emailAddress, preferredPickupBranch } = patron;
+    if (
+      reservablePidsFromAnotherLibrary?.length &&
```
spaceo, line 176: I think we should make a function or a variable called something like: `materialIsReservableFromAnotherLibrary`to reflect the domain instead of referring to some inner values. Does it make sense?

src/components/reservation/UserListItems.tsx
```diff
@@ -55,49 +55,52 @@ const UserListItems: FC<UserListItemsProps> = ({
...
-  const pickupBranch = selectedBranch
-    ? getPreferredBranch(selectedBranch, branches)
-    : getPreferredBranch(preferredPickupBranch, branches);
+  // Set default values for interest period and pickup branch
+  useEffect(() => {
+    if (!selectedInterest) {
+      setSelectedInterest(Number(interestPeriods.defaultInterestPeriod.value));
+    }
+    if (!selectedBranch) {
+      selectBranchHandler(preferredPickupBranch);
```
spaceo, line 64: Is it not possible to set these values as initial values in the useState hooks?

kasperbirch1:
Look at these todo's
https://github.com/danskernesdigitalebibliotek/dpl-react/blob/develop/src/apps/reservation-list/modal/reservation-details/reservation-details.test.ts#L503-L506
https://github.com/danskernesdigitalebibliotek/dpl-react/blob/develop/src/apps/reservation-list/modal/reservation-details/reservation-details.test.ts#L546-L547

spaceo (changes requested):
I don't understand what the fix you did what it consists of it what it solves.
I wrote some comments that are related to code style/practise.
Also I think that the `ReservationModalBody` has too much logic inside of the comopnent. It would be nice to på the `saveReservation` handler and state in an external hook

src/components/reservation/ReservationModalBody.tsx
kasperbirch1: I can only find `isAnonymous` ? But I still need the `patron` check? So, is it necessary?

src/components/reservation/ReservationModalBody.tsx
kasperbirch1: It comes from the `useReservableFromAnotherLibrary` hook. Should I rename it there?

kasperg:
This has been lingering for quite a while now. I am assigning this to you @kasperbirch1 so you can figure out what is needed to move this along.

kasperbirch1:
@spaceo now it finally works 👍 


https://reload.atlassian.net/browse/DDFNEXT-82

src/components/reservation/ReservationModalBody.tsx
spaceo: I see what you mean.
But actually I think we should use `isAnonymous` as a part of the `usePatronData()` request by setting enabled: 
```js
{
  enabled: !isAnonymous()
}
``` 

In that way we do not send the patron request if it is not needed.

src/components/reservation/ReservationModalBody.tsx
spaceo: I think it makes sense to create the variable right after this:
```typescript
  const reservablePidsFromAnotherLibrary = useReservableFromAnotherLibrary(
    selectedManifestations
  );
const materialIsReservableFromAnotherLibrary = Boolean(reservablePidsFromAnotherLibrary.length);
```

src/components/reservation/UserListItems.tsx
spaceo: Please consider this and reply :)

spaceo (commented):
Please look at my comments

src/components/reservation/ReservationModalBody.tsx
kasperbirch1: Fixet

src/components/material/material-buttons/physical/MaterialButtonReservableFromAnotherLibrary.tsx
```diff
@@ -40,10 +40,8 @@ const MaterialButtonReservableFromAnotherLibrary: FC<
       dataCy={dataCy}
       label={
         size === "small"
-          ? t("reservableFromAnotherLibraryText")
-          : `${t(
-              "reservableFromAnotherLibraryText"
-            )} ${manifestationMaterialType}`
+          ? t("reserveText")
+          : `${t("reserveText")} ${manifestationMaterialType}`
```
spaceo, line 44: This is just a note. But it is not the right use of `t`.
`manifestationMaterialType` should have been inserted via a placeholder.

spaceo (approved):
LGTM 👍

PR #861, opened 2024-01-17
Change materialTypes field on work as well

spaceo:
#### Link to issue

https://reload.atlassian.net/browse/DDFLSBP-380

#### Description

Apparently there is both a materialTypes field on work as well as manifestations. This changes the one at Work level.


Adamik10 (approved):
LGTM

PR #864, opened 2024-01-18
DDFLSBP-377 - Added a check to see if blockedPatronELinkUrl is the sa…

Dresse:
#### Link to issue

https://reload.atlassian.net/browse/DDFLSBP-377

#### Description

This PR checks if the blockedPatronELinkUrl is the same as the current location URL. If this is the case, we do not want to show the blockedPatronELinkUrl and text in the blockedUser Modal.

This PR has a sister PR in dpl-cms. It can be found here: https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/653

### DIFF
--- a/src/core/utils/url.tsx
+++ b/src/core/utils/url.tsx
@@ -8,7 +8,11 @@ export const useUrls = () => {
   const { data } = useSelector((state: RootState) => state.url);
   const urls = useMemo(() => turnUrlStringsIntoObjects(data), [data]);
 
-  return (name: string) => {
+  return (name: string, returnFalseIfUndefined = false) => {
+    if (returnFalseIfUndefined) {
+      return urls[name] || false;
+    }
+
     if (!urls[name]) {
       throw new Error(`The url ${name} is not defined`);
     }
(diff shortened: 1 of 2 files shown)


spaceo (changes requested):
We have dound another solution. This should probably go to the eternal afterlife of non merged PR's

src/core/utils/url.tsx
```diff
@@ -8,7 +8,11 @@ export const useUrls = () => {
...
 
-  return (name: string) => {
+  return (name: string, returnFalseIfUndefined = false) => {
+    if (returnFalseIfUndefined) {
+      return urls[name] || false;
+    }
+
     if (!urls[name]) {
       throw new Error(`The url ${name} is not defined`);
     }
```
spaceo, line 18: This is just a comment and something we should be aware of:
We do [the check](https://github.com/danskernesdigitalebibliotek/dpl-react/blob/e4eaa0d779fe702f97319fe2253b10f5dec658da/src/core/utils/text.tsx#L116) differently in `useText`.
So it would be nice to do it in the same manner. In that way we would expect both systems to behave the same way.

spaceo (approved):
LGTM, but please see comment

PR #866, opened 2024-01-18
Fix/double fees in group modal

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFLSBP-365

#### Description
Fee detail materials were shown multiple time on the list inside the fee list details modal. Now they only show a singular time per piece.

#### Screenshot of the result
-
-


Dresse (approved):
LGTM

PR #869, opened 2024-01-19
Don't open material details modal after interacting with the pager

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFLSBP-367

#### Description
Pager gives focus to the next stackable material too quickly while still
registering the key press. So when interacting with the pager on the
keyboard, material details of the following, newly shown item would open
unintentionally.

#### Screenshot of the result
-
-


spaceo (approved):
Great that you found a solution. I hope it is a sustainable solution as well. Great with the comments! LGTM 👍

PR #870, opened 2024-01-19
Change wrong wiremock request.url property

spaceo:
#### Description

Change wrong wiremock request.url property

The pattern matching should be more precise in order to match the exact
query parameters or no query parameters.

Wiremock is failing delivering fees which breaks stories that depends on those.


Adamik10 (approved):
✨️

PR #888, opened 2024-01-23
Decimals with comma

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFLSBP-386

#### Description

This PR makes sure that when rendering currency we use comma instead of a dot for decimals as per European standards.

#### Screenshot of the result
image

-


Dresse (approved):
LGTM!

PR #891, opened 2024-01-24
Fix Group Modal for Reservations [DDFLSBP-374]

clausbruun:
#### Link to issue

https://reload.atlassian.net/browse/DDFLSBP-374

#### Description

* Fixed double arrows by providing parameter for the "default case". 
* Changed `IndexOf()` to `findIndex()` for object array matching. `IndexOf()` compares objects based on their reference in memory, not their content. If you have a separate object with the same properties as an object in the array, `IndexOf()` won't find it because it checks for reference equality.
 
#### Screenshot of the result

https://github.com/danskernesdigitalebibliotek/dpl-react/assets/105956/f482f1c1-d6ee-491f-9803-3401246cf6ba

Updated screenshot with:
* Remove queue number label for physical materials
Skærmbillede 2024-01-26 kl  10 56 10

*

### DIFF
--- a/src/apps/reservation-list/reservation-material/reservation-info.tsx
+++ b/src/apps/reservation-list/reservation-material/reservation-info.tsx
@@ -104,12 +104,19 @@ const ReservationInfo: FC<ReservationInfoProps> = ({
   }
 
   if (state === "reserved" && pickupBranch && numberInQueue && expiryDate) {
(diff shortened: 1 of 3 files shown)


Adamik10 (approved):
Very nice! Good job! 🎉 🎈 🎆

PR #893, opened 2024-01-25
Fix/change interest period and pickup on reserved items

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFLSBP-342

#### Description
It is now possible to change interest period & pickup branch on users current reservations. Tests were also updated based on these changes. 

#### Screenshot of the result
-

-


spaceo (changes requested):
Great work!
There is a block that I think could be simplified a bit. I know that these changes are on top of the shoulders from our predecessors (the heritage of the giants).
But anyways I'll attach a patch (note: it is untested!)...
[physical-list-details.patch](https://github.com/danskernesdigitalebibliotek/dpl-react/files/14083069/physical-list-details.patch)

spaceo (approved):
LGTM 👍

PR #895, opened 2024-01-26
Fix/extend-reservations

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFLSBP-407
https://reload.atlassian.net/browse/DDFLSBP-382

#### Description
We don't use the accept extra fees modal for renewing reservationsOn the dashboard + loan list anymore. The business logic is moving away from this functionality for now. It was also causing errors submitting the renewal request.
I kept the actual accept modal component as it is highly likely that this functionality will need to be implemented in the future.

#### Screenshot of the result
-
-


spaceo (approved):
LGTM 👍

PR #896, opened 2024-01-26
Use date range input field instead of date input in pause reservation modal

spaceo:
#### Link to issue

https://reload.atlassian.net/browse/DDFLSBP-352

#### Description

Use date range input field instead of date input in pause reservation modal.
The Flatpickr package is used and it has already been styled in the design system

#### Screenshot of the result

image


.github/workflows/create-release-on-branch-changes.yml
```diff
@@ -46,6 +46,8 @@ jobs:
           registry-url: 'https://npm.pkg.github.com'
           always-auth: true
           scope: '@danskernesdigitalebibliotek'
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
```
Adamik10, line 50: Is this why that job was always failing the first time on GH? Very good!! :))

src/apps/patron-page/PatronPage.dev.tsx
```diff
@@ -70,12 +70,8 @@ export default {
       defaultValue: "Close pause reservations modal",
       control: { type: "text" }
     },
-    dateInputsStartDateLabelText: {
-      defaultValue: "Start date",
-      control: { type: "text" }
-    },
-    dateInputsEndDateLabelText: {
-      defaultValue: "End date",
+    pauseReservationModalDateRangeLabelText: {
```
Adamik10, line 73: Is this a newly introduced text? If so, then we need a PR in dpl-cms as well with it 😉

src/apps/reservation-list/list/reservation-list.dev.tsx
```diff
@@ -224,14 +224,6 @@ export default {
       defaultValue: "Close pause reservations modal",
       control: { type: "text" }
     },
-    dateInputsStartDateLabelText: {
```
Adamik10, line 227: If these texts are no longer used at all, I would suggest removing them from the CMS as well :)

src/apps/reservation-list/modal/delete-reservation/pause-reservations.test.ts
```diff
@@ -80,13 +80,9 @@ describe("Pause reservation modal test", () => {
...
         "Pause your reservations early, since reservations that are already being processed, will not be paused."
       );
-    // ID 12 2.c. datepicker: start date
-    cy.get(".modal.modal-cta [data-cy='start-date']")
-      .should("exist")
-      .should("have.attr", "value");
 
-    // ID 12 2.e. datepicker: end date
-    cy.get(".modal.modal-cta [data-cy='end-date']").should("exist");
+    // There should be a date range component for choosing a pause period.
```
Adamik10, line 84: 👏

src/core/storybook/reservationListArgs.ts
```diff
@@ -189,16 +189,12 @@ export default {
...
-  },
-  dateInputsStartDateLabelText: {
-    defaultValue: "Start date",
+  pauseReservationModalDateRangeLabelText: {
+    defaultValue: "Pause period",
     control: { type: "text" }
   },
-  dateInputsEndDateLabelText: {
-    defaultValue: "End date",
+  pauseReservationModalCloseModalText: {
```
Adamik10, line 196: Same as above - a CMS counterpart for this PR will be needed.

src/components/date-inputs/DateRangeInput.tsx
```diff
@@ -0,0 +1,81 @@
...
+    <div data-cy={dataCy} className={className}>
+      <div className="date-range__input">
+        <label
+          ref={refLabel}
+          htmlFor="date-range"
+          className="text-body-medium-regular"
+        >
+          {label}
+        </label>
+        <Flatpickr
```
Adamik10, line 55: Is there a way to catch an onKeyUp here for the ESC button and Enter? 
When I'm trying to close the popup with dates selected, it closes the whole modal instead. Same goes for Enter.

Adamik10, line 55: Also, would there be a way to introduce a placeholder text (something like "choose date period", and maybe a little arrow pointing downwards to indicate that this field is interactive? I think it would go a long way.
image

Adamik10 (commented):
Wonderful job!! 
I have a couple of comments, and thoughts, but you did great.

src/apps/patron-page/PatronPage.dev.tsx
spaceo: It's defined here:
https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/696

src/components/date-inputs/DateRangeInput.tsx
spaceo: This is tricky. I completely get your point, but I am afraid of adding substantially more time to this task, that already took some time.
I would have to implement some functionality thet prevents the genral modal system of catching esc/enter in case of that the flatpickr is open. That is not trivial.

spaceo: Great feedback!
I have added a placeholder (also in the cms).
Again I think the time invested in getting it to work with an arrow in the design system is not worth it compared to the time already spent.
If DDF in the future wants an arrow I have made sure a class is added to the field in case it is empty.

Adamik10 (approved):
APPROVED! :)))

PR #915, opened 2024-01-31
Retry yarn install if it fails in release workflows

spaceo:
We have continuously experienced the install failing because of problems downloading design system. Hopefully it helps retrying...


Adamik10 (approved):
Perfect! 🎉

PR #917, opened 2024-01-31
Refactor `getInstantLoanBranchHoldings`

kasperbirch1:
#### Description
This PR modified the `getInstantLoanBranchHoldings` to use precise matching with `===` instead of using the `includes` method for material group description.

### DIFF
--- a/src/tests/unit/__vitest_data__/instant-loan.ts
+++ b/src/tests/unit/__vitest_data__/instant-loan.ts
@@ -22,7 +22,7 @@ const branchHoldings: HoldingsV3[] = [
         periodical: undefined,
         materialGroup: {
           name: "standard",
-          description: "I am supposed to be matched, but I am not available"
+          description: "I NOT supposed to be matched"
         }
       },
       {
@@ -57,7 +57,7 @@ const branchHoldings: HoldingsV3[] = [
         periodical: undefined,
         materialGroup: {
           name: "standard",
-          description: "I am supposed to be matched, but I am not available"
+          description: "I am also supposed to be matched"
         }
       },
       {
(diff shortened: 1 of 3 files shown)


kasperbirch1:
It is worth mentioning that Tue wants a change to this functionality in the future, so we check by `name` instead of `description`

src/tests/unit/__vitest_data__/instant-loan.ts
```diff
@@ -22,7 +22,7 @@ const branchHoldings: HoldingsV3[] = [
         periodical: undefined,
         materialGroup: {
           name: "standard",
-          description: "I am supposed to be matched, but I am not available"
+          description: "I NOT supposed to be matched"
```
spaceo, line 25: Are you sure that the text should be changed here?
A lot of work went into writing the tests.

Note: "I NOT supposed to be matched" is not grammatically correct.

src/tests/unit/__vitest_data__/instant-loan.ts
```diff
@@ -83,7 +83,7 @@ const branchHoldings: HoldingsV3[] = [
         periodical: undefined,
         materialGroup: {
           name: "standard",
-          description: "I am supposed to be matched, seriously!"
```
spaceo, line 86: Are you sure that the text should be changed here?
I do not remember clearly. But the word `seriously` was probably added to make it distinct/stick out?

spaceo (changes requested):
Ihave some questions

src/tests/unit/__vitest_data__/instant-loan.ts
kasperbirch1: The new change dos that it is only the exact string that causes it to match.

src/tests/unit/__vitest_data__/instant-loan.ts
kasperbirch1: I'm adding a new commit to clarify why it should fail

spaceo (approved):
LGTM 👍

PR #918, opened 2024-01-31
Fix loan renewal error

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFLSBP-407

#### Description
When renewing loans - use loanIds to send to the service. We were sending nothing before.

#### Screenshot of the result
-
-


Dresse (approved):
LGTM

PR #920, opened 2024-01-31
Remove status labels in the user menu [DDFLSBP-252]

clausbruun:
#### Link to issue

https://reload.atlassian.net/browse/DDFLSBP-252

#### Description

This PR removes status labels in the user menu due to limited space.

#### Screenshot of the result

image

*

### DIFF
--- a/src/apps/dashboard/dashboard-notification-list/Notifications.tsx
+++ b/src/apps/dashboard/dashboard-notification-list/Notifications.tsx
@@ -19,8 +19,8 @@ export interface NotificationsProps {
 
 const Notifications: FC<NotificationsProps> = ({
   materials,
-  showOnlyNotifications,
-  showStatusLabel
+  showOnlyNotifications = false,
+  showStatusLabel = false
 }) => {
   const displayedNotifications = showOnlyNotifications
     ? materials.filter(({ showNotificationDot }) => showNotificationDot)
--- a/src/apps/dashboard/dashboard-notification-list/Notifications.tsx
+++ b/src/apps/dashboard/dashboard-notification-list/Notifications.tsx
@@ -14,11 +14,13 @@ export interface NotificationMaterialsList {
 export interface NotificationsProps {
   materials: NotificationMaterialsList[];
   showOnlyNotifications: boolean;
+  showStatusLabel: boolean;
 }
 
 const Notifications: FC<NotificationsProps> = ({
   materials,
-  showOnlyNotifications
+  showOnlyNotifications,
+  showStatusLabel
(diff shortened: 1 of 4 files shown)


src/apps/dashboard/dashboard-notification-list/Notifications.tsx
```diff
@@ -14,11 +14,13 @@ export interface NotificationMaterialsList {
 export interface NotificationsProps {
   materials: NotificationMaterialsList[];
   showOnlyNotifications: boolean;
+  showStatusLabel: boolean;
```
Adamik10, line 17: I normally make these boolean props optional (so something like `showStatusLabel?: boolean`) so that you don't have to specify it when you don't want the status label every time. Then the value is false-y by default when not specifying it. 
Same with `showOnlyNotifications`

Adamik10 (approved):
Very nice! I have one caveat that I think you should consider, but I am approving this nevertheless. It's just a small change. :))

src/apps/dashboard/dashboard-notification-list/Notifications.tsx
clausbruun: @Adamik10 Completely agree, I'm happy to make the changes.

clausbruun: @Adamik10 Could you kindly check again at your convenience? Thank you in advance.

Adamik10: Oh sorry @clausbruun, just saw this now. I am looking at it :)

Adamik10: @clausbruun very nice! 
Now that I'm thinking about it again, if we want to make 100% sure that this won't be causing any issues in the existing code (for example if we further down in the component tree check for this value being specifically `false`, not undefined or null), we can conditionally assign a false value to both `showStatusLabel` and `showOnlyNotifications` when we destructure them on lines 22 and 23 in the Notifications file, and on line 24 in the dashboard-notification file. It would look something like `showOnlyNotifications = false`. This assignment will only happen in case the prop is undefined. 
What do you say to that? :) 
PR: sorry for a late reply. You can ping me on Zulip next time when I'm not seeing notifications here 😉 ❤️

clausbruun: It makes sense, I think. Thank you for the suggestion.

Adamik10 (approved):
Yay! So very nice! 🎉  Great job!

PR #922, opened 2024-02-02
Change various icon on loan list page

spaceo:
#### Link to issue

https://reload.atlassian.net/browse/DDFLSBP-59

#### Description

This PR:
* Changes the list icon in the loan list header

#### Screenshot of the result

icon-list-stacked

Is linked with [this](https://github.com/danskernesdigitalebibliotek/dpl-design-system/pull/478).


src/components/icon-various/icon-various.tsx
```diff
@@ -0,0 +1,8 @@
+import React from "react";
+import VariousIcon from "@danskernesdigitalebibliotek/dpl-design-system/build/icons/collection/Various.svg";
+
+const IconVarious = () => {
+  return <img width={24} height={24} src={VariousIcon} alt="Icon various" />;
```
Adamik10, line 5: Weren't we talking about moving away from using  tags for icons?

Adamik10 (commented):
From the screenshot it seems like the spacing between the items is not consistent with the design system and Figma. Could we investigate this before merging it in?
image
Also, I'm not sure which version of the design system we are using here, but just to be sure - it should be [this one](https://github.com/danskernesdigitalebibliotek/dpl-design-system/pull/482), right?

spaceo:
The work here is solved in [this PR](https://github.com/danskernesdigitalebibliotek/dpl-react/pull/926).

PR #926, opened 2024-02-06
Add loan list skeletons

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFLSBP-432

#### Description
This PR introduces loading skeletons to the loan list page.

#### Screenshot of the result
https://github.com/danskernesdigitalebibliotek/dpl-react/assets/28546954/f1558ffd-7d75-462b-9e11-5cb631136b71

-

### DIFF
--- a/package.json
+++ b/package.json
@@ -144,7 +144,7 @@
     "prop-types": "Since we use former ddb-react components that depend on prop-types we keep this. Should be removed when usage of prop-types is deprecated."
   },
   "dependencies": {
-    "@danskernesdigitalebibliotek/dpl-design-system": "0.0.0-3f502070bec74d8c91f0956b27591e35e33348bf",
+    "@danskernesdigitalebibliotek/dpl-design-system": "0.0.0-3a24ca8b06632650b7c6c3b4676d8354294b9c19",
     "@reach/alert": "^0.17.0",
     "@reach/dialog": "^0.18.0",
     "@reduxjs/toolkit": "^1.9.7",
--- a/package.json
+++ b/package.json
@@ -144,7 +144,7 @@
     "prop-types": "Since we use former ddb-react components that depend on prop-types we keep this. Should be removed when usage of prop-types is deprecated."
   },
   "dependencies": {
-    "@danskernesdigitalebibliotek/dpl-design-system": "0.0.0-c1bd8af5ea614fd9fa958955f9837499e47cfbab",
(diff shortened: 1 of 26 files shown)


spaceo:
If I checkout the branch and run yarn dev I notice following:

The spacing between the buttons in the header is to wide between the buttons:
image
It should be half according to [this change](https://github.com/danskernesdigitalebibliotek/dpl-design-system/pull/482/commits/b211d1eaf54e7016ab27fc4689ee9fdda0971e06#diff-7aba47028435a592ea6ca168018948257788cfb3346d82562a0b98971c0d2184R31).
So it is probably not the correct/updated version of the design system that is being referred to in package.json
 
There is `.dpl-list-buttons__buttons` elements that are nested inside of `.dpl-list-buttons__buttons`.  That is incorrect and should be fixed both in this repo and in the design system:
image

spaceo (changes requested):
Looks good but I discovered som oddities. Please look at them :)

Adamik10:
Oookay - so in regards to the first comment you had about the icon margins - I changed that change back in [this commit](https://github.com/danskernesdigitalebibliotek/dpl-design-system/pull/482/commits/5f16f13825dfda4e0a72d1b72cc7335f8e36e70e) so that it matches the design as was previously.

Aaand the second one - I thought that the deal was to bring the structure over into the design system as it was. This screenshot is taken from release/2024-8-0 and it matches perfectly: 
image

@spaceo

spaceo:
> Oookay - so in regards to the first comment you had about the icon margins - I changed that change back in [this commit](https://github.com/danskernesdigitalebibliotek/dpl-design-system/pull/482/commits/5f16f13825dfda4e0a72d1b72cc7335f8e36e70e) so that it matches the design as was previously.
> 
> Aaand the second one - I thought that the deal was to bring the structure over into the design system as it was. This screenshot is taken from release/2024-8-0 and it matches perfectly: image
> 
> @spaceo

Ok
1. Fair enough. I changed the margin, honestly, by personal preference. I thought there was too much space between the buttons. Let's just leave it be.
2. Sure, the design system should reflect what is in react. But the problem is here that react is wrong. I changed it in my changes, and I think we should fix it while we're at it .because it is wrong.

spaceo (commented):
I wrote some answers. One of them needs to be clarified.

Adamik10:
Great point, @spaceo :)) I have adjusted it now.

spaceo (approved):
LGTM 👍

PR #933, opened 2024-02-08
DDFLSBP-413 - Removed click icon on mouseover on fees groupmodal

Dresse:
#### Link to issue

https://reload.atlassian.net/browse/DDFLSBP-413

#### Description

This PR introduces a check on openDetailsModal that only cursor: pointer classes if openDetailsModel is set.


spaceo (approved):
LGTM 👍

PR #949, opened 2024-02-14
Handling of unregistered users

spaceo:
#### Link to issue

https://reload.atlassian.net/browse/DDFLSBP-403

#### Description

This PR introduces the concept of identifying if a user is unregistered. Which means that the user has not registered via the register formular.

The PR is dependant on the work that has been done in [this PR](https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/760).

### DIFF
--- a/src/apps/create-patron-user-info/CreatePatron.entry.tsx
+++ b/src/apps/create-patron-user-info/CreatePatron.entry.tsx
@@ -59,6 +59,7 @@ const CreatePatronEntry: FC<CreatePatronProps> = () => {
   const u = useUrls();
   const dashboardUrl = u("dashboardUrl");
 
+  // TODO: Do we need to wrap the redirect in a useEffect?
   useEffect(() => {
     // If the user has already been registered, redirect to the dashboard.
     if (!isUnregistered()) {
--- a/src/apps/create-patron-user-info/CreatePatron.entry.tsx
+++ b/src/apps/create-patron-user-info/CreatePatron.entry.tsx
@@ -1,10 +1,12 @@
-import React, { FC } from "react";
+import React, { FC, useEffect } from "react";
 import { withConfig } from "../../core/utils/config";
 import { withText } from "../../core/utils/text";
-import { withUrls } from "../../core/utils/url";
+import { useUrls, withUrls } from "../../core/utils/url";
 import CreatePatron from "./CreatePatron";
 import { getToken, hasToken } from "../../core/token";
 import { GlobalEntryTextProps } from "../../core/storybook/globalTextArgs";
+import { isUnregistered } from "../../core/utils/helpers/user";
+import { redirectTo } from "../../core/utils/helpers/url";
 
 interface CreatePatronConfigProps {
   agencyConfig: string;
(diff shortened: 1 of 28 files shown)


src/apps/create-patron-user-info/CreatePatron.entry.tsx
```diff
@@ -54,15 +56,17 @@ export interface CreatePatronProps
...
-  }
+  useEffect(() => {
+    // If the user has already been registered, redirect to the dashboard.
+    if (!isUnregistered()) {
+      redirectTo(dashboardUrl);
+    }
+  }, [dashboardUrl]);
 
-  return <CreatePatron />;
+  return isUnregistered() ? <CreatePatron /> : null;
```
Adamik10, line 69: Do we need this if statement here if we otherwise already redirect the user to dashboard?

src/apps/create-patron-user-info/CreatePatron.tsx
```diff
@@ -28,21 +23,12 @@ const CreatePatron: FC = () => {
...
     const {
-      attributes: { agencies, cpr: userCpr }
+      attributes: { cpr: userCpr }
     } = userInfo;
 
-    const userWasAlreadyCreated = agencies.some(
-      (agency) => agency.agencyId === agencyId
-    );
-
-    // If the user was already created, redirect to the dashboard.
```
Adamik10, line 38: And we remove it from here, because we added this functionality into the entry file, right?

src/apps/dashboard/dashboard-fees/dashboard-fees.tsx
```diff
@@ -12,11 +12,11 @@ const DashboardFees: FC = () => {
 
   const feesPageUrl = u("feesPageUrl");
   const { data: fbsFees } = useGetFeesV2();
-  const [feeCount, setFeeCount] = useState<number>();
+  const [feeCount, setFeeCount] = useState<number>(0);
   const [totalFeeAmount, setTotalFeeAmount] = useState<string>("0");
 
   useEffect(() => {
-    if (fbsFees) {
+    if (fbsFees && Array.isArray(fbsFees)) {
```
Adamik10, line 19: Very nice.

Adamik10 (approved):
Very good job 👏 I have a couple of remarks, but nothing blocking from what I could see.

src/apps/create-patron-user-info/CreatePatron.entry.tsx
spaceo: This is a very good question.
The thing is that useEffect is activated after rendering. So if the user is NOT unregistered we would get a glimpse of weirdness and even errors if we retun the `CreatePatron` component.
One could ask if we need the useEffect at all. But it works like this and I would rather test that part later. I'll leave a TODO about that.

src/apps/create-patron-user-info/CreatePatron.tsx
spaceo: yup

kasperg (approved):
:+1: All this made sense to me and I am somewhat happy that we can keep the changes limited.

My primary worry here is in regard to the error handling from FBS which we may want to do differently.

src/apps/create-patron-user-info/CreatePatron.entry.tsx
```diff
@@ -54,15 +56,18 @@ export interface CreatePatronProps
...
+  const u = useUrls();
+  const dashboardUrl = u("dashboardUrl");
 
-  // The application using this app should handle the case where the user is not logged in.
-  // Eg by returning 403 Forbidden or redirecting to the login page.
-  if (!userToken) {
-    return null;
-  }
+  // TODO: Do we need to wrap the redirect in a useEffect?
+  useEffect(() => {
```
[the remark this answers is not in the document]
spaceo: Changed comment to be more descriptive.

src/apps/dashboard/dashboard-fees/dashboard-fees.tsx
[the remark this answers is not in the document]
spaceo: I get the point and it makes sense. The things is that maybe, in this case, we actually want some empty views on the various pages. Throwing errors for unregistered users introduces that we also need to handle those errors instead of showing an error boundary.

PR #950, opened 2024-02-15
Distinguish singular/plural on delete reservation

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFLSBP-172

#### Description
In this PR we adjust two text string definitions for when the user is deleting their queued reservations from the dashboard. There was no distinction between singular and plural form depending on how many reservations were deleted before, which is now fixed. 

#### Screenshot of the result
-

[Sibling PR](https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/767)


Dresse (approved):
LGTM!

PR #952, opened 2024-02-16
Don't show material type badge on reservation page if no data is present

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFLSBP-234

#### Description
We no longer show material type on reservation page if no data is present.

#### Screenshot of the result
-

-


Dresse (approved):
LGTM

PR #964, opened 2024-02-20
Move global text argument reservationListDigitalPickupText into global

Adamik10:
#### Link to issue
-

#### Description
While testing for [this task](https://reload.atlassian.net/browse/DDFLSBP-170) I have discovered that the dashboard app is failing and ´reservationListDigitalPickupText´ needs to be in global text args definition.

#### Screenshot of the result
-

-


spaceo (approved):
LGTM 👍

PR #967, opened 2024-02-21
Fix/reservation and loan skeletons

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFLSBP-456

#### Description
The skeleton screens were behaving a bit strangely where they would still show even when we already had some items to show on the reservation list. And on the loan list they would flicker between skeletons and loaded items with a noticable empty page screen. This PR fixes both issues.

#### Screenshot of the result
-

-


spaceo (approved):
LGTM :+1

PR #971, opened 2024-02-22
DDFFORM 313

LasseStaus:
- Use release 2024.9.0 branch of design system
- Add component & Story MaterialGridAutomatic
- Add Entry, mount  for MaterialGridAutomatic
- Add skeleton component for MaterialGridAutomatic

#### Link to issue

[DDFFORM-313](https://reload.atlassian.net/browse/DDFFORM-313)

Other relevant PRS: 

- CMS PR: https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/796
- Design system PR: https://github.com/danskernesdigitalebibliotek/dpl-design-system/pull/517 

#### Description

This PR adds a react app <MaterialGridAutomatic/> which displays a list of up to 32 materials based on a search string. 

At the time of this PR, a search string is generated manually by copying the 'CQL Search String' from an advanced search. 

#### Screenshot of the result

image

#### Comments or questions

Be thorough when looking through the configurations regarding config-branches or the use of the hook for querying the materials.

PR #973, opened 2024-02-23
DDFFORM 314

LasseStaus:
#### Link to issue

[DDFFORM-314](https://reload.atlassian.net/browse/DDFFORM-314)

#### Description

This PR adds a grid-display of recommended materials, that can be selected added by editors, by adding the workIds of the materials. 

This PR is very similar to the [MaterialGridAutomatic](https://github.com/danskernesdigitalebibliotek/dpl-react/pull/971) and depends on the CSS that was made, as part of that ticket. 

#### Screenshot of the result

image

### DIFF
--- a/src/apps/material-grid/manual/MaterialGridManual.tsx
+++ b/src/apps/material-grid/manual/MaterialGridManual.tsx
@@ -1,8 +1,6 @@
 import * as React from "react";
-import { useEffect } from "react";
 import { WorkId } from "../../../core/utils/types/ids";
 import MaterialGrid from "../MaterialGrid";
-import MaterialGridSkeleton from "../MaterialGridSkeleton";
 import { calculateAmountToDisplay } from "../materiel-grid-util";
 
 export type MaterialGridManualProps = {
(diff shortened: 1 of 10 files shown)


src/apps/material-grid-manuel/MaterialGridManuel.tsx
kasperg, line 1: Please consider how this can be refactored.

You implemented this directly on `release/2024-9-0-1` instead of on top of #971. Given the similarity of the two components think we end up with a large amount of duplicated code for the component.

While we may need to have some degree of duplication for `.dev.tsx` and `.entry.tsx` I think there is able room to share components between the two versiond of the grid.

With that in mind I think you should give this another go.

One suggestion would be to restructure the code with something like this:

```
/apps/material-grid
 - MaterialGrid.tsx
 - MaterialGridManuelSkeleton.tsx
 - /automatic
   - MaterialGridManuel.dev.tsx
   - MaterialGridManuel.entry.tsx
   - MaterialGridManuel.mount.tsx
 - /manual
   - MaterialGridAutomatic.dev.tsx
   - MaterialGridAutomatic.entry.tsx
   - MaterialGridAutomatic.mount.tsx
```

kasperg (changes requested):

src/apps/material-grid/manual/MaterialGridManual.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+  useEffect(() => {
+    // Materials must be parsed from string to WorkId[] array as they are
+    // received as a string from the backend.
+    if (typeof materialIds === "string") {
+      setParsedMaterials(JSON.parse(materialIds));
+    }
+  }, [materialIds]);
+
+  if (!parsedMaterialIds.length || typeof parsedMaterialIds === "string") {
```
kasperg, line 30: Consider moving this handling of `materialIds`

I do not understand how TypeScript will accept a seemingly irrelevant type check for `materialIds` which we have declared should be `WorkId[]`.

I suggest moving the handling to the `.entry.tsx` file which is intended for CMS/non-Storybook usage. Here we can correctly declare the type of `materialIds` to be a string and then parse it to be an array of material ids we can pass along to the shared `MaterialGridManual` component.

kasperg (approved):
:+1: This looks good and is a nice continuation of the work in #971.

I only have a single comment for your consideration.

src/apps/material-grid/manual/MaterialGridManual.tsx
LasseStaus: Makese sense. 

I refactored and moved the logic to the entry file. Have a look 🚀

kasperg:
Superseded by #987.

PR #1005, opened 2024-03-08
Insert retry delay in installs in branch release workflow

spaceo:
We have the suspicion that the design system npm packages takes time in being ready. Let's wait a bit more for them to be available...

### DIFF
--- a/.github/workflows/create-release-on-branch-changes.yml
+++ b/.github/workflows/create-release-on-branch-changes.yml
@@ -54,7 +54,7 @@ jobs:
         with:
           command: yarn install --frozen-lockfile
           attempt_limit: 3
-          attempt_delay: 2000
+          attempt_delay: 360000
 
       - name: Build artifacts
         run: yarn build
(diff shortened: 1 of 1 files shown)


.github/workflows/create-release-on-branch-changes.yml
```diff
@@ -54,7 +54,7 @@ jobs:
         with:
           command: yarn install --frozen-lockfile
           attempt_limit: 3
-          attempt_delay: 2000
```
kasperg, line 57: As I read it this will retry 3 times with a 10 minutes separation.

10 minutes is a long time to way. How about changing it to 15 (or 20 or 30) times with 1 minute separation?

kasperg (approved):
:+1: Let's try this whatever configuration you decide on.

.github/workflows/create-release-on-branch-changes.yml
spaceo: Not a bad idea. Let's try that.

PR #1006, opened 2024-03-08
DDFFORM 369

LasseStaus:
#### Link to issue

[DDFFORM-369](https://reload.atlassian.net/browse/DDFFORM-369)

This is connected to [CMS PR](https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/867)

#### Description

This PR adds an optional materialType field to all components currently using the recommendedMaterial for displaying materials. 

Remember to read the commit messages and consider any question specified there. 

_Current styling for the material Grids are only in develop, which is not merged onto form-sprint-8_

Consider the used type casting. I was not sure what alternative to this there was. 

Consider wether the combined type in materialGrid `MaterialGridItemProps`should be a general type and used for all components using recommendedMaterial, or if the current implmenetation is better. I am still unsure about this, but i decided to initiately use this as i found it "cleaner".

### DIFF
--- a/src/apps/material-grid/manual/MaterialGridManual.dev.tsx
+++ b/src/apps/material-grid/manual/MaterialGridManual.dev.tsx
@@ -10,6 +10,41 @@ import MaterialGridManual, {
   MaterialGridManualEntryProps
 } from "./MaterialGridManual.entry";
 
+// 31 materials. Intentionally not using 32 in order to demonstrate the
+// logic for only displaying intervals of 4 materials.
+const materials = [
+  { wid: "work-of:870970-basis:25660722", materialType: "bog" },
+  { wid: "work-of:870970-basis:22383590", materialType: "e-bog" },
+  { wid: "work-of:870970-basis:25932625", materialType: "film" },
+  { wid: "work-of:870970-basis:26264340", materialType: "musik (online)" },
+  { wid: "work-of:870970-basis:52646251", materialType: "lydbog" },
+  { wid: "work-of:870970-basis:26856353", materialType: "artikel" },
+  { wid: "work-of:870970-basis:27275745", materialType: "tegneserie (online)" },
+  { wid: "work-of:870970-basis:22383590", materialType: "tidsskrift" },
+  { wid: "work-of:870970-basis:29788596", materialType: "cd" },
+  { wid: "work-of:870970-basis:52646251", materialType: "podcast" },
+  { wid: "work-of:870970-basis:50689360", materialType: "film (online)" },
+  { wid: "work-of:870970-basis:22383590", materialType: "lydbog (cd-mp3)" },
+  { wid: "work-of:870970-basis:46510534", materialType: "artikel (online)" },
+  { wid: "work-of:870970-basis:134877804", materialType: "tegneserie" },
+  { wid: "work-of:870970-basis:54129807", materialType: "tidsskrift (online)" },
+  { wid: "work-of:870970-basis:52646251", materialType: "billedbog" },
+  { wid: "work-of:870970-basis:25660722", materialType: "billedbog (online)" },
+  { wid: "work-of:870970-basis:25932625", materialType: "lydbog (online)" },
+  { wid: "work-of:870970-basis:26264340", materialType: "musik (online)" },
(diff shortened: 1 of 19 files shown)


src/apps/material/helper.ts
```diff
@@ -454,6 +457,35 @@ export const useGetHoldings = ({
...
 
+export const getManifestationBasedOnType = (
+  work: Work,
+  materialType: DisplayMaterialType
+): Manifestation => {
+  const { bestRepresentation, all } = work.manifestations;
+
+  // Retrieves the material type of the best representation.
+  const bestRepresentationMaterialType =
+    getManifestationMaterialTypes(bestRepresentation);
```
kasperg, line 468: Consider this level of commenting. 

You are basicly writing out in plain english what the following line of code does even though it is not very complex.

src/apps/material/helper.ts
```diff
@@ -454,6 +457,35 @@ export const useGetHoldings = ({
...
+
+  // Checks if the best representation matches the desired material type.
+  if (materialType === bestRepresentationMaterialType) {
+    return bestRepresentation;
+  }
+
+  // Filters and sorts the manifestations if the best representation does not match.
+  const filteredByType = filterManifestationsByType(materialType, all);
+  const sortedByYear = orderManifestationsByYear(filteredByType);
```
kasperg, line 477: Consider dropping the explicit ordering.


`filterManifestationsByType()` already does ordering by year (even though it is not evident by the title). You replace the ordering by a comment.

src/apps/material/helper.ts
```diff
@@ -454,6 +457,35 @@ export const useGetHoldings = ({
...
+
+  // Returns the newest manifestation based on type if available.
+  const newestManifestationBasedOnType = sortedByYear[0];
+  if (newestManifestationBasedOnType) {
```
kasperg, line 481: Consider using `first()` from lodash instead of `[0]`.

[I have described why in Zulip](https://reload.zulipchat.com/#narrow/stream/240325-DDF/topic/.5B0.5D.20.3D.3E.20_.2Ehead.28.29/near/339889428) but I see that you have implemented a check which is good.

src/apps/recommended-material/RecommendedMaterial.dev.tsx
```diff
@@ -17,6 +18,10 @@ export default {
       defaultValue: "work-of:870970-basis:136336282",
       control: { type: "text" }
     },
+    materialType: {
+      defaultValue: "bog",
+      control: { type: "select", options: DisplayMaterialTypeOptions }
+    },
```
kasperg, line 24: Consider adding a story without a preselected material type to showcase the consequences of using this.

Using a work where we know that there will be a significant difference in the presentation between material types would also be good. This might already be the case. If so: Carry on :smile:.

src/apps/material-grid/manual/MaterialGridManual.dev.tsx
```diff
@@ -10,6 +10,40 @@ import MaterialGridManual, {
...
+  { wid: "work-of:870970-basis:29788596", materialType: "podcast" },
+  { wid: "work-of:870970-basis:52646251", materialType: "film (online)" },
+  { wid: "work-of:870970-basis:50689360", materialType: "lydbog" },
+  { wid: "work-of:870970-basis:53045650", materialType: "tegneserie (online)" },
+  { wid: "work-of:870970-basis:46510534", materialType: "tidsskrift" },
+  { wid: "work-of:870970-basis:134877804", materialType: "cd" },
+  { wid: "work-of:870970-basis:54129807", materialType: "podcast" },
+  { wid: "work-of:870970-basis:52646251", materialType: "film (online)" },
+  { wid: "work-of:870970-basis:25660722", materialType: "lydbog" }
+];
```
kasperg, line 46: Consider not setting the `materialType` on at least one of the items.

kasperg (approved):
:+1: I have a few things that you might consider but in general I think this could be merged as is.

src/apps/recommended-material/RecommendedMaterial.dev.tsx
LasseStaus: I found one with 3 differences 🚀 

Also added a story with an "undefined" materialType. 👀

PR #1023, opened 2024-03-13
Implementation of a new react app for displaying opening hours from CMS with FullCalendar integration - Ddfform 414 

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFFORM-414

#### Description
This pull request adds a new React app to be used for displaying/adding opening hours from the CMS on a weekly or monthly view.

We have used FullCalendar to solve this task
https://fullcalendar.io/

Although the Opening Hours API is not yet complete, it already exposes a Swagger spec that Orval has used to generate some new hooks, which we use to fetch data. 

Since the API does not yet expose anything, we have used WireMock to display data from `https://dpl-cms.docker/dpl_opening_hours`.

We also demonstrate the ability to add new opening hours, but these are only saved in local state for now.

#### Screenshot of the result


orval.config.ts
```diff
@@ -139,7 +139,7 @@ export default defineConfig({
     },
     input: {
       target:
-        "https://raw.githubusercontent.com/danskernesdigitalebibliotek/dpl-cms/develop/openapi.json",
+        "https://raw.githubusercontent.com/danskernesdigitalebibliotek/dpl-cms/e1edb7e35beacb93e78eef0541c8ccebad68d6f1/openapi.json",
```
kasperbirch1, line 142: This must be changed back when this pr is merged

https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/890

src/core/dpl-cms/dpl-cms.ts
```diff
@@ -29,7 +35,7 @@ type AwaitedInput<T> = PromiseLike<T> | T;
 type Awaited<O> = O extends AwaitedInput<infer T> ? T : never;
 
 /**
- * @summary Get campaign matching search result facets
+ * @summary Hent kampagne som matcher facetter fra søgeresultatet
```
LasseStaus, line 38: Tænker dette burde forblive engelsk

src/core/storybook/serviceUrlArgs.ts
```diff
@@ -14,7 +14,7 @@ export default {
   },
   [serviceUrlKeys.dplCms]: {
     name: "Base url for the DPL CMS API",
-    defaultValue: "https://dpl-cms.docker",
+    defaultValue: process.env.CMS_BASEURL ?? "https://dpl-cms.docker",
```
LasseStaus, line 17: Is this correct? 

I wonder why there is a fallback for an .env variable, which defaults to what i assume you are supposed to set in your .env. 

Also in the .env we have another url than `https://dpl-cms.docker`

Not saying anything is necessarily wrong here, just wondering.

src/apps/opening-hours-editor/helper.ts
```diff
@@ -1,15 +1,25 @@
 import { DplOpeningHoursListGET200Item } from "../../core/dpl-cms/model";
 
-const formatCmsEventsToFullCalendar = (
+function formatDateTimeString(date: string, time: string) {
+  return `${date}T${time}:00`;
+}
+
+export function createCmsEventId(title: string, startDay: Date) {
+  return `${title}-${startDay.toISOString()}`;
+}
```
LasseStaus: I'm unsure if this is necessary. The way i see it, these ID's are not necessary to differentiate what is being created/edited in the Calender, and i would think that the DB will create necessary ids when the openings hours are created. 

I could be wrong, but i think you should check this.

src/apps/opening-hours-editor/helper.ts
```diff
@@ -25,6 +25,10 @@ export const formatCmsEventsToFullCalendar = (
   });
 };
 
+const isSameDay = (startDay: Date, endDay: Date) => {
+  return startDay.toDateString() === endDay.toDateString();
+};
+
```
LasseStaus: 👍  🧑‍🚒

src/apps/opening-hours-editor/OpeningHoursEditor.tsx
```diff
@@ -41,7 +41,7 @@ const OpeningHoursEditor: React.FC = () => {
       }
       eventClick={handleEventClick}
       eventContent={(eventInput) =>
-        renderOpeningHoursEditorEventContent({
+        OpeningHoursEditorEventContent({
```
LasseStaus, line 28: In regards to your commit message, is it something like this you are looking for ? 

```
     eventContent={(eventInput) => {
        return (
          <OpeningHoursEditorEventContent
            eventInput={eventInput}
            handleEventRemove={() =>
              handleEventRemove({ eventToRemove: eventInput, events })
            }
          />
        );
      }}
      
```

LasseStaus (approved):
🥳 🎈  

This looks good to me as a starting point for all the rest of the functionality.  

We should make a ticket for styling stuff and get started on that soon too.  

I have added very few comments you can take a look at. 

I will highlight here also, that i think there should be some default color for the calendar, and that the category colors should be opt-in. This can also be added later. 


In terms of structure, its' always hard to know when starting something like this, so naturally there will be a lot of rearranging. For the future tho, where possible, see if we can remove / change some commits. I think some of this could have been  resetted and just committed anew instead of restructuring in new commits. 

Its' not a problem and not ment negatively though 😃 

nice one!

src/apps/opening-hours-editor/helper.ts
kasperbirch1: It is necessary to edit events, but the id should come from the API in the future.

src/core/storybook/serviceUrlArgs.ts
kasperbirch1: This is the way we handle working with WireMock. We will only add the `CMS_BASEURL` (via environment variables) if we have to fetch against WireMock.

src/core/dpl-cms/dpl-cms.ts
kasperbirch1: It is auto generated code that comes from https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/890/files

.docker/wiremock/cms/mappings/delete-event-f52a1b96-38d2-479a-aeaf-65367637d365.json
```diff
@@ -0,0 +1,18 @@
+{
+  "id": "f52a1b96-38d2-479a-aeaf-65367637d365",
+  "name": "Delete event",
+  "request": {
+    "urlPathPattern": "/dpl_opening_hours/[0-9]+",
+    "method": "DELETE"
+  },
+  "response": {
+    "status": 200,
+    "body": "Event deletet",
```
kasperg, line 10: Consider updating this. It does not follow the OpenAPI specification.

src/apps/opening-hours-editor/helper.ts
```diff
@@ -0,0 +1,112 @@
...
+  return {
+    id: Number(event.id),
+    category: {
+      title: event.title,
+      color
+    },
+    date: startDate.format("YYYY-MM-DD"),
+    start_time: startDate.format("HH:mm"),
+    end_time: endDate.format("HH:mm"),
+    branch_id: 0
```
kasperg, line 46: Please add a comment about why you set `branch_id` to 0.

I expect it is to conform with the type but you could also specify the type as being `DplOpeningHoursListGET200Item` without the branch and then force the caller to set the value before it is passed to the API.

src/apps/opening-hours-editor/OpeningHoursEditor.entry.tsx
```diff
@@ -0,0 +1,26 @@
...
+  openingHoursEditorCategoriesConfig: string;
+  openingHoursBranchIdConfig: string;
+}
+
+const OpeningHoursEditorEntry: React.FC<
+  OpeningHoursEditorEntryTextProps &
+    OpeningHoursEditorType &
+    OpeningHoursEditorEntryConfigProps
+> = ({ useWireMockStartDate }) => (
+  <OpeningHoursEditor useWireMockStartDate={useWireMockStartDate} />
```
kasperg, line 23: Please use the start date as a prop instead of a boolean referencing Wiremock.

In general we do not want Wiremock specific code inside our components. It should be controlled using Storybook controls or the like as it is used for development. Then in the Storybook we can set a default for the start date prop and add a comment referencing Wiremock there.

src/apps/opening-hours-editor/DialogFomularEdit.tsx
```diff
@@ -0,0 +1,91 @@
...
+  const t = useText();
+
+  const handleSubmit: EventFormOnSubmitType = (
+    category,
+    startTime,
+    endTime
+  ) => {
+    if (!eventInfo.start || !eventInfo.end) {
+      // eslint-disable-next-line no-alert
+      alert("Invalid event");
```
kasperg, line 38: :+1: I think it is OK that we use simple alerts for error handling at the moment.

Consider making the text translatable since it is editor facing.

Also I wonder if it would make sense to log more detailed information about the error state.

src/apps/opening-hours-editor/DialogFomularAdd.tsx
kasperg, line 1: Please adjust the name here and in DialogFormularEdit.

`Formular` is not a proper English word.

src/apps/opening-hours-editor/EventForm.tsx
```diff
@@ -0,0 +1,118 @@
...
+    }
+  };
+
+  return (
+    <form onSubmit={handleSubmit} className="opening-hours-editor-form">
+      <label
+        className="opening-hours-editor-form__label"
+        htmlFor="event-form-title"
+      >
+        Title:
```
kasperg, line 60: Please ensure that the texts here are translateable.

src/apps/opening-hours-editor/helper.ts
```diff
@@ -0,0 +1,112 @@
...
+
+  return {
+    end: adjustedEndDay,
+    endStr: formatDateStr(adjustedEndDay)
+  };
+};
+
+export const extractTime = (date: Date) => {
+  return dayjs(date).format("HH:mm");
+};
```
kasperg, line 107: Consider making a new type to handle times.

There are lot of places where we are basicly passing strings around. It would be preferable if we could ensure they were proper times.

src/apps/opening-hours-editor/helper.ts
```diff
@@ -0,0 +1,112 @@
+import dayjs from "dayjs";
+import { EventInput } from "@fullcalendar/core";
+import { EventImpl } from "@fullcalendar/core/internal";
+import { DplOpeningHoursListGET200Item } from "../../core/dpl-cms/model";
+
+const formatDateTimeString = (date: string, time: string): string => {
+  return `${date}T${time}:00`;
```
kasperg, line 7: Consider using dayjs for this instead of simple string manipulation.

If we need the data to be in a specific format e.g. ISO8601 they also have helpers for this.

src/apps/opening-hours-editor/helper.ts
```diff
@@ -0,0 +1,112 @@
...
+    },
+    date: startDate.format("YYYY-MM-DD"),
+    start_time: startDate.format("HH:mm"),
+    end_time: endDate.format("HH:mm"),
+    branch_id: 0
+  };
+};
+
+export const formatDateStr = (date: Date) => {
+  return dayjs(date).format("YYYY-MM-DDTHH:mm:ssZ");
```
kasperg, line 51: This is ISO8601 [as supported by dayjs](https://day.js.org/docs/en/display/as-iso-string) and required by the API. I would prefer that we used this.

src/components/dialog/Dialog.tsx
```diff
@@ -0,0 +1,39 @@
+// The dialog element allready has keyboard support
+/* eslint-disable jsx-a11y/no-noninteractive-element-interactions */
+/* eslint-disable jsx-a11y/click-events-have-key-events */
```
kasperg, line 3: :+1: Excellent commenting on why is it relevant t disable these.

src/components/dialog/Dialog.tsx
```diff
@@ -0,0 +1,39 @@
...
+            closeDialog();
+          }
+        }}
+      >
+        <button
+          type="button"
+          onClick={closeDialog}
+          className="dialog__close-button"
+        >
+          <img src={iconCross} alt="close icon" />
```
kasperg, line 32: Please provide a proper alt text.

Close icon is not suitable. We might already have something usable elsewhere. You could also make the `alt` text empty and provide an `aria-label` with text for the button.

src/components/dialog/useDialog.tsx
```diff
@@ -0,0 +1,42 @@
+import { useRef, useState } from "react";
+
+declare global {
+  interface HTMLDialogElement {
+    showModal(): void;
+    close(): void;
+  }
+}
```
kasperg, line 8: Consider adding a reference/link to some documentation for this or replacing this with a third party package.

I wonder why we need to do this ourselves and how you came up with this.

src/components/dialog/useEscapeKey.tsx
```diff
@@ -0,0 +1,19 @@
+import { useEffect } from "react";
+
+const useEscapeKey = ({ closeDialog }: { closeDialog: () => void }) => {
```
kasperg, line 3: Consider replacing this/some of this with the hooks provided by react-use.

We already have it in the project https://github.com/streamich/react-use?tab=readme-ov-file

src/core/dpl-cms/model/dplOpeningHoursGETFormat.ts
```diff
@@ -0,0 +1,15 @@
+/**
+ * Generated by orval v6.8.1 🍺
```
kasperg, line 2: Please consider if this and other files like this are still relevant.

They are marked with the old release of Orval. Consequently it seems as if they have not been created/overridden after upgrading. This makes me wonder if they are still being used.

This goes for a good deal of the new files here.

src/core/utils/types/ids.ts
```diff
@@ -14,6 +14,7 @@ export type GuardedAppId =
   | "recommended-material"
   | "recommendation"
   | "material-grid-automatic"
-  | "material-grid-manual";
+  | "material-grid-manual"
+  | "opening-hours-editor";
```
kasperg, line 18: I do not think we need the extra app here.

kasperg (approved):
Lots of code, lots of stuff here and there for you to look into.

I will approve this so we can get this merged once you approve it.

src/apps/opening-hours-editor/helper.ts
kasperbirch1: We will handle it in a future pull request

PR #1024, opened 2024-03-13
Delete text on user me loans

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFLSBP-463

#### Description

Removes subheadline from loan list modal about where the loan can be returned.

#### Screenshot of the result
-

-


Dresse (changes requested):
Good that you have removed it in dpl-react, but I also think it would be best to remove it from dpl-cms.

Also shouldn't it also be removed from the groupModalArgs.ts file?
https://github.com/danskernesdigitalebibliotek/dpl-react/blob/sprint/lsbp-12/src/core/storybook/groupModalArgs.ts

Dresse (approved):
LGTM

Dresse:
> Jeez, thx for this! I am a bit slow when it comes to thinking these days 😅

No worries! 😄

PR #1031, opened 2024-03-15
Don't show digital material quota if not logged in home municipality

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFLSBP-439

#### Description
Don't show digital material loan quota on the user profile page if the user is logged in in a different municipality then their home one.

#### Screenshot of the result
-

-


Dresse (approved):
LGTM 👍

PR #1045, opened 2024-03-19
Fix fees and compensations modals

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFLSBP-411

#### Description
In this PR we start to distinguish between fees and compensations when creating fee modals.

Before, if there ever were materials that had both fees and compensations on them, clicking on either one of them would result in opening the same modal. The other item would then have incorrect items inside the modal. Creating separate modal IDs for each fee/compensation fixes this problem.

#### Screenshot of the result
-

-


Dresse (approved):
I think it looks good :) 

I do however have some generel questions (because I'm curious 😄 )

1. What is the faustId used for?
2. Is feeId a part of the feeData?

PR #1046, opened 2024-03-20
Add possibility for loading "global" material data at the material page

spaceo:
#### Link to issue

https://reload.atlassian.net/browse/DDFLSBP-495

#### Description

This PR introduces the concept of loading material data either locally or globally on a material page depending on where the data is present.
And some visibility rules are also added depending of the material is considered being local or global.

##### First commit message

Both for the work query handling and visibility of material page elements
we need to introduce a way to know if a loaded material is from a local
library or the global scope.

So a work loading hook `useGetWork()` is introduced.
It starts with trying to load local material data and otherwise from a
global scope.

The `global`/`local` scope is controlled via the FBI profiles (`next`,
`next-present`).
* `next` is considered being `local`
* `next-present` is considered being global.


src/apps/material/material.dev.tsx
```diff
@@ -902,3 +902,8 @@ export const overbygningsMatriale = Template.bind({});
 overbygningsMatriale.args = {
   wid: "work-of:870970-basis:44926407"
 };
+
+export const globalMaterial = Template.bind({});
```
Adamik10, line 906: Great that you're adding a story 💯

src/apps/material/material.tsx
```diff
@@ -194,25 +193,27 @@ const Material: React.FC<MaterialProps> = ({ wid }) => {
...
-                  key={manifestation.pid}
-                  manifestation={manifestation}
-                  workId={wid}
-                />
-              );
-            }
-          )}
-        </>
-      </MaterialDisclosure>
+      {workType === "local" && (
```
Adamik10, line 196: So we don't show editions for global materials? Maybe it would be nice with a comment explaining the logic behind this.

src/core/dbc-gateway/queryMap.ts
```diff
@@ -10,11 +10,12 @@ export default {
   searchFacet: serviceUrlKeys.fbiSearch,
   searchWithPagination: serviceUrlKeys.fbiSearch,
   suggestionsFromQueryString: serviceUrlKeys.fbiSearch,
+  getMaterial: serviceUrlKeys.fbiSearch,
```
Adamik10, line 13: Very nitpicky, but I think these `get...` lines were ordered alphabetically 🙈 But I see the rest isn't, so maybe who cares(?)

Adamik10 (approved):
Very good job! I am in love with the hook you made 💟 
Got some smaller comments you can think about, but nothing big. 
OUT WITH IT! 🎉

src/core/dbc-gateway/queryMap.ts
spaceo: No you're right. I made some changes to where getMaterial belonged and did not sort the lines afterwards. Iøll take care of it..

PR #1047, opened 2024-03-20
Differenciate between blacklistBranchesInstantLoan & pickupBlacklist

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFLSBP-491

#### Description
In the reservation modal - when changing between pickup branches, we didn't ask for the correct configuration - forgetting blacklistedPickupBranchesConfig branches. Now we add those on top of already used instant loan blacklisted branches, and as a result when an administrator changes this config in the backend, the changes take effect in the reservation modal as well.

#### Screenshot of the result
-
-


clausbruun (approved):
Nicely found!

PR #1048, opened 2024-03-20
POC: Add `Dialog` - Ddfform 500 

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFFORM-500

#### Description
This is a proof of concept for utilizing the "new" native HTML `dialog` element with the `open` attribute. By default, it is accessible with focus and keyboard navigation.

In this version, it triggers a dialog/modal upon clicking on an event. The dialog includes a close button (the dialog also closes when pressing the escape key or clicking on the backdrop) and a button that currently displays an alert with the event's updated data. `handleEventEditing` will later be an update request to the API.

**Note:**
Run `docker compose up` in dpl-react to spin up wiremock

I found some inspiration for the dialog setup in this video, but I have made changes to keep the logic in a custom React hook for simplicity and readability.
https://www.youtube.com/watch?v=YwHJMlvZRCc&t=809s

#### Screenshot of the result

https://github.com/danskernesdigitalebibliotek/dpl-react/assets/49920322/72082f38-86ea-4351-b3a8-b36ac91e1c23


.docker/wiremock/cms/mappings/dummy-479e26d6-5197-4d8f-a2dd-fefe64aabc14.json
```diff
@@ -7,7 +7,7 @@
   },
   "response" : {
     "status" : 200,
-    "body" : "[\n  {\n    \"category\": {\n      \"title\": \"Open\"\n    },\n    \"date\": \"2024-03-14\",\n    \"start_time\": \"09:00\",\n    \"end_time\": \"17:00\"\n  },\n    {\n    \"category\": {\n      \"title\": \"Open\"\n    },\n    \"date\": \"2024-03-15\",\n    \"start_time\": \"09:00\",\n    \"end_time\": \"17:00\"\n  }\n]",
+    "body" : "[\n  {\n    \"category\": {\n      \"title\": \"Open\"\n    },\n    \"date\": \"2024-03-22\",\n    \"start_time\": \"09:00\",\n    \"end_time\": \"17:00\"\n  },\n    {\n    \"category\": {\n      \"title\": \"Open\"\n    },\n    \"date\": \"2024-03-23\",\n    \"start_time\": \"09:00\",\n    \"end_time\": \"17:00\"\n  }\n]",
```
spaceo, line 10: Just a note about the naming of this file:
I guess it is temporary that the file is called dummy-* 
If it is going to be used in the future setup please remeber to call it something meaningful :)

src/apps/opening-hours-editor/Dialog.tsx
```diff
@@ -0,0 +1,39 @@
...
+  children: React.ReactNode;
+  closeDialog: () => void;
+};
+
+const Dialog = forwardRef<HTMLDialogElement, DialogType>(
+  ({ children, closeDialog }, ref) => {
+    return (
+      <dialog
+        ref={ref}
+        // close dialog when clicking outside of it (::backdrop pseudo-element)
```
spaceo, line 16: Remember to capitalize first letter and end comment with a period

src/apps/opening-hours-editor/Dialog.tsx
```diff
@@ -0,0 +1,39 @@
+// The dialog element allready has keyboard support
```
spaceo, line 1: Would it be an idea to make it a global component?
It seems very generic and we could use it in other contexts/apps?
As far as I remember we do not have a dialog component already...

src/apps/opening-hours-editor/Dialog.tsx
```diff
@@ -0,0 +1,39 @@
...
+        }}
+      >
+        <button
+          type="button"
+          onClick={closeDialog}
+          style={{
+            border: "1px solid black",
+            padding: "5px",
+            background: "grey"
+          }}
```
spaceo, line 30: This should be a class and the styling should be defined in the design system

src/apps/opening-hours-editor/DialogFomular.tsx
```diff
@@ -0,0 +1,70 @@
...
+  };
+
+  return (
+    <form
+      onSubmit={handleSubmit}
+      style={{ display: "grid", marginTop: "20px" }}
+    >
+      <label htmlFor="title">Title:</label>
+      <input
+        id="title"
```
spaceo, line 39: Isn't `title` a too generic id?
How about `dialog-formular-title`.
Same goes for the rest of the ids...

src/apps/opening-hours-editor/DialogFomular.tsx
```diff
@@ -0,0 +1,70 @@
...
+        id="end-time"
+        type="time"
+        value={endTime}
+        onChange={(e) => setEndTime(e.target.value)}
+        min={startTime}
+        max="00:00"
+      />
+      <button
+        type="submit"
+        style={{ backgroundColor: "green", marginTop: "20px" }}
```
spaceo, line 62: Design system class here instead (I know it is a POC, just for you to remember :))

src/apps/opening-hours-editor/DialogFomular.tsx
```diff
@@ -0,0 +1,70 @@
...
+    const endDate = updateEventTime(eventInfo.endStr, endTime);
+    eventInfo.setDates(startDate, endDate);
+
+    handleEventEditing(eventInfo);
+  };
+
+  return (
+    <form
+      onSubmit={handleSubmit}
+      style={{ display: "grid", marginTop: "20px" }}
```
spaceo, line 35: Design system class here instead (I know it is a POC, just for you to remember :))

src/apps/opening-hours-editor/DialogFomular.tsx
```diff
@@ -0,0 +1,70 @@
...
+  const [endTime, setEndTime] = useState(extractTime(eventInfo.endStr));
+
+  const handleSubmit = (e: React.FormEvent<HTMLFormElement>) => {
+    e.preventDefault();
+
+    eventInfo.setProp("title", title);
+    const startDate = updateEventTime(eventInfo.startStr, startTime);
+    const endDate = updateEventTime(eventInfo.endStr, endTime);
+    eventInfo.setDates(startDate, endDate);
+
```
spaceo, line 28: I think we should stretch us as far as possible to pass date objects around. Or dayjs objects since that has been the chosen library so far (I dont care which we choose to use as long as we pass objects around).
So as soon as possible we convert date/time to objects. So we are never in doubt what we are dealing with and we can make use of a more strict typing, instead of just strings.

src/apps/opening-hours-editor/helper.ts
```diff
@@ -37,3 +37,17 @@ export const adjustEndDateToStartDay = (startDay: Date, endDay: Date) => {
...
+  const minutes = date.getMinutes().toString().padStart(2, "0");
+  return `${hours}:${minutes}`;
+};
+
+export const updateEventTime = (dateStr: string, timeStr: string) => {
+  const [hours, minutes] = timeStr.split(":").map(Number);
+  const date = new Date(dateStr);
+  date.setHours(hours, minutes);
+  return date;
+};
```
spaceo, line 53: As written earlier. I really think we should handle time/dates as much as we can as objects and use methods on those objects for for formatting.
And if possible keep the time/dates as objects when passing them around between functions.

src/apps/opening-hours-editor/useDialog.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+  const dialogRef = useRef<HTMLDialogElement>(null);
+
+  const closeDialog = () => {
+    dialogRef.current?.close();
+  };
+
+  const openDialogWithContent = (content: React.ReactNode) => {
+    setDialogContent(content);
+    dialogRef.current?.showModal();
```
spaceo, line 21: What if `current` is undefined?

src/apps/opening-hours-editor/useOpeningHours.tsx
```diff
@@ -48,16 +49,9 @@ const useOpeningHours = () => {
...
     // eslint-disable-next-line no-alert
-    const newTitle = prompt(
-      "Enter a new title for this event",
-      clickInfo.event.title
-    );
-
-    if (newTitle) {
-      clickInfo.event.setProp("title", newTitle);
-    }
+    alert(JSON.stringify(eventInfo, null, 2));
```
spaceo, line 54: For POC use I guess?

spaceo (commented):
Looking good! 
I have written some thoughts...

PR #1049, opened 2024-03-21
Cleanup query map naming

spaceo:
It makes more sense to call the scopes local/global instead of search/material because local/global illustrates if the material comes from the local library inventory or the cross library inventory.

NB!!: [This](https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/919) should be merged first.


Dresse (approved):
LGTM

PR #1072, opened 2024-04-02
Fix/availability blacklisting

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFLSBP-493

#### Description
This PR makes changes to fetching availability data for works - for find on shelf modal, and the reserve button on material page for materials that can be loaned from another library (overbygningsmaterialer). We forgot to take the Drupal config blacklisted availability branches into account before. 

#### Screenshot of the result
-

-

### DIFF
--- a/src/apps/material/helper.ts
+++ b/src/apps/material/helper.ts
@@ -411,7 +411,7 @@ export const isParallelReservation = (manifestations: Manifestation[]) =>
   hasCorrectAccessType(AccessTypeCode.Physical, manifestations) &&
   !isArticle(manifestations);
 
-// Because we  need to exclude the branches that are blacklisted, we need to
+// Because we need to exclude the branches that are blacklisted, we need to
 // use a custom hook to prevent duplicate code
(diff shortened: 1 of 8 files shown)


src/apps/material/helper.ts
```diff
@@ -411,19 +411,29 @@ export const isParallelReservation = (manifestations: Manifestation[]) =>
   hasCorrectAccessType(AccessTypeCode.Physical, manifestations) &&
   !isArticle(manifestations);
 
-// Because we  need to exclude the branches that are blacklisted, we need to use a custom hook to prevent duplicate code
+// Because we  need to exclude the branches that are blacklisted, we need to
```
Dresse, line 414: Very very small nitty picky thing. Can we remove the extra space between "we" and "need"? 😄 
```suggestion
// Because we need to exclude the branches that are blacklisted, we need to
```

Dresse (approved):
LGTM - Added small comment, but otherwise approved! 👍

PR #1097, opened 2024-04-09
Ddfform 555 opdateringer til orval

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFFORM-555

#### Description

This pull request updates the `FBS`, `Cover` and `Publizon` Orval clients, addressing the gaps after updating Orval to its latest version.

A command has been added to remove existing files in the models for auto-generated code. This ensures that we don't retain hooks associated with removed endpoints.

The only code changes in our React apps relate to `invalidateQueries`. In this PR, these changes are specifically for `getGetHoldingsV3QueryKey`, where I've added `faustIds` as a parameter. This update allows for fetching new holdings after a reservation is made.

Additionally, we have uppercased our fetcher methods to align with the case expectations of the requests.

#### Screenshot of the result


src/core/fbs/fbs-adapter.yaml
kasperg, line 1: This file contains a lot on whitespace changes that are not relevant to this PR.

I understand that this is because your editor has automatically reformatted the file. In the future if you are to add such changes then please keep them in a separate commit.

Currently the important change about changing `securitySchemes` to `securityDefinitions` is very hard to spot.

kasperg (approved):
:+1: Nice.

I general consider splitting your changes into even more commits. As I see it some of the bullets in the three current commits could be a commit of their own.

This would also allow you more room to explain why a change is needed e.g. why do we need to update the url to the cover service, why do we think it is OK to update the security schemes for FBS.

PR #1098, opened 2024-04-09
Use the proper svg for the list icon

spaceo:
#### Link to issue

https://reload.atlassian.net/browse/DDFLSBP-515

#### Description

Since Adam and I stepped on each others toes in previous tasks we decided to go with some of Adams changes and missed that the icon was not changed properly in dpl-react.
This PR fixes the problem.

#### Screenshot of the result
listicon-ezgif com-video-to-gif-converter


Dresse (approved):
LGTM

PR #1099, opened 2024-04-09
Renewal reason should fit in

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFLSBP-405

#### Description
This PR follows up on [changes made in the design system](https://github.com/danskernesdigitalebibliotek/dpl-design-system/pull/567) and adjusts the component and HTML structure to reflect these changes. 
The reasoning behind these changes is meant to ensure that if a loan can't be renewed, the reasoning message fits well in the item card without pushing content around. 
 
#### Screenshot of the result
-

-


src/apps/loan-list/utils/helpers.ts
```diff
@@ -83,4 +83,24 @@ export const getStackedItems = (
   return returnLoans;
 };
 
+export const getLoanDeliveryDate = (
+  loanType: LoanType,
+  formatDate: (date: string) => string,
```
spaceo, line 88: I really bugs me that we are passing date strings around when we should be using Date objects. But that is a matter that does not need to be solved in this PR

src/apps/loan-list/materials/selectable-material/selectable-material.tsx
```diff
@@ -111,44 +113,37 @@ const SelectableMaterial: FC<SelectableMaterialProps & MaterialProps> = ({
               {materialType}
             </div>
           </div>
+          {statusBadgeComponentMobile || ""}
```
spaceo, line 116: Not a biggie but I think it is more standard to emit `null`if you want nothing rendered in React

src/apps/loan-list/materials/utils/status-badge.tsx
```diff
@@ -26,29 +29,61 @@ const StatusBadge: FC<StatusBadgeProps> = ({
...
   if (daysBetweenTodayAndDue < threshold.danger && dangerText) {
     return (
-      <div className="status-label status-label--danger">{dangerText}</div>
+      <div
+        className={clsx("status-label status-label--danger", [
+          { "hide-on-desktop ml-16": mobileVersion }
+        ])}
+      >
+        {dangerText}
+      </div>
```
spaceo, line 38: Since this block is repeated multiple times with `status-label--*` as the only part changing I think you should either make a function or a component (I think I perefer a component, eg. StatusBadgeText) for that part of the rendering:
```typescript
<div
        className={clsx(`status-label status-label--${type}`, [
          { "hide-on-desktop ml-16": mobileVersion }
        ])}
      >
        {text}
      </div>
```

spaceo (changes requested):
Looks good :) I have a suggestion

src/apps/loan-list/materials/selectable-material/selectable-material.tsx
Adamik10: Great comment, thx!

src/apps/loan-list/materials/utils/status-badge-content.tsx
```diff
@@ -0,0 +1,29 @@
+import clsx from "clsx";
+import React, { FC } from "react";
+
+interface StatusBadgeContentProps {
+  statusLabelType: "danger" | "warning" | "neutral" | "info";
+  text: string;
+  mobileVersion?: boolean;
```
spaceo, line 7: Great with a separate component!
I think we decided on having a naming convention around booleans?
So in that case it should be: `isMobileVersion `?

spaceo (changes requested):
Great improvements!
I have a few suggestionz...

src/apps/loan-list/materials/utils/status-badge-content.tsx
```diff
@@ -0,0 +1,29 @@
+import clsx from "clsx";
+import React, { FC } from "react";
+
+interface StatusBadgeContentProps {
+  statusLabelType: "danger" | "warning" | "neutral" | "info";
```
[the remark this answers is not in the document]
Adamik10: Nopes

spaceo (approved):
LGTM 👍

PR #1100, opened 2024-04-09
Show a loading reserve button on material page until all data is loaded

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFLSBP-519

#### Description
Show a loading reserve button on material page until all data is loaded. We were waiting for patron data to load, but not for material availability data, which was being loaded inside a custom React hook. React query caching handling is used in our advantage here in order to extract whether the loading state inside the hook.

#### Screenshot of the result
-

-


Dresse (approved):
LGTM!

PR #1122, opened 2024-04-16
More dependabot groups

kasperg:
#### Description

Packages within the Testing library and Fullcalendar projects may be updated about the
same time.

Group them to make them move in parallel.


spaceo (approved):
LGTM 👍

PR #1135, opened 2024-04-18
Improved cover url resolving

spaceo:
#### Link to issue

https://reload.atlassian.net/browse/DDFLSBP-536

#### Description

This PR makes sure that we have a higher probability for getting a cover url in various places in our applications.
* Several pids are being sent at the same time to the cover service
* If we can find the best representation cover we'll use that otherwise we try to pick the first available cover

### DIFF
--- a/src/components/cover/helper.ts
+++ b/src/components/cover/helper.ts
@@ -31,8 +31,8 @@ export const getCoverUrl = ({
 
   const bestRepresentationCover = first(
     coverData.filter(
-      (item: CoverType) =>
-        bestRepresentation && item.id === bestRepresentation.pid
+      (cover: CoverType) =>
+        bestRepresentation && cover.id === bestRepresentation.pid
     )
   );
 
--- a/src/components/cover/helper.ts
+++ b/src/components/cover/helper.ts
@@ -3,7 +3,7 @@ import {
   CoverImageUrls,
   Cover as CoverType
 } from "../../core/cover-service-api/model";
-import { CoverProps } from "./cover";
+import { Manifestation } from "../../core/utils/types/entities";
 
 type CoverServiceSizes = keyof CoverImageUrls;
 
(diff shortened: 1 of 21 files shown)


src/core/dbc-gateway/generated/graphql.tsx
```diff
@@ -76,6 +76,8 @@ export enum AccessUrlType {
 
 export type Audience = {
   __typename?: "Audience";
+  /** PEGI age rating for games  */
+  PEGI?: Maybe<Pegi>;
```
Adamik10, line 80: Is this just an updated version of the API, or do we now get PEGI for a specific reason?

src/components/cover/helper.ts
```diff
@@ -0,0 +1,187 @@
...
+}: {
+  coverData: CoverType[] | null | undefined;
+  bestRepresentation?: Manifestation;
+  size: CoverServiceSizes;
+}) => {
+  if (!coverData) {
+    return null;
+  }
+
+  // Make sure we only have covers in our data that has an url in the given size.
```
Adamik10, line 43: Small typo: has an url -> has a url

src/components/cover/helper.ts
```diff
@@ -0,0 +1,187 @@
...
+  if (bestRepresentationCover && getUrl(bestRepresentationCover, size)) {
+    return getUrl(bestRepresentationCover, size);
+  }
+
+  // If the best representation method failed we try the first cover.
+  if (firstCover && getUrl(firstCover, size)) {
+    return getUrl(firstCover, size);
+  }
+
+  // Everything else failed. We don't know what to do 🤷.
```
Adamik10, line 71: Hahahaha love this.

Adamik10 (approved):
Very well done! I have a small question in regards to some updated generated files, and a tiny typo, but otherwise it's cleared from here :)

src/core/dbc-gateway/generated/graphql.tsx
spaceo: Yes, it is whatever is coming from the FBI graphql schema. I change a query and needed to regenerate the code.

src/components/cover/helper.ts
```diff
@@ -0,0 +1,187 @@
...
+  return null;
+};
+
+export default {};
+
+// ************** VITEST ***************
```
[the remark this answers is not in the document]
spaceo: 🙂

PR #1137, opened 2024-04-19
Searching for space

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFLSBP-529

#### Description
This PR introduces handling of search results with the query not defined, or under 3 characters in length.

#### Screenshot of the result
-

[Sibling PR in the CMS](https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/1017)

### DIFF
--- a/src/apps/search-result/helper.ts
+++ b/src/apps/search-result/helper.ts
@@ -18,4 +18,13 @@ export const getFirstMaterialTypeFromFilters = (
     : undefined;
 };
 
+export const shouldShowInvalidSearchPage = (
+  q: string,
+  qUrlParameter: string | null
+) => {
+  return (
+    (!qUrlParameter && !q) || (!qUrlParameter && q === "null") || q.length < 3
+  );
+};
+
 export default {};
--- a/src/apps/search-result/helper.ts
+++ b/src/apps/search-result/helper.ts
@@ -18,13 +18,4 @@ export const getFirstMaterialTypeFromFilters = (
     : undefined;
 };
 
-export const shouldShowInvalidSearchPage = (
-  q: string,
-  qUrlParameter: string | null
-) => {
-  return (
-    (!qUrlParameter && !q) || (!qUrlParameter && q === "null") || q.length < 3
-  );
-};
-
 export default {};
--- a/src/apps/search-result/search-result.tsx
+++ b/src/apps/search-result/search-result.tsx
@@ -30,7 +30,6 @@ import useFilterHandler from "./useFilterHandler";
 import SearchResultSkeleton from "./search-result-skeleton";
 import SearchResultZeroHits from "./search-result-zero-hits";
 import SearchResultInvalidSearch from "./search-result-not-valid-search";
-import { shouldShowInvalidSearchPage } from "./helper";
 
 interface SearchResultProps {
   q: string;
@@ -38,8 +37,6 @@ interface SearchResultProps {
 }
 
 const SearchResult: React.FC<SearchResultProps> = ({ q, pageSize }) => {
-  const queryParametersFromUrl = new URLSearchParams(window.location.search);
-  const qUrlParameter = queryParametersFromUrl.get("q");
   const { filters, clearFilter, addFilterFromUrlParamListener } =
     useFilterHandler();
   const cleanBranches = useGetCleanBranches();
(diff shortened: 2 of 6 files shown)


src/apps/search-result/search-result.tsx
```diff
@@ -29,13 +29,17 @@ import useGetCleanBranches from "../../core/utils/branches";
...
+import { shouldShowInvalidSearchPage } from "./helper";
 
 interface SearchResultProps {
   q: string;
   pageSize: number;
 }
 
 const SearchResult: React.FC<SearchResultProps> = ({ q, pageSize }) => {
+  const queryParametersFromUrl = new URLSearchParams(window.location.search);
+  const qUrlParameter = queryParametersFromUrl.get("q");
```
spaceo, line 42: This is not necessary.
The `q` parameter that is being passed to this component comes from `SearchResultEntry` that uses `getParams` which is also written in the commentary [here](https://github.com/danskernesdigitalebibliotek/dpl-react/blob/373a04708072387274812a8bbbdda9a4a6521851/src/apps/search-result/search-result.entry.tsx#L67-L69).

src/apps/search-result/helper.ts
```diff
@@ -18,4 +18,13 @@ export const getFirstMaterialTypeFromFilters = (
     : undefined;
 };
 
+export const shouldShowInvalidSearchPage = (
+  q: string,
+  qUrlParameter: string | null
+) => {
+  return (
+    (!qUrlParameter && !q) || (!qUrlParameter && q === "null") || q.length < 3
```
spaceo, line 26: I think it is a flaw that `q` can be "null" it should not be possible.
These changes to `getParams` should fix that:
```typescript
export const getParams = (props: Record<string, string | undefined>) =>
  Object.entries(props).reduce<Record<string, string>>(
    (acc, [property, value]) => {
      const paramValue = value || getUrlQueryParam(property);
      return {
        ...acc,
        [property]: paramValue ? String(paramValue) : ""
      };
    },
    {}
  );

spaceo: Also it should not be necessary to check for `q.length` if we add this to the search query:
```typescript
  const { data, isLoading } = useSearchWithPaginationQuery(
    {
      q: { all: q },
      offset: page * pageSize,
      limit: pageSize,
      filters: createFilters(filters, cleanBranches)
    },
    {
      enabled: !isEmpty(q) && q.length > 2
    }
  );

spaceo: ... so I think it is nice to have a helper function called: `shouldShowInvalidSearchPage`. But the logic of it should be revised after my proposals :)

spaceo (changes requested):
I have some change requests ;)

src/apps/search-result/search-result.tsx
Adamik10: That's good and well, but we still need to somehow figure out whether there is anything in the search bar or not in order to determine whether the user is searching for the string "null" or whether there is no q parameter there at all (which it should always be present when on the search result page - bc the autosuggest and search bar redirect to `/search?q={searchTerm}`

src/apps/search-result/helper.ts
Adamik10: Very good suggestion with the revision of `getParams` helper, I added it in. But I don't necessarily agree with your second point, and I think that it's a matter of preference. 

If we don't ever make a call and we already are on the search result page, then we're never gonna get any data, which currently just returns `null` instead of the whole component. We could of course then return our new `< SearchResultInvalidSearch  />` component, but are we sure that that's always the correct case there? Asking a dedicated function whether to render the `< SearchResultInvalidSearch  />` component based on all the criteria that we need to know about in order to make the right call seems like the better options to me.

Adamik10: It didn't really make sense to keep it - the statement is tiny now :)

Great suggestion with refactoring the `getParams ` helper

src/apps/search-result/search-result.tsx
Adamik10: yeah it makes sense together with your other comment :) I made it happen

spaceo (approved):
We could put the min length conf in our core configuration. 
But these changes makes me overwhelmingly happy.

PR #1143, opened 2024-04-23
Refactored removeModalIdFromUrl to use common URL functions

clausbruun:
#### Description

Refactored removeModalIdFromUrl  to use common URL functions for consistency. Also, see the Zulip thread [here](https://reload.zulipchat.com/#narrow/stream/413259-DDF---brahma/topic/Brug.20af.20f.C3.A6lles.20url.20funktioner/near/434798325).


Adamik10 (approved):
👏

PR #1149, opened 2024-04-24
Fix/ever loading availability labels on search page

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFLSBP-571

#### Description
This PR moves "articles are always available" logic into useAvailabilityData hook. Previously it was in the availability label component, but it makes more sense for it to be located together with the rest of the logic that decides whether a material is available or not. This also fixes an issue we were having with availability labels for articles loading although they were just supposed to be available and not load any data.

#### Screenshot of the result
-
-

### DIFF
--- a/src/components/availability-label/helper.ts
+++ b/src/components/availability-label/helper.ts
@@ -12,17 +12,20 @@ import {
 import { FaustId } from "../../core/utils/types/ids";
 import useGetAvailability from "../../core/utils/useGetAvailability";
 import { publizonProductStatuses } from "./types";
+import { ManifestationMaterialType } from "../../core/utils/types/material-type";
 
 export const useAvailabilityData = ({
   accessTypes,
   access,
   faustIds,
-  isbn
+  isbn,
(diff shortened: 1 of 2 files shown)


src/components/availability-label/helper.ts
```diff
@@ -37,13 +40,23 @@ export const useAvailabilityData = ({
     }
   }, [isOnline]);
 
+  useEffect(() => {
+    // Articles are always available.
```
spaceo, line 44: I think we should also have a check if `available` has been set or not.
So in the beginning of `useEffect` we would have a check:
```typescript
if (isAvailable !== null) {
  return;
}
```

spaceo: And then we need `isAvailable` as a dependency for the hook of course...

spaceo (changes requested):
Looks good. I have a performance proposal.

spaceo (approved):
LGTM 👍

PR #1150, opened 2024-04-25
User registration form v2

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFLSBP-563
https://reload.atlassian.net/browse/DDFLSBP-594

#### Description
User registration page V2 according to the new design brief & styling in the sibling PR.

#### Screenshot of the result
-
[Sibling PR design system](https://github.com/danskernesdigitalebibliotek/dpl-design-system/pull/616)
[Sibling PR CMS](https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/1080)


src/apps/create-patron-user-info/CreatePatron.tsx
```diff
@@ -4,31 +4,38 @@ import { useText } from "../../core/utils/text";
...
     const {
       attributes: { cpr: userCpr }
     } = userInfo;
-
     // Otherwise set the cpr so we can show the create patron form.
     setCpr(String(userCpr));
-  }, [agencyId, isLoading, userInfo]);
+  }, [agencyId, isLoading, userInfo, storybookContextCpr]);
+
+  if (storybookContextCpr) {
```
spaceo, line 36: I think we should be very careful with mxing tools with application code. Sometimes its unavoidable, but its should always be possible to remove a tool and the remaining application code would still be running. This goes for eg. Storybook and testing.
So it seems like a small thing, but I think we can keep the prop on `<CreatePatron />` but instead call it `cpr`, make it optional and pass it to the `<UserInfo />` in line 37

src/apps/create-patron-user-info/UserInfo.tsx
```diff
@@ -35,27 +35,44 @@ const UserInfo: FC<UserInfoProps> = ({ cpr }) => {
...
+  const [isSubmitError, setIsSubmitError] = useState<boolean>(false);
+  const getSubmitButtonText = () => {
+    if (isLoading) {
+      return t("createPatronButtonLoadingText");
+    }
+    if (isSubmitError) {
+      return t("createPatronButtonErrorText");
+    }
+    return t("createPatronConfirmButtonText");
+  };
```
spaceo, line 48: Could we move this to a helper.ts file? I have another PR on its way where I have introduced a helper so it could fit in there.

src/apps/create-patron-user-info/UserInfo.tsx
```diff
@@ -35,27 +35,44 @@ const UserInfo: FC<UserInfoProps> = ({ cpr }) => {
...
         },
         {
           onSuccess: () => {
+            setIsLoading(false);
             redirectTo(redirectOnUserCreatedUrl);
+          },
+          onError: () => {
+            setIsLoading(false);
+            setIsSubmitError(true);
           }
```
spaceo, line 76: I remember I created a whole system for a situation like this, where you have a submit button and then you get a feedback modal that can either fail or display a success message. I don't know if it would make sense to use it here. I also have PR coming with changes to this bit that forces the user to be logged out and log in again.

spaceo (changes requested):
Looks good 👏 
I have some comments though.

src/apps/create-patron-user-info/CreatePatron.tsx
Adamik10: We need to call it something different than CPR because we already have a variable called that. And it indeed is optional.

src/apps/create-patron-user-info/UserInfo.tsx
Adamik10: In here we're not in a modal context.

src/apps/create-patron-user-info/CreatePatron.dev.tsx
```diff
@@ -134,7 +134,7 @@ export default {
         "The phone number must be 6 to 15 characters in length and should be comprised solely of numbers or begin with a +",
       control: { type: "text" }
     },
-    storybookContextCpr: {
+    fakeCpr: {
```
spaceo, line 137: Sorry, but my idea was just to call it: `cpr` :) I can accept `fakeCpr` but I think `cpr` makes more sense.

src/apps/create-patron-user-info/helper.tsx
```diff
@@ -0,0 +1,17 @@
+import { UseTextFunction } from "../../core/utils/text";
```
spaceo, line 1: Our convention is to call helper files that do not have jsx rendering: `helper.ts`

spaceo (changes requested):
Almost there. Most important is the naming of the create-patron-user-info/helper.tsx file

src/apps/create-patron-user-info/CreatePatron.dev.tsx
Adamik10: It can't really be called that, as I explained previously there already is a variable called `cpr`.

spaceo (commented):
Please look at my suggestion.

spaceo (approved):
LGTM 👍

PR #1165, opened 2024-05-02
Don't show digital loans on the dashboard

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFLSBP-584

#### Description
Don't show digital loans in loan modals on dashboard:
- Returned too late modal
- To be returned soon modal
- Longer return time modal

#### Screenshot of the result
-
-


clausbruun (approved):
Very nice! 🎉

PR #1166, opened 2024-05-02
Registration redirect login message

spaceo:
#### Link to issue

https://reload.atlassian.net/browse/DDFLSBP-615

#### Description

Adds a story showing the redirect login message after submission of the patron registration form.

#### Screenshots
image

 Belongs to a trinity together with:
+ https://github.com/danskernesdigitalebibliotek/dpl-design-system/pull/619
+ https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/1091

### DIFF
--- a/src/apps/create-patron-user-info/CreatePatron.dev.tsx
+++ b/src/apps/create-patron-user-info/CreatePatron.dev.tsx
@@ -14,6 +14,16 @@ export default {
     ...pincodeArgs,
     ...globalTextArgs,
     ...globalConfigArgs,
+    authUrl: {
+      name: "Url where user can authenticate",
+      defaultValue: "/login",
+      control: { type: "text" }
+    },
+    logoutUrl: {
+      name: "Url where user can logout",
+      defaultValue: "/logout",
+      control: { type: "text" }
+    },
     blacklistedPickupBranchesConfig: {
       defaultValue: "FBS-751032,FBS-751031,FBS-751009,FBS-751027,FBS-751024",
       control: { type: "text" }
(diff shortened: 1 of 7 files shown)


src/apps/create-patron-user-info/CreatePatron.dev.tsx
```diff
@@ -14,6 +14,16 @@ export default {
     ...pincodeArgs,
     ...globalTextArgs,
     ...globalConfigArgs,
+    authUrl: {
```
Dresse, line 17: Maybe we should consider calling this `loginUrl`?

src/apps/create-patron-user-info/RedirectToLoginMessage.tsx
```diff
@@ -0,0 +1,55 @@
...
+import { redirectTo } from "../../core/utils/helpers/url";
+import { useText } from "../../core/utils/text";
+import { getRedirectUrl } from "./helper";
+import { Button } from "../../components/Buttons/Button";
+
+const RedirectToLoginMessage: FC = () => {
+  const t = useText();
+  const u = useUrls();
+  const redirectUrl = getRedirectUrl({
+    loginUrl: u("authUrl"),
```
Dresse, line 12: Same as above comment.

Dresse (approved):
LGTM, but I have a suggestion 👍

src/apps/create-patron-user-info/CreatePatron.dev.tsx
spaceo: It has been named like this since the dawn of time. Would require a refactor in the cms as well. Lets keep that for a rainy day ;) But good point!

Dresse: Fair enough 😄

spaceo:
Moved to separate release:
https://github.com/danskernesdigitalebibliotek/dpl-react/pull/1187

PR #1170, opened 2024-05-03
Remove reserve and find on shelf from articles

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFLSBP-622

#### Description
There are many types of articles, and ALL of them need to always show the "see online" button on the material page and treat them as online materials. This PR adjusts the isArticle() helper function so that it searches for any material types containing the word "article" (in danish).

#### Screenshot of the result
image

-


clausbruun (approved):
Perfect  🎉

PR #1171, opened 2024-05-03
Set an empty alt tag on the ButtonTag close button.

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFLSBP-582

#### Description
Because screen readers announce it as a toggle button, which is according to https://www.w3.org/Translations/WCAG21-da/#info-and-relationships

#### Screenshot of the result


clausbruun (approved):
Fixed! 🔨  🎉

PR #1172, opened 2024-05-03
Make it possible for availability labels to be non-interactive elements

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFLSBP-581

#### Description
Make it possible for availability labels to be non-interactive elements, and use them like this on search result page for search result items that only have one material type + on material page in the top for materials that only have one material type + on material page in the "editions" disclosure for each item, as these availability labels don't serve an interactive function, but merely are visual labels.

#### Screenshot of the result
-

-


clausbruun (approved):
There's a small merge conflict, but otherwise, bravo! 🤩

PR #1183, opened 2024-05-06
Add focus to the last added facet in FacetLineSelected [DDFLSBP-590]

clausbruun:
#### Link to issue

https://reload.atlassian.net/browse/DDFLSBP-590

#### Description

Implemented focus management in ButtonTag and auto-focus on new facets in FacetLineSelected for enhanced accessibility

#### Screenshot of the result

https://github.com/danskernesdigitalebibliotek/dpl-react/assets/105956/b15d836e-5cac-443e-b8ee-6387b8b1959f

### DIFF
--- a/src/components/Buttons/ButtonTag.tsx
+++ b/src/components/Buttons/ButtonTag.tsx
@@ -11,35 +11,31 @@ type ButtonTagProps = {
   dataCy?: string;
 };
 
-const ButtonTag: React.FunctionComponent<ButtonTagProps> = ({
-  onClick,
-  selected,
-  children,
-  size,
-  removable = false,
-  dataCy
-}) => {
-  const className = clsx(
-    "tag",
-    selected && "tag--fill",
-    size && `tag--${size}`,
-    "cursor-pointer"
-  );
+const ButtonTag = React.forwardRef<HTMLButtonElement, ButtonTagProps>(
+  ({ onClick, selected, children, size, removable = false, dataCy }, ref) => {
+    const className = clsx(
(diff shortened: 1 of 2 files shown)


src/components/Buttons/ButtonTag.tsx
```diff
@@ -11,35 +11,31 @@ type ButtonTagProps = {
...
-  dataCy
-}) => {
-  const className = clsx(
-    "tag",
-    selected && "tag--fill",
-    size && `tag--${size}`,
-    "cursor-pointer"
-  );
+const ButtonTag = React.forwardRef<HTMLButtonElement, ButtonTagProps>(
+  ({ onClick, selected, children, size, removable = false, dataCy }, ref) => {
```
Adamik10, line 15: Is there a reason why we don't destructure the ref inside the curly brackets? 🤔

Adamik10 (approved):
Very nice 👏 just got one question :)

src/components/Buttons/ButtonTag.tsx
clausbruun: As I understand it, `ref` is not a prop in the typical sense, but a way to interact directly with a DOM element, as described here: https://react.dev/reference/react/forwardRef

Maybe something like this would be more clear? 🙂

```

const ButtonTag = React.forwardRef<HTMLButtonElement, ButtonTagProps>(
  (props, ref) => {
    const {
      onClick,
      selected,
      children,
      size,
      removable = false,
      dataCy
    } = props;
```

PR #1185, opened 2024-05-06
Fix/selectable material inner button accessibility

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFLSBP-610

#### Description
Add button role to selectable material content that served as a button.

#### Screenshot of the result
-

-


clausbruun (approved):
Refined! 🎉

PR #1186, opened 2024-05-07
Fix/failing work page

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFLSBP-617

#### Description
This PR fixes a specific case for the work with work ID of `work-of:150086-netmusik:BIS-2067`. It was happening because we didn't know that a FaustId can also contain dashes apart from numbers and letters.

#### Screenshot of the result
-

-


clausbruun (approved):
Dashing 🎉

PR #1189, opened 2024-05-10
Announce search result updates

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFLSBP-591

#### Description
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-live

#### Screenshot of the result
-

-


spaceo (approved):
LGTM 
You need to fix this in the cms as well :)

PR #1200, opened 2024-05-13
Fix incorrect pincode message

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFLSBP-633

#### Description
Don't let users submit user profile changes if password isn't valid.

#### Screenshot of the result
-

-


spaceo (approved):
LGTM 👍

spaceo:
I know that you have two UX-camps:

One that thinks its bad practise to disable buttons as a part of validation, and errors should be reflected either on the fields or as a global form error message.

One that think its ok to disable the submit button like you do here :)

PR #1204, opened 2024-05-15
Enhance `MaterialGrid` functionality to focus on next item after "show more" click

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFFORM-751

#### Description
This pull request ensures that the next item in the `MaterialGrid` receives focus after the "show more" button is clicked.

#### Screenshot of the result

https://github.com/danskernesdigitalebibliotek/dpl-react/assets/49920322/eaaf153c-a6f9-4303-a2d6-4647b0bcc7a4

### DIFF
--- a/src/apps/material-grid/MaterialGrid.tsx
+++ b/src/apps/material-grid/MaterialGrid.tsx
@@ -1,5 +1,5 @@
 import * as React from "react";
-import { useState } from "react";
+import { useState, useEffect } from "react";
 import { WorkId } from "../../core/utils/types/ids";
 import RecommendedMaterial from "../recommended-material/RecommendedMaterial";
 import { useText } from "../../core/utils/text";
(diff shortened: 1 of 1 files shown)


LasseStaus (changes requested):
This could be fine maybe but i am abit confused about this: 

The MaterialListItem you are adding, sets a tabIndex of -1 on the Li-element, indicating that it the Li-element should not be tab-able. But when you set the new focus, you are explicitly setting this to the Li element. 

I guess optimally it should focus the next tab-able element in the fifth item, but maybe i am wrong? 

Not sure what the best approach to doing that is. Maybe a utility function finding the next tab-able element? 

If you think this is fine and shouldn't be changed tho let me know.  🚀

kasperbirch1:
@LasseStaus I understand your point. I prefer this approach because it ensures consistency across the project, aligning with the implementation in both `FavoritesList.tsx` and `SearchResultList.tsx`

If you still have any concerns about this approach, we can always discuss it with the team? ☺️

LasseStaus (approved):
@kasperbirch1 

That is fine if you prefer this 👍  

Just wanted to highlight it since it seemed a bit strange to me. 

I think it is a good idea to streamline it so it is the same everywhere, but maybe it wasn't properly constructed when it was first implemented. - I'm not saying it isn't, just that that should be considered. 

Will approve and let you decide 🚀  👍

PR #1208, opened 2024-05-18
Add workflow that creates a pr externally in the dpl-cms repo

spaceo:
#### Link to issue

https://reload.atlassian.net/browse/DDFLSBP-643

#### Description

This PR adds a Github Workflow `create-cms-pr` that requests [dpl-cms to create a PR](https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/1158).
For separation of concerns the workflow is triggered by the `create-release-on-branch-changes` workflow. In that way we try to avoid leaving all steps in the `create-release-on-branch-changes` and isolate the remote pr action steps in a separate yml file.

The changes here are similar to the changes in the [dpl-design-ssystem repo](https://github.com/danskernesdigitalebibliotek/dpl-design-system/pull/632).

We need to set secrret and variables similiar to what has been set in:
https://github.com/reload/poc-dpl-react-pr-trigger

### DIFF
--- a/.github/workflows/create-cms-pr.yml
+++ b/.github/workflows/create-cms-pr.yml
@@ -1,0 +1,65 @@
+name: Create CMS PR
+on:
+  workflow_run:
+    workflows: ["Create release on branch changes"]
+    types:
+      - completed
+
+jobs:
+  invoke-external-pr-workflow:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/download-artifact@v4
+        with:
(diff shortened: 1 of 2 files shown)


spaceo:
I have made changes according to some of your requests.
Please note that the: "Create release on branch changes / Create release on branch changes (push)" is failing because the workflow is not merged into develop.

kasperg (approved):
:+1: - same comments as for the design system.

PR #1219, opened 2024-05-20
Integrate `handleDateSet` into `useOpeningHoursEditor` hook

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFFORM-787

#### Description
This pull request enhances the fetching mechanism for opening hours in the admin area, ensuring that we only retrieve the necessary data for the current view.

The `datesSet` method from FullCalendar, which provides the currently displayed dates and updates with navigation button clicks, is now utilized. This ensures we fetch only the relevant opening hours for the current view.

<s>Initially, I attempted to use the `enabled` property in the options object of `useDplOpeningHoursListGET` but encountered the following error: "Object literal may only specify known properties, and 'enabled' does not exist in type '{ query?: UseQueryOptions<DplOpeningHoursListGET200Item[] | null, void, DplOpeningHoursListGET200Item[] | null, QueryKey> | undefined; }'."
I suspect this issue is related to the operations in dplCms within orval.config.ts.

As a workaround, I implemented a ternary operator in `useDplOpeningHoursListGET` that conditionally sets `from_date` and `to_date` only when `datesSet is available.</s>

Integrated `dpl_opening_hours_list:GET? in `orval.config`. This enhancement enables the use of the enabled option in `useDplOpeningHoursListGET`, allowing the fetching of opening hours only when `datesSet` is not `null`.

Rearranged headerToolbar layout in FullCalendar. The title (week/month) is now centralized for consistent positioning of the previous and next buttons, regardless of the title length. This addresses the issue where navigating through the month was difficult when the buttons changed positions.

#### Local test from samsø 
https://github.com/danskernesdigitalebibliotek/dpl-react/assets/49920322/ad8ae9af-9783-446e-be5a-d47f6fd104e7

### DIFF
--- a/src/apps/opening-hours-editor/OpeningHoursEditor.tsx
+++ b/src/apps/opening-hours-editor/OpeningHoursEditor.tsx
@@ -36,8 +36,13 @@ const OpeningHoursEditor: React.FC<OpeningHoursEditorType> = ({
   const fullCalendarRef = React.useRef<FullCalendar>(null);
   const fullCalendarApi = fullCalendarRef.current?.getApi();
 
-  const { events, handleEventAdd, handleEventEditing, handleEventRemove } =
-    useOpeningHoursEditor();
+  const {
+    events,
+    handleEventAdd,
+    handleEventEditing,
+    handleEventRemove,
+    handleDatesSet
+  } = useOpeningHoursEditor();
 
   const { dialogContent, openDialogWithContent, closeDialog, dialogRef } =
     useDialog({
@@ -97,6 +102,7 @@ const OpeningHoursEditor: React.FC<OpeningHoursEditorType> = ({
         height="auto"
         selectMirror
         allDaySlot={false}
+        datesSet={handleDatesSet}
       />
     </>
   );
--- a/src/apps/opening-hours-editor/OpeningHoursEditor.tsx
+++ b/src/apps/opening-hours-editor/OpeningHoursEditor.tsx
@@ -62,9 +62,9 @@ const OpeningHoursEditor: React.FC<OpeningHoursEditorType> = ({
         ref={fullCalendarRef}
         plugins={[dayGridPlugin, timeGridPlugin, interactionPlugin]}
         headerToolbar={{
-          left: "title",
-          center: "prev,next today",
-          right: "dayGridMonth,timeGridWeek"
+          left: "dayGridMonth,timeGridWeek",
+          center: "title",
+          right: "prev,next today"
         }}
         initialView="timeGridWeek"
         locale={da}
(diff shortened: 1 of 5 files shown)


src/apps/opening-hours-editor/useOpeningHoursEditor.tsx
```diff
@@ -21,9 +21,16 @@ const useOpeningHoursEditor = () => {
...
   const queryClient = useQueryClient();
   const { data: openingHoursData } = useDplOpeningHoursListGET({
-    branch_id: openingHoursBranchId
+    branch_id: openingHoursBranchId,
+    ...(datesSet
+      ? {
+          from_date: getStringForDateInput(datesSet.start),
+          to_date: getStringForDateInput(datesSet.end)
+        }
+      : {})
```
kasperg, line 33: Consinder updating the Orval client to add an enabled property.

I appreciate your comment in the PR. You can see how we have added this for other requests in [`orval.config.js`](https://github.com/danskernesdigitalebibliotek/dpl-react/blob/1a9f9df85eb133c1ff7583cbe60b0cdefa45d9bb/orval.config.ts#L132-L132).

If fetching all dates is actually a problem we should consider whether we want to mark from and to dates as required. That does not prevent callers from specifying from "1900-01-01 " and to "2100-01-01". Then we will end up with the same problem.

src/apps/opening-hours-editor/OpeningHoursEditor.tsx
```diff
@@ -57,9 +62,9 @@ const OpeningHoursEditor: React.FC<OpeningHoursEditorType> = ({
         ref={fullCalendarRef}
         plugins={[dayGridPlugin, timeGridPlugin, interactionPlugin]}
         headerToolbar={{
-          left: "title",
-          center: "prev,next today",
-          right: "dayGridMonth,timeGridWeek"
+          left: "dayGridMonth,timeGridWeek",
+          center: "title",
+          right: "prev,next today"
```
kasperg, line 67: I wonder why you changed this.

I can see that you put in a separate commit but I do not understand the reasoning behind the change. It is not mentioned in the PR or in the ticket.

kasperg (changes requested):
:+1: nice and simple change. I still think it needs a little more thought.

src/apps/opening-hours-editor/OpeningHoursEditor.tsx
kasperbirch1: I found it annoying that the button changes position when I click through the months.

kasperg (approved):
:+1: LGTM

PR #1224, opened 2024-05-24
Fix/link from fees

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFLSBP-553

#### Description
If the user has overdue loans, a warning bar pops up on the fees page. This warning bar has a CTA button that was linking to the loans page. Now it opens the modal with overdue loans as the client asked for.

#### Screenshot of the result
-

-

### DIFF
--- a/src/apps/fee-list/FeeList.entry.tsx
+++ b/src/apps/fee-list/FeeList.entry.tsx
@@ -7,6 +7,11 @@ import FeeList from "./FeeList";
 import GlobalUrlEntryPropsInterface from "../../core/utils/types/global-url-props";
 import { withConfig } from "../../core/utils/config";
 import { GlobalEntryTextProps } from "../../core/storybook/globalTextArgs";
+import { pageSizeGlobal } from "../../core/utils/helpers/general";
+import { GroupModalProps } from "../../core/storybook/groupModalArgs";
+import { GroupModalLoansProps } from "../../core/storybook/loanGroupModalArgs";
+import { MaterialDetailsModalProps } from "../../core/storybook/materialDetailsModalArgs";
+import { RenewalArgs } from "../../core/storybook/renewalArgs";
 
 export interface IntermedateListEntryConfigProps {
   expirationWarningDaysBeforeConfig: string;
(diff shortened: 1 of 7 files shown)


src/core/utils/helpers/modal-helpers.ts
```diff
@@ -45,3 +45,9 @@ export const getDetailsModalId = (queryParam: string, prefix: string) => {
   }
   return "";
 };
+
+export const createLoanModalId = (
+  dueDate: string | null | undefined,
+  dueDateModal: string | number | Record<string, unknown>,
+  allLoansId: string | number | Record<string, unknown>
+) => (dueDate ? `${dueDateModal}-${dueDate}` : String(allLoansId));
```
spaceo, line 53: I think we should define modalId functions in the modal components.
That makes it more modular. If we want to delete the modal in the future the whole thing is self-conatined.
Plus that if there is some functionality involved in finding out the omdal id fragments, it probably comes from functionality inside of the component itself.

spaceo: For consistency I think you should use this function when creating the id:
https://github.com/danskernesdigitalebibliotek/dpl-react/blob/4f75d55e348f0395031fa4d8d76397ece159e504/src/core/utils/helpers/modal-helpers.ts#L3-L5

src/apps/fee-list/FeeList.entry.tsx
```diff
@@ -62,8 +67,21 @@ const FeeListEntry: FC<
...
+    RenewalArgs
+> = ({ pageSizeDesktop, pageSizeMobile }) => {
+  const pageSize = pageSizeGlobal(
+    {
+      desktop: pageSizeDesktop,
+      mobile: pageSizeMobile
+    },
+    "pageSizeLoanList"
+  );
+  return <FeeList pageSize={pageSize} />;
```
spaceo, line 83: Would be nice with a comment here.
You are writing `pageSizeLoanList ` although we are dealing with a fee list?

spaceo (changes requested):
Nice work. I have some comments.

src/core/utils/helpers/modal-helpers.ts
Adamik10: @spaceo but we've been creating this specific modal id differently. I just took the function out into a helper.

spaceo: I don't think you need a full overhaul. Just like it has been done eg. here:
https://github.com/danskernesdigitalebibliotek/dpl-react/blob/4f75d55e348f0395031fa4d8d76397ece159e504/src/components/find-on-shelf/FindOnShelfModal.tsx#L12-L14

spaceo (changes requested):
Comment back

src/core/utils/helpers/modal-helpers.ts
Adamik10: donezies :)

In regards to the self-containment of modal related stuff - I think it's pretty self contained when the function exists in a file called `modal-helpers.ts`

spaceo (approved):
LGTM 👍

PR #1235, opened 2024-05-27
Handle libraryProfile zero values correctly [DDFLSBP-639]

clausbruun:
#### Link to issue

https://reload.atlassian.net/browse/DDFLSBP-639

#### Description

This is a fix for another existing bug discovered during testing of DDFLSBP-639

#### Screenshot of the result

image


Adamik10 (approved):
👏

PR #1240, opened 2024-05-29
Fix advanced search row focusing - the sequel

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFLSBP-598

#### Description
Focus on the new last advanced search row when deleting last row

#### Screenshot of the result
-

-


clausbruun (changes requested):
I don't think this one is quite right yet, try this:
* Add one row
* Delete 2nd row

https://github.com/danskernesdigitalebibliotek/dpl-react/assets/105956/96d3a105-66bd-4b46-84e8-5e1d38afdf66

Adamik10:
Very keen eye @clausbruun. I was aware of this, but decided that I won't use more time on debugging it as I don't think it's a likely scenario, and only a small portion of the user base actually use advanced search. I will speak to DDF about it when I present it 😉

clausbruun (approved):
Good decision 🎉

PR #1245, opened 2024-05-30
DDFFORM-807 - Enable `initialDate` parameter in opening hours apps (For Cypress testing)

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFFORM-807
https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/1190

#### Description

This pull request enables both the `OpeningHours` and `OpeningHoursEditor` apps to set the `initialDate` via a URL parameter. This enhancement allows tests to begin from a specific date, streamlining the testing of the complete opening hours workflow from dpl-cms.

### DIFF
--- a/src/apps/opening-hours-editor/OpeningHoursEditor.entry.tsx
+++ b/src/apps/opening-hours-editor/OpeningHoursEditor.entry.tsx
@@ -5,6 +5,7 @@ import OpeningHoursEditor, {
   OpeningHoursEditorType
 } from "./OpeningHoursEditor";
 import { withConfig } from "../../core/utils/config";
+import { getInitialDateFromUrl } from "./helper";
 
 interface OpeningHoursEditorEntryTextProps {
   openingHoursRemoveEventButtonText: string;
@@ -36,8 +37,9 @@ const OpeningHoursEditorEntry: React.FC<
   OpeningHoursEditorEntryTextProps &
     OpeningHoursEditorType &
     OpeningHoursEditorEntryConfigProps
-> = ({ initialDate = new Date() }) => {
-  return <OpeningHoursEditor initialDate={initialDate} />;
+> = ({ initialDate }) => {
+  const initialDateParam = getInitialDateFromUrl();
+  return <OpeningHoursEditor initialDate={initialDate ?? initialDateParam} />;
 };
 
 export default withConfig(withUrls(withText(OpeningHoursEditorEntry)));
--- a/src/apps/opening-hours-editor/OpeningHoursEditor.entry.tsx
+++ b/src/apps/opening-hours-editor/OpeningHoursEditor.entry.tsx
@@ -39,7 +39,11 @@ const OpeningHoursEditorEntry: React.FC<
     OpeningHoursEditorEntryConfigProps
 > = ({ initialDate }) => {
   const initialDateParam = getInitialDateFromUrl();
-  return <OpeningHoursEditor initialDate={initialDate ?? initialDateParam} />;
+  return (
+    <OpeningHoursEditor
+      initialDate={initialDate ?? (initialDateParam || new Date())}
+    />
+  );
 };
 
 export default withConfig(withUrls(withText(OpeningHoursEditorEntry)));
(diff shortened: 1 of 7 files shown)


src/apps/opening-hours-editor/helper.ts
```diff
@@ -172,3 +172,10 @@ export const isOpeningHourWeeklyRepetition = (
     DplOpeningHoursListGET200ItemRepetitionType.weekly
   );
 };
+
+// This used for cypress testing in in cms where we test the complete flow of
+// opening hours between the editor (Admin) and the calendar (Frontend)
+export const getInitialDateFromUrl = (): string | null => {
+  const query = new URLSearchParams(window.location.search);
+  return query.get("initialDate");
+};
```
kasperg, line 181: Please consider your typing.

As I see it you have a function `getInitialDateFromUrl()` which returns a string. This causes changes all the way down call stack which has been designed to accomodate a Date.

As I see it a Date is the proper type to use for this. Instead of changing the type I suggest that you ensure that the return value from `getInitialDateFromUrl()` is in fact a Date.

If that is not the case you can return null (and log the problem to `console.debug()`) or throw an error.

kasperg (changes requested):
:+1: but I think your type handling needs another round.

src/apps/opening-hours-editor/OpeningHoursEditor.entry.tsx
```diff
@@ -36,8 +37,13 @@ const OpeningHoursEditorEntry: React.FC<
   OpeningHoursEditorEntryTextProps &
     OpeningHoursEditorType &
     OpeningHoursEditorEntryConfigProps
-> = ({ initialDate = new Date() }) => {
-  return <OpeningHoursEditor initialDate={initialDate} />;
+> = ({ initialDate }) => {
+  const initialDateParam = getInitialDateFromUrl();
+  return (
+    <OpeningHoursEditor
+      initialDate={initialDate ?? (initialDateParam || undefined)}
```
kasperbirch1, line 44: I’m considering whether `initialDate={initialDate ?? (initialDateParam || new Date())}` should be here. However, `OpeningHoursEditor` operates as expected without an `initialDate`, so I’d rather avoid passing a `new Date()`.

src/apps/opening-hours-editor/helper.ts
```diff
@@ -172,3 +173,26 @@ export const isOpeningHourWeeklyRepetition = (
...
+  const query = new URLSearchParams(window.location.search);
+  const initialDateString = query.get("initialDate");
+
+  if (!initialDateString) {
+    return null;
+  }
+
+  const date = new Date(initialDateString);
+  if (isDate(date)) {
+    return date;
```
kasperg, line 189: This might be me worrying too much. I expect that you have tested this but when I reason about it I wonder if it works.

According to [the Lodash docs](https://lodash.com/docs/4.17.15#isDate) `isDate()` checks if value is classified as a Date object.

[According to MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/Date#return_value) `new Date()` returns a new Date object but the return value of different methods will signal whether it is a date object.

[According to a Lodash issue `isDate()` will not return false if the value is not a date](https://github.com/lodash/lodash/issues/5040).

src/apps/opening-hours-editor/OpeningHoursEditor.entry.tsx
kasperg: I wonder why you want to avoid passing `new Date()`. As I see it this was what we did before if `initialDate` was not passed.

src/apps/opening-hours-editor/helper.ts
```diff
@@ -172,3 +173,26 @@ export const isOpeningHourWeeklyRepetition = (
...
+  }
+
+  // eslint-disable-next-line no-console
+  console.log(
```
kasperg, line 193: Consider using `console.debug` instead. This is was we have done elsewhere.

PR #1257, opened 2024-06-03
Forward exceptions to window.onerror  [DDFLSBP-569]

clausbruun:
#### Link to issue

https://reload.atlassian.net/browse/DDFLSBP-569

#### Description

Enhance ErrorBoundary to forward exceptions to window.onerror for logging in jsnlog.

#### Screenshot of the result

(Exception fremprovokeret for test)
image

#### Sibling PR
https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/1077

### DIFF
--- a/src/core/mount.js
+++ b/src/core/mount.js
@@ -6,6 +6,7 @@ import Store from "../components/store";
 import { persistor, store } from "./store";
 import ErrorBoundaryAlert from "../components/error-boundary-alert/ErrorBoundaryAlert";
 import { closeLastModal } from "./modal.slice";
+import extractErrorDetails from "./utils/extractErrorDetails";
 
 /**
  * We look for containers and corresponding applications.
@@ -37,6 +38,24 @@ function mount(context) {
                 // Logging should be acceptable in an error handler.
                 // eslint-disable-next-line no-console
                 console.error(error, info);
+
+                const { filename, lineNumber, column } = extractErrorDetails(
+                  info.componentStack
+                );
+
+                // Call window.onerror to send the error to the error logging system.
+                if (window.onerror) {
+                  window.onerror(
+                    `${error.name}: ${error.message}`,
+                    filename,
+                    lineNumber,
+                    column,
+                    {
+                      ...info,
+                      stack: info.componentStack
+                    }
+                  );
+                }
               }
             }),
             {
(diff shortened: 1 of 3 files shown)


src/core/utils/extractErrorDetails.js
```diff
@@ -0,0 +1,31 @@
+/**
+ * Extracts error details from a stack trace.
+ *
+ * @param {string} stack - The stack trace.
+ * @returns {Object} An object containing filename, lineNumber, and column.
+ */
```
spaceo, line 6: 👍 for jsdoc in js context

src/core/utils/extractErrorDetails.js
```diff
@@ -0,0 +1,31 @@
...
+const extractErrorDetails = (stack) => {
+  const stackLines = (stack || "").split("\n");
+  const match = stackLines
+    .map((line) => {
+      const parts = line.match(/(?:\()?(.*?):(\d+):(\d+)(?:\))?/);
+      if (parts) {
+        return {
+          filename: parts[1],
```
spaceo, line 14: I think it is prettier with array destructing.
Example:
https://stackblitz.com/edit/typescript-udzwh8?file=index.ts

src/core/mount.js
```diff
@@ -37,6 +38,24 @@ function mount(context) {
...
+                    `${error.name}: ${error.message}`,
+                    filename,
+                    lineNumber,
+                    column,
+                    {
+                      ...info,
+                      stack: info.componentStack
+                    }
+                  );
+                }
```
spaceo, line 58: Great that you found a way to forward the errors. I am excited to see it in action!
I think it would be worth commenting here why it is necessary to do so.
And that we could be the whole forwarding block in an external utils function called something like: `forwardError`. In order to reduce clutter in the mount.js file.

src/core/utils/extractErrorDetails.js
```diff
@@ -0,0 +1,31 @@
...
+ * Extracts error details from a stack trace.
+ *
+ * @param {string} stack - The stack trace.
+ * @returns {Object} An object containing filename, lineNumber, and column.
+ */
+const extractErrorDetails = (stack) => {
+  const stackLines = (stack || "").split("\n");
+  const match = stackLines
+    .map((line) => {
+      const parts = line.match(/(?:\()?(.*?):(\d+):(\d+)(?:\))?/);
```
spaceo, line 11: The regex here is not totally clear to me. But I guess I have to see it in action 😅

spaceo (changes requested):
Good work. I have a comment

PR #1262, opened 2024-06-05
Remove reservation text unnecessary info

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFLSBP-678

#### Description
Don't show reservation count in the "successful reservation" modal.

#### Screenshot of the result
-

-

### DIFF
--- a/src/components/material/StockAndReservationInfo.tsx
+++ b/src/components/material/StockAndReservationInfo.tsx
@@ -4,7 +4,7 @@ import { useText } from "../../core/utils/text";
 
 export interface StockAndReservationInfoProps {
   stockCount: number;
-  reservationCount: number;
+  reservationCount?: number;
   numberInQueue?: number;
 }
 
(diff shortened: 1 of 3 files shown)


src/components/material/StockAndReservationInfo.tsx
```diff
@@ -4,7 +4,7 @@ import { useText } from "../../core/utils/text";
 
 export interface StockAndReservationInfoProps {
   stockCount: number;
-  reservationCount: number;
+  reservationCount?: number;
```
spaceo, line 7: Why not remove it totally?
Is it needed elsewhere?
Would be nice to have that kind of information in the commit, so it is possible to follow your thoughts

spaceo (commented):
👍  I have a questione

src/components/material/StockAndReservationInfo.tsx
Adamik10: Yes, it is needed elsewhere :)

Dresse (approved):

PR #1274, opened 2024-06-13
Number advanced search text field labels from 1 upwards

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFLSBP-669

#### Description
Number advanced search text field labels from 1 upwards instead from 0.

#### Screenshot of the result
-

-


Dresse (approved):
LGTM - But some tests are failing

PR #1289, opened 2024-06-21
Don't delete users' data when setting their reservations on hold

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFLSBP-749

#### Description
Don't delete users' data when setting their reservations on hold

#### Screenshot of the result
-

-

### DIFF
--- a/src/apps/reservation-list/modal/pause-reservation/pause-reservation.tsx
+++ b/src/apps/reservation-list/modal/pause-reservation/pause-reservation.tsx
@@ -28,26 +28,24 @@ const PauseReservation: FC<PauseReservationProps> = ({ id, user }) => {
   const { close } = useModalButtonHandler();
   const { pauseReservation } = getModalIds();
   const saveFormId = useId();
-
   const currentDate = dayjs().format("YYYY-MM-DD");
   const [startDate, setStartDate] = useState<string>(currentDate);
   const [endDate, setEndDate] = useState<string>("");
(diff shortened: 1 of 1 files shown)


src/apps/reservation-list/modal/pause-reservation/pause-reservation.tsx
```diff
@@ -41,12 +41,7 @@ const PauseReservation: FC<PauseReservationProps> = ({ id, user }) => {
       }
       // TODO: consolidate with the other save patron function
       // be aware the defaults are not necessarily the same in the different save patron functions
```
Dresse, line 43: This TODO mentions another save patron function. What function is that, and should we do a similar fix there?

src/apps/reservation-list/modal/pause-reservation/pause-reservation.tsx
```diff
@@ -41,12 +41,7 @@ const PauseReservation: FC<PauseReservationProps> = ({ id, user }) => {
...
-      const saveData = {
-        preferredPickupBranch: user.preferredPickupBranch,
-        receiveEmail: user.receiveEmail,
-        receivePostalMail: user.receivePostalMail,
-        receiveSms: user.receiveSms
-      } as Patron;
+      const saveData = user as Patron;
```
Dresse, line 44: I am curios, why didn't the previous logic work?

Dresse (approved):
LGTM 👍 

Have some questions though 😄

PR #1302, opened 2024-06-25
Fix/missing reservations

Adamik10:
#### Link to issue

#### Description
Show backup data from FBS to show all reservations, even those that don't have proper PID. We can't get the material type, but that's okay - we just don't show any label on these reservation items.

#### Screenshot of the result
Note missing reservation for Tue's colleague, Daniel's user:
image

-


Adamik10:
follow in #1303

PR #1303, opened 2024-06-25
Fix missing reservations

Adamik10:
#### Link to issue

#### Description
Show backup data from FBS to show all reservations, even those that don't have proper PID. We can't get the material type, but that's okay - we just don't show any label on these reservation items.

#### Screenshot of the result
Note missing reservation for Tue's colleague, Daniel's user:
image

-

### DIFF
--- a/src/apps/loan-list/materials/selectable-material/selectable-material.tsx
+++ b/src/apps/loan-list/materials/selectable-material/selectable-material.tsx
@@ -13,9 +13,11 @@ import ArrowButton from "../../../../components/Buttons/ArrowButton";
 import { isDigital } from "../../utils/helpers";
 import { listId, ListType } from "../../../../core/utils/types/list-type";
 import SelectableMaterialSkeleton from "./selectable-material-skeleton";
+import { ILLBibliographicRecord } from "../../../../core/fbs/model";
 
 interface SelectableMaterialProps {
   identifier?: string | null;
+  ilBibliographicRecord?: ILLBibliographicRecord;
   disabled?: boolean;
   item?: ListType;
   onMaterialChecked?: (listItem: ListType) => void;
@@ -32,6 +34,7 @@ interface SelectableMaterialProps {
 
 const SelectableMaterial: FC<SelectableMaterialProps & MaterialProps> = ({
   material,
+  ilBibliographicRecord,
   disabled,
   onMaterialChecked,
   selected,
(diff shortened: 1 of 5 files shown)


src/apps/loan-list/materials/selectable-material/selectable-material.tsx
```diff
@@ -13,9 +13,11 @@ import ArrowButton from "../../../../components/Buttons/ArrowButton";
 import { isDigital } from "../../utils/helpers";
 import { listId, ListType } from "../../../../core/utils/types/list-type";
 import SelectableMaterialSkeleton from "./selectable-material-skeleton";
+import { ILLBibliographicRecord } from "../../../../core/fbs/model";
 
 interface SelectableMaterialProps {
   identifier?: string | null;
+  ilBibliographicRecord?: ILLBibliographicRecord;
```
clausbruun, line 20:
```suggestion
  illBibliographicRecord?: ILLBibliographicRecord;
```
And the same for remaining `ilBibliographicRecord`´s

src/apps/loan-list/materials/selectable-material/selectable-material.tsx
```diff
@@ -48,14 +51,24 @@ const SelectableMaterial: FC<SelectableMaterialProps & MaterialProps> = ({
...
+  // There is a chance material isn't provided, if corresponding reservation PID is
+  // incorrect (that's unfortunately possible). Then we use ilBibliographicRecord data.
+  if (!material) {
+    authorsShort = ilBibliographicRecord?.author || "";
+    materialType = "";
+    year = ilBibliographicRecord?.publicationDate || "";
+    title = ilBibliographicRecord?.title || "";
+    lang = ilBibliographicRecord?.language || "";
+  }
+
```
clausbruun, line 71: Maybe use destructuring?

```suggestion
  if (!material) {
    const {
      author = "",
      publicationDate = "",
      title = "",
      language = ""
    } = ilBibliographicRecord || {};
  
    authorsShort = author;
    year = publicationDate;
    lang = language;
    materialType = "";
  }
```

clausbruun (approved):
LGTM 🎉 Just a typo I think and a suggestion to consider.

src/apps/loan-list/materials/utils/material-fetch-hoc.tsx
```diff
@@ -8,12 +8,14 @@ import { Product } from "../../../../core/publizon/model";
 import { BasicDetailsType } from "../../../../core/utils/types/basic-details-type";
 import { mapManifestationToBasicDetailsType } from "../../../../core/utils/helpers/list-mapper";
 import { ListType } from "../../../../core/utils/types/list-type";
+import { ILLBibliographicRecord } from "../../../../core/fbs/model";
```
spaceo, line 11: We should distinguish between our domain language and the domain language of the services.
I do not understand ILLBibliographicRecord. I discussed it with Claus and we agreed that it is probably and old way of defining a material record in FBS.
I think we should take ownership of that kind of entity and define it in our pool of entities: `src/core/utils/types/entities.ts`. We could call the type something like:  `LegacyFbsMaterial`and use that as a prop name as well.

src/core/utils/types/reservation-type.ts
```diff
@@ -13,6 +14,8 @@ interface Reservation extends ListType {
   pickupBranch: string;
   title: string;
   periodical: string;
+  reservationType: string;
+  ilBibliographicRecord?: ILLBibliographicRecord;
```
spaceo, line 18: Please see former comment

spaceo (approved):
Because of time pressure we need to push this through.
But there is some changes that needs to be done. 
We'll create a task to revisit the comments here when we have time on the other side of the summer vacation.

kasperg:
I reimplemented the feature in #1322 without being aware of this existing change.

I think the approach in 1322 is more elegant given the fact that it builds on the the existing `BasicDetailsType` used across the application. This makes the implementation simpe and usable across both reservations and loans across the application.
With that in mind I will close this.

PR #1305, opened 2024-06-27
Use Ereol url if available [DDFLSBP-698]

clausbruun:
#### Link to issue

https://reload.atlassian.net/browse/DDFLSBP-698

#### Description

* We cannot always be certain that the `Ereol` object is located at index 0. To ensure we link to https://ereolen.dk whenever possible, we first attempt to find an `Ereol` object. If it's not found, we then default to using the object at index 0.

* Simplify the URL translation logic in MaterialButtonOnlineExternal to ensure rerendering on externalUrl changes by using a single state for translatedUrl and consolidating effects.

#### Screenshot of the result

Skærmbillede 2024-06-27 kl  14 49 22


Dresse (approved):
LGTM 👍

PR #1321, opened 2024-07-08
DDFFORM 883 deleting materialsearch

LasseStaus:
- **Add error hiddenInputsNotFound.**
- **Rewrite retrieval and usage of selectedWorkId and materialType.**
- **Update / Refactor materialSearch story to reflect changes in the approach of retrieving the workId and type through the DOM.**

#### Link to issue

Jira Ticket: [DDFFORM-883](https://reload.atlassian.net/browse/DDFFORM-883)

Related tickets: 

#### Description

This PR attempts to update the logic for retrieving dpl_fbi_work_id saved values, and use it in the materialSearch app. 

The PR builds ontop of [DDFFORM-900](https://reload.atlassian.net/browse/DDFFORM-900). 

The saved values in drupal are now retrieving using JS instead of being passed a prop. 
This is due to an issue, where formState is not correctly moving the app "value" associated with field, which means that if an entry is deleted from i.e. material-grid, then the rerendered fields could be displaying an incorrect selected material. 

This new logic is added in a new hook. 

Additionally i have added an error that will be displayed, if either of the hiddin fields cannot be found for whatever reason. 

Comment on whether this is fine, needs or story, or should be removed entirely. 

#### Screenshot of the result

image

image

It is important that you test this out in Drupal. You are welcome to test in react, but mostly drupal. 

With this change, as well as [DDFFORM-914](https://reload.atlassian.net/browse/DDFFORM-914)
and [DDFFORM-900](https://reload.atlassian.net/browse/DDFFORM-900), it should be possible to: 

- Create multiple recommendations that will in any variation, keep displaying the correct material that was selected for that field. 

- Create a material grid on the same page that will do the same. 

- Delete any entry in any order, and see that the correct field item and app is being deleted, and not an incorrect one. 

- Save the items with incorrect ids/types ( find the selected and uncheck display:none; for the fields_container ), and when reopening this seeing all individual applications display the appropriate error. 

[DDFFORM-883]: https://reload.atlassian.net/browse/DDFFORM-883?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DDFFORM-900]: https://reload.atlassian.net/browse/DDFFORM-900?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DDFFORM-914]: https://reload.atlassian.net/browse/DDFFORM-914?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/apps/material-search/useGetSelectedWork.ts
+++ b/src/apps/material-search/useGetSelectedWork.ts
@@ -1,4 +1,4 @@
-import { useState } from "react";
+import { useEffect, useState } from "react";
 import { useGetMaterialQuery } from "../../core/dbc-gateway/generated/graphql";
 import { getMaterialTypes } from "../../core/utils/helpers/general";
 import { Work } from "../../core/utils/types/entities";
@@ -36,7 +36,11 @@ const useGetSelectedWork = ({
 
   const [errorState, setErrorState] = useState<ErrorState>(ErrorState.NoError);
 
-  const { data, isLoading: isSelectedWorkLoading } = useGetMaterialQuery(
+  const {
+    data,
+    isLoading: isSelectedWorkLoading,
+    refetch
+  } = useGetMaterialQuery(
     { wid: selectedWorkId },
     {
       enabled: !!selectedWorkId && selectedWorkId.length > 0,
@@ -65,6 +69,12 @@ const useGetSelectedWork = ({
       }
     }
   );
+
+  useEffect(() => {
+    if (selectedWorkId) {
+      refetch();
+    }
+  }, [selectedWorkId, selectedMaterialType, refetch]);
 
   const work = (data?.work as Work) ?? null;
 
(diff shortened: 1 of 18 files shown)


src/apps/material-search/useGetSelectedWork.ts
```diff
@@ -66,6 +70,12 @@ const useGetSelectedWork = ({
     }
   );
 
+  useEffect(() => {
+    if (selectedWorkId) {
+      refetch();
+    }
+  }, [selectedWorkId, selectedMaterialType, refetch]);
```
kasperg, line 67: As I see it this is a problem because we unmount and remount the material search application in combination with the `onSuccess()` handler.

Another approach to do this would be to [clear the react query cache](https://tanstack.com/query/latest/docs/reference/QueryCache/#querycacheclear) [when applications are unmounted](https://github.com/danskernesdigitalebibliotek/dpl-react/blob/309df6a631653e3eee92414181cb52e3bb51245f/src/core/mount.js#L63-L63). This would ensure that they get a fresh state if they are remounted.

kasperg (approved):
:+1: LGTM

PR #1322, opened 2024-07-08
Use FBS record for interlibrary loans and reservations where material cannot be found in FBI

kasperg:
#### Link to issue

#### Description

This PR extends list types used for reservations and loans with the possibility of attaching basic details. This is used when returning data from FBS.

These basic details can then be used as fallback information for physical materials in case these cannot be retrieved from FBI in the material fetch higher-order component.

Without this then the material will not be rendered at all.

One known situation where this occurs is interlibrary loans from the Royal Danish Library.

Add a test showing that this works.

### DIFF
--- a/src/apps/loan-list/list/loan-list.test.ts
+++ b/src/apps/loan-list/list/loan-list.test.ts
@@ -672,6 +672,84 @@ describe("Loan list", () => {
     );
   });
 
+  it("It shows interlibrary record when work is not found", () => {
+    cy.intercept("GET", "**/external/agencyid/patrons/patronid/loans/v2**", {
+      statusCode: 200,
+      body: [
+        {
+          isRenewable: true,
+          isLongtermLoan: false,
+          loanDetails: {
+            loanId: 956250509,
+            materialItemNumber: "3846990827",
+            recordId: "28843238",
+            periodical: null,
+            loanDate: "2022-10-14T16:43:25.325",
+            dueDate: "2022-10-28",
+            loanType: "loan",
+            ilBibliographicRecord: {
+              author: "Thorpe, D.R.",
+              bibliographicCategory: "mono",
+              edition: null,
+              isbn: "9781844135417",
+              issn: null,
+              language: "eng",
+              mediumType: "a xx",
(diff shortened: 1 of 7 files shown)


src/apps/loan-list/list/loan-list.test.ts
```diff
@@ -672,6 +672,84 @@ describe("Loan list", () => {
...
+      body: {
+        data: {}
+      }
+    }).as("work_not_found");
+
+    cy.visit("/iframe.html?path=/story/apps-loan-list--loan-list-entry");
+
+    cy.get(".list-reservation-container")
+      .find(".list-reservation")
+      .get(".list-reservation__header")
```
kasperg, line 738: It is a bit odd that this test for loans uses selectors which mentions reservation.

However this is the case for this entire test and it is outside the scope of this PR to change this.

src/core/utils/helpers/list-mapper.ts
```diff
@@ -68,7 +68,17 @@ export const mapFBSLoanToLoanType = (list: LoanV2[]): LoanType[] => {
...
-      loanId: loanDetails.loanId
+      loanId: loanDetails.loanId,
+      details: loanDetails.ilBibliographicRecord
+        ? {
+            title: loanDetails.ilBibliographicRecord.title,
+            authors: loanDetails.ilBibliographicRecord.author,
+            authorsShort: loanDetails.ilBibliographicRecord.author,
+            firstAuthor: loanDetails.ilBibliographicRecord.author,
+            year: loanDetails.ilBibliographicRecord.publicationDate,
+            lang: loanDetails.ilBibliographicRecord.language
```
Dresse, line 79: Do we not need LSBN for loans?

Dresse (approved):
Looks good. Just have a single question. Approved 👍

Dresse (approved):
LGTM 👍

src/core/utils/helpers/list-mapper.ts
kasperg: :+1: Fixed.

PR #1323, opened 2024-07-10
Implement fallback for material page.

LasseStaus:
DDFFORM-903

#### Link to issue

Jira [DDFFORM-903](https://reload.atlassian.net/browse/DDFFORM-903)

#### Description

If the specified type in the URL does not exist in the data, we will fallback to the type in bestRepresentation.

This change is implemented because currently, if an incorrect type is specified, the page will load indefinitely.

Run it locally or on the Pr. 
Go the the page of a material. 

Change the type in the URL to something that doesn't exist. 

Verify that the fallback to bestRepresentation works. 

If you want to verify bestRep, look at the graphql data. 

[DDFFORM-903]: https://reload.atlassian.net/browse/DDFFORM-903?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/apps/material/material.tsx
+++ b/src/apps/material/material.tsx
@@ -1,43 +1,43 @@
-import React, { useEffect, useState } from "react";
-import VariousIcon from "@danskernesdigitalebibliotek/dpl-design-system/build/icons/collection/Various.svg";
 import CreateIcon from "@danskernesdigitalebibliotek/dpl-design-system/build/icons/collection/Create.svg";
 import Receipt from "@danskernesdigitalebibliotek/dpl-design-system/build/icons/collection/Receipt.svg";
+import VariousIcon from "@danskernesdigitalebibliotek/dpl-design-system/build/icons/collection/Various.svg";
+import React, { useEffect, useState } from "react";
 import { useDeepCompareEffect } from "react-use";
-import { WorkId } from "../../core/utils/types/ids";
+import DisclosureControllable from "../../components/Disclosures/DisclosureControllable";
+import DisclosureSummary from "../../components/Disclosures/DisclosureSummary";
+import DigitalModal from "../../components/material/digital-modal/DigitalModal";
+import InfomediaModal from "../../components/material/infomedia/InfomediaModal";
+import { hasCorrectAccess } from "../../components/material/material-buttons/helper";
 import MaterialDescription from "../../components/material/MaterialDescription";
-import { MaterialReviews } from "../../components/material/MaterialReviews";
-import MaterialMainfestationItem from "../../components/material/MaterialMainfestationItem";
-import { useText } from "../../core/utils/text";
 import MaterialDetailsList from "../../components/material/MaterialDetailsList";
+import MaterialHeader from "../../components/material/MaterialHeader";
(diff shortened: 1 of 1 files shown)


src/apps/material/material.tsx
```diff
@@ -104,25 +106,32 @@ const Material: React.FC<MaterialProps> = ({ wid }) => {
...
-    // If there is no type in the url, we select one.
-    if (!type) {
-      const bestMaterialType = getBestMaterialTypeForWork(work);
-      setSelectedManifestations(manifestationsByMaterialType[bestMaterialType]);
-      setQueryParametersInUrl({
-        type: bestMaterialType
-      });
+
+    if (urlType && urlTypeIsPresentInManifestations) {
+      setSelectedManifestations(manifestationsByMaterialType[urlType]);
```
kasperg, line 127: Consider refactoring this code.

As I read it we split `work.manifestations.all` by type twice:

1. When running `filterManifestationsByType()` to determine `urlTypeIsPresentInManifestations`
2. When running `divideManifestationsByMaterialType()` to determine `manifestationsByMaterialType`

I am not sure how computationally expensive is but consider if we can do it once. One way would be to determine `urlTypeIsPresentInManifestations` from `manifestationsByMaterialType`.

src/apps/material/material.tsx
```diff
@@ -1,43 +1,45 @@
...
 import Receipt from "@danskernesdigitalebibliotek/dpl-design-system/build/icons/collection/Receipt.svg";
+import VariousIcon from "@danskernesdigitalebibliotek/dpl-design-system/build/icons/collection/Various.svg";
+import React, { useEffect, useState } from "react";
 import { useDeepCompareEffect } from "react-use";
-import { WorkId } from "../../core/utils/types/ids";
+import DisclosureControllable from "../../components/Disclosures/DisclosureControllable";
+import DisclosureSummary from "../../components/Disclosures/DisclosureSummary";
+import DigitalModal from "../../components/material/digital-modal/DigitalModal";
+import InfomediaModal from "../../components/material/infomedia/InfomediaModal";
+import { hasCorrectAccess } from "../../components/material/material-buttons/helper";
```
kasperg, line 10: The changes to the dependencies are not relevant to the change at hand. I expect it to be the result of a local plugin or the like which orders dependencies by internal/external ad then alphabetically. This is not really something that is part of our general setup/code style.

I do not mind the change as such but next time around I would like to see it in a separate commit.

If it is something that we would like to enforce in general we should move it to our eslint/prettier setup.

kasperg (approved):
:+1: Nice change.

I think this can be merged as is but I have a couple of comments for you to consider before you do.

src/apps/material/material.tsx
LasseStaus: That's fair. For the record, i think something uniform should be added and enforced then, but i won't do it as part of this 👍

src/apps/material/material.tsx
LasseStaus: Updated the code to only do one check 👍

PR #1348, opened 2024-07-26
Availability bug

spaceo:
#### Link to issue

#### Description

#### Screenshot of the result


tsconfig.json
```diff
@@ -8,11 +8,15 @@
       "vitest/importMeta"
     ],
     "target": "es6",
+    "moduleResolution": "nodenext",
+    "module": "NodeNext"
```
Adamik10, line 12: Yeah, should this be looked into? The error we're getting from GH actions.

src/tests/unit/availability.test.tsx
```diff
@@ -0,0 +1,452 @@
...
+      useGetAvailabilityV3: vi.fn()
+    }));
+    vi.mock("../../core/utils/config", () => ({
+      useConfig: vi.fn()
+    }));
+
+    // Make sure that the config hook returns an array with an empty string.
+    // In that way we do not have any blacklisted branches (they are not needed for the test).
+    // Typescript does not understand our mocked hook.
+    // So we gracefully ignore the error :).
```
Adamik10, line 35: Great with a comment. And also amazing with the tests ✨

src/components/availability-label/usePhysicalAvailabilityData.ts
```diff
@@ -0,0 +1,74 @@
...
+  // If hook is not enabled make it clear that the loading and availability status is unknown.
+  if (!enabled) {
+    return {
+      isLoading: null,
+      isAvailable: null
+    };
+  }
+
+  // Articles are always available.
+  if (manifestText === ManifestationMaterialType.article) {
```
Adamik10, line 44: Okay, so there is something that I think should probably be updated here (and also wherever we check for whether it is an article). 
Since this was made, we added new manifestation types that are considered an "article" - under the `ManifestationMaterialType`. There are these different types of articles - article, paperArticle, onlineArticle, and earticle. See picture below.
image

I believe all of these are always supposed to be available, so maybe we should introduce a helper function that checks whether the `manifestText` is any of these, instead of just the `ManifestationMaterialType.article` .

Adamik10 (commented):
I have one comment that I think you should consider before moving on with this :) 
Otherwise I can't see any more potential problems :) Fingers crossed that it all works as they intended, since DDF has soooo much business logic in regard to availability 🤯
Luckily, someone from DDF (probably Tue) will be reviewing this too.

tsconfig.json
spaceo: On it. Thought that I solved it with the one commit. But we still get a TS warning here in GH.

src/components/availability-label/usePhysicalAvailabilityData.ts
spaceo: It current code onlyManifestationMaterialType.article is being considered:
https://github.com/danskernesdigitalebibliotek/dpl-react/blob/develop/src/components/availability-label/helper.ts
Let's get this through and create a new task about this problem.

tsconfig.json
spaceo: Ok. So there is a difference in what my local linter and GH's linter is reporting here. Furthermore: The changes in this PR is not introducing anything that should cause this problem.
I suggest that instead of me headlessly trying to solve the problem, we should do it in a controlled manner. And therefor I will ignore the warnings for now.

Adamik10 (approved):
Love it ❤️

src/components/availability-label/usePhysicalAvailabilityData.ts
spaceo: We decided to fix it in current PR anyways...

PR #1358, opened 2024-08-01
Updated the required check to display empty string instead of null.

Dresse:
#### Link to issue

https://reload.atlassian.net/browse/DDFLSBP-690

#### Description
Updated the required check to display empty string instead of null.
Also added an extra space between the asterisk to comply with how required fields are indidcated in the rest of the project.

Similar changes has been done in dpl-cms and dpl-react
https://github.com/danskernesdigitalebibliotek/dpl-design-system/pull/681
https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/1363

### DIFF
--- a/src/apps/patron-page/util/BranchesDropdown.tsx
+++ b/src/apps/patron-page/util/BranchesDropdown.tsx
@@ -28,7 +28,7 @@ const BranchesDropdown: FC<BranchesDropdownProps> = ({
         htmlFor="branches-dropdown"
         className="text-body-medium-medium mt-32 mb-8"
       >
-        {`${t("pickupBranchesDropdownLabelText")}${required ? "*" : null}`}
+        {`${t("pickupBranchesDropdownLabelText")}${required ? " *" : ""}`}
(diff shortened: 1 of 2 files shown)


src/apps/patron-page/util/BranchesDropdown.tsx
```diff
@@ -28,7 +28,7 @@ const BranchesDropdown: FC<BranchesDropdownProps> = ({
         htmlFor="branches-dropdown"
         className="text-body-medium-medium mt-32 mb-8"
       >
-        {`${t("pickupBranchesDropdownLabelText")}${required ? "*" : null}`}
+        {`${t("pickupBranchesDropdownLabelText")}${required ? " *" : ""}`}
```
spaceo, line 31: I think it could be nice to make an atom component (maybe organised in the atom compinent folder in a sub flder called forms?).
It could be a component that receives two props:
required: boolean
text: string
That would make this place and the `TextInput`label place look more neat I think.

PR #1365, opened 2024-08-05
Feature/find on shelf changes

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFLSBP-712

#### Description
This PR adds the author name to the end of the Find on shelf modal from the nameSort property that exists on the Creators entity.
In order to get TS to be happy, I had to cast into one specific function. Let me know if you see a better way to deal with that, much appreciated 😁

#### Screenshot of the result
-

ALSO UPDATE DPL DESIGN SYSTEM BEFORE MERGING THIS 
https://github.com/danskernesdigitalebibliotek/dpl-design-system/pull/668


src/apps/material-search/MaterialSearchList.tsx
```diff
@@ -58,7 +59,7 @@ const MaterialSearchListResults: FC<MaterialSearchListResultsProps> = ({
       </div>
       <ol className="material-search-list__results">
         {data.map((work, index) => {
-          const authors = flattenCreators(work.creators);
+          const authors = flattenCreators(work.creators as Work["creators"]);
```
spaceo, line 62: I think we should try to infer the `Work` type as early as possible so we do not need to use type assertions later on.
Please look at this [suggestion](https://stackblitz.com/edit/vitejs-vite-hzqbpm?file=src%2FMaterialSearchList.tsx%3AL54&terminal=dev)

src/components/find-on-shelf/FindOnShelfModalBody.tsx
```diff
@@ -55,6 +57,10 @@ const FindOnShelfModalBody: FC<FindOnShelfModalBodyProps> = ({
     config
   });
   const author = creatorsToString(flattenCreators(authors), t);
+  // Inverted author - last name, then first name is used in find on shelf modal.
+  const invertedAuthor = capitalizeFirstLetters(
+    creatorsToString(flattenCreatorsLastNameFirst(authors), t)
```
spaceo, line 62: Nice way of composing existing functions 👍

src/core/utils/helpers/general.ts
```diff
@@ -47,6 +54,11 @@ export const flattenCreators = (creators: Work["creators"]) =>
     return creator.display;
   });
 
+export const flattenCreatorsLastNameFirst = (creators: Work["creators"]) =>
+  creators.map((creator: Work["creators"][0]) => {
```
spaceo, line 58: I dont think it is necessary to define the creators type again. TS is capable of inferring that by itself:
```suggestion
  creators.map((creator) => {
```

spaceo (changes requested):
I have some commentz...

src/apps/material-search/MaterialSearchList.tsx
Adamik10: That's a great way to do it! thx for the suggestion ✨

spaceo (approved):
LGTM 👍

PR #1399, opened 2024-08-29
Ddfform 822 modal til abningstider

kasperbirch1:
#### Links
https://reload.atlassian.net/browse/DDFFORM-822

#### Depend on
https://github.com/danskernesdigitalebibliotek/dpl-design-system/pull/713
https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/1528

#### Description
This pull request introduces the `OpeningHoursSidebar` React app. The app is designed to render a button that opens the sidebar modal, with the button appearing either as a small icon or a large opening hours clock. To optimize performance in Drupal, we deploy two instances of the same app, where the size prop determines which button markup should be rendered, thereby reducing the amount of JavaScript that needs to be loaded.

#### Test
https://varnish.pr-1528.dpl-cms.dplplat01.dpl.reload.dk/


rasben:
I feel pretty lost trying to review this, so i'm gonna take my self off - maybe we can get @Adamik10 or @LasseStaus to look at it? :)

src/apps/opening-hours-sidebar/OpeningHoursSidebarButtonLarge.tsx
```diff
@@ -0,0 +1,36 @@
+import React, { FC } from "react";
+import iconWatch from "@danskernesdigitalebibliotek/dpl-design-system/build/icons/basic/icon-watch-static.svg";
+import { useText } from "../../core/utils/text";
+
+type ButtonLargeProps = {
+  openDialog: () => void;
+};
+
+const ButtonLarge: FC<ButtonLargeProps> = ({ openDialog }) => {
```
Adamik10, line 9: What is the difference in usage between the small and big button? 
Would it make sense to make a story with one and another with the other?

src/apps/opening-hours-sidebar/OpeningHoursSidebarSidebar.tsx
```diff
@@ -0,0 +1,45 @@
+import React, { FC } from "react";
+import iconWatch from "@danskernesdigitalebibliotek/dpl-design-system/build/icons/basic/icon-watch-static.svg";
+import DisclosureControllable from "../../components/Disclosures/DisclosureControllable";
+import OpeningHoursSidebarSummary from "./OpeningHoursSidebarSummary";
+import OpeningHoursSidebarDetails from "./OpeningHoursSidebarDetails";
+import { LibraryType, toDayString } from "./helper";
+import { useText } from "../../core/utils/text";
+
+const OpeningHoursSidebarSidebar: FC<{ libraries: LibraryType[] }> = ({
```
Adamik10, line 9: Would it make more sense to call this something like `OpeningHoursSidebarContent` instead? I think the double sidebar maybe doesn't make much sense.

kasperbirch1: Yes, that's a good ide ☺️

kasperbirch1:
@Adamik10 

Yes, it is correct. Most of the data comes from Drupal. If you want to see it together with the content and CSS, please check https://varnish.pr-1528.dpl-cms.dplplat01.dpl.reload.dk/frontpage, or you can use Wiremock locally. 🙈

Adamik10 (approved):
👍

PR #1405, opened 2024-09-02
Fix favourites list

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFBRA-23

#### Description
This PR Fixes a minor skeleton screen on the favourites list. It also expands what a FaustID can look like based on new findings - it can contain an underscore "_". This was breaking the favourites list page before.

#### Screenshot of the result
-

-

### DIFF
--- a/src/apps/favorites-list/FavoritesList.tsx
+++ b/src/apps/favorites-list/FavoritesList.tsx
@@ -44,8 +44,8 @@ const FavoritesList: React.FC<FavoritesListProps> = ({ pageSize }) => {
 
   const skeletonList = (
     <>
-      <div className="ssc">
-        <div className="ssc-line w-10 my-32">&nbsp;</div>
+      <div className="ssc content-list">
+        <div className="ssc-line w-10 mt-32">&nbsp;</div>
       </div>
       <ul className="content-list">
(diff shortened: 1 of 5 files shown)


src/apps/favorites-list/FavoritesList.tsx
```diff
@@ -44,8 +44,8 @@ const FavoritesList: React.FC<FavoritesListProps> = ({ pageSize }) => {
 
   const skeletonList = (
     <>
-      <div className="ssc">
-        <div className="ssc-line w-10 my-32">&nbsp;</div>
+      <div className="ssc content-list">
+        <div className="ssc-line w-10 mt-32">&nbsp;</div>
```
Dresse, line 48: This is tailwind classes, or? 

Just currious 😄

Dresse (approved):
LGTM 👍

src/apps/favorites-list/FavoritesList.tsx
Adamik10: hihi, nopes - we're using a skeleton screen package that has some classes 😉 like ssc, ssc-line. and w-10

PR #1423, opened 2024-09-13
Use MaterialTypesGeneral instead of WorkTypes in filters

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFBRA-48

#### Description
This PR corrects a mistake where we were using incorrect filters in the facet browser.

#### Screenshot of the result
-

-


Dresse (approved):
LGTM 👍

PR #1434, opened 2024-09-16
DDFHER-67 - Add checkbox for `receiveSms` when creating user

kasperbirch1:
#### Link to issue

https://reload.atlassian.net/browse/DDFHER-67

#### Description

This pull request enables users to subscribe to SMS notifications (`receiveSms`) if the library allows SMS notifications (`text-notifications-enabled-config`) during the user creation process. If the user opts in, their phone number will be displayed as "mobile (notif)" in Cicero; otherwise, it will continue to appear under "fastnet."

#### Test

Create an adult user at Copenhagen Library via Frontpage | DPL CMS

During user creation, there is now a checkbox below the phone number.

If this is checked, the number in Cicero will be listed as “mobile (notif)”, otherwise it will be listed as “fastnet”.

https://varnish.pr-1570.dpl-cms.dplplat01.dpl.reload.dk/

image

image

### DIFF
--- a/src/apps/create-patron-user-info/UserInfo.tsx
+++ b/src/apps/create-patron-user-info/UserInfo.tsx
@@ -47,6 +47,7 @@ const UserInfo: FC<UserInfoProps> = ({ cpr, registerSuccessCallback }) => {
     set(copyUser, key, newValue);
     setPatron(copyUser);
   };
+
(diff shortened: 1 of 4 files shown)


Dresse (approved):
LGTM 👍 
Just a minor question, should this also be added to the registration page?

PR #1440, opened 2024-09-18
DDFHER-76 - Add "location" and "sublocation" URL parameters in advanced search (CQL)

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFHER-76

#### Depend on
https://github.com/danskernesdigitalebibliotek/dpl-design-system/pull/735
https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/1585

#### Description

This pull request implements URL parameters for `location` and `sublocation`. Additionally, the parameters from the URL are processed in `useEffectOnce`, ensuring they are included on the first load. The `removeQueryParametersFromUrl` function is used to remove the parameters if their strings are empty.

It also refactors the logic to allow users to search across multiple locations or sublocations by using a comma-separated string in the input fields.

### Test
http://varnish.pr-1585.dpl-cms.dplplat01.dpl.reload.dk/advanced-search

https://github.com/user-attachments/assets/7dc1f75e-e3ee-4141-8f4f-9d18202156f4


src/apps/advanced-search/CqlSearchHeader.tsx
```diff
@@ -51,13 +79,17 @@ const CqlSearchHeader: React.FC<CqlSearchHeaderProps> = ({
         id="location"
         label="Location"
```
kasperbirch1, line 80: Rember translation

src/apps/advanced-search/CqlSearchHeader.tsx
```diff
@@ -51,13 +79,17 @@ const CqlSearchHeader: React.FC<CqlSearchHeaderProps> = ({
...
         type="text"
-        onChange={(location) => onLocationChange(location)}
+        onChange={(location) => handleInputChange("location", location)}
+        value={inputValues.location}
       />
       <TextInput
-        id="location"
+        id="sublocation"
         label="Sublocation"
```
kasperbirch1, line 87: And here

src/apps/advanced-search/CqlSearchHeader.tsx
```diff
@@ -30,42 +34,77 @@ const CqlSearchHeader: React.FC<CqlSearchHeaderProps> = ({
...
         data-cy={dataCy}
       >
         {t("cqlSearchTitleText")}
       </h1>
-      <textarea
-        className="advanced-search__cql-input focus-styling__input"
+
+      <Textarea
+        id="cql"
+        name="name"
```
kasperg, line 73: Please exclude `name` if it is not necessary.

It is not something we use for the other form components like `TextInput`. 

I do not think the current value provides any meaning either.

src/components/forms/label/Label.tsx
```diff
@@ -0,0 +1,16 @@
+import clsx from "clsx";
+import React, { FC } from "react";
+
+export interface LabelProps {
+  id: string;
+  children: string;
+  className?: string;
+}
+
+const Label: FC<LabelProps> = ({ id, className, children }) => (
```
kasperg, line 10: :+1: for adding this to be used with the `Textarea` component.

However I think it is a problem that we do not use it within the current `TextInput` component which served a similar purpose.

I suggest that you either use the new `Label` component both places or drop it for now and fall back to a simple `<label>` element.

src/components/forms/textarea/Textarea.tsx
kasperg, line 1: Please consider the location of this component (and `Label`).

This creares a new `forms` directory` but we already have `TextInput` recide within `atoms/input`.

Either move this component within input or move `TextInput` into `forms`.

kasperg (changes requested):
:+1: This seems to work well.

I am not totally satisfied about the code structure in advanced search in the way we handle url parameters etc. but that is none of your fault. The implementation follows what is.

I have a few comments regarding the code structure of the next form components.

I appreciate the updated design for the form elements. I think the form looks better.
Consider if they take up too much vertical space between the title and the form.

Monosnap | DPL CMS 2024-10-01 10-53-36

As I see it some of the issue is caused by different approaches in the "normal" form and in the CQL form. One uses margins. The other grip/gap.

src/components/forms/textarea/Textarea.tsx
kasperbirch1: I want to move it to reflect the folder structure in dpl-design-system

.storybook/preview.js
```diff
@@ -51,7 +45,11 @@ const App = ({ story }) => <Store>{WrappedStory(story)}</Store>;
 // Consideration for the future - using addon-redux could bring value.
 // It wasn't implemented to begin with because it wasn't compatible with Storybook 6.
 export const decorators = [
-  Story => <><App story={Story} /></>
+  (Story) => (
```
kasperbirch1, line 48: This was a Prettier (format on save) change. However, it should be the same result.

kasperg (approved):

PR #1442, opened 2024-09-19
Ensure creatorsToString() always returns a string

kasperg:
#### Link to issue

https://reload.atlassian.net/browse/DDFNEXT-519

#### Description

The current implementation will return undefined if the creators array is empty. This causes other functions which expect a string (which is what TypeScript assumes) to fail.

Add a return type to the function to assert the intention of the function. We could also have allowed it to return undefined if the creators array is empty but that seems to be against the naming of the function, creatorsToString.

Add a check to handle single array and empty creators array separately. If there are no creators then return an empty string.

You can test this by checking the Periodical story for the Material app before and after the change.

### DIFF
--- a/src/core/utils/helpers/general.ts
+++ b/src/core/utils/helpers/general.ts
@@ -1,6 +1,6 @@
 import { useEffect, useRef } from "react";
 import dayjs from "dayjs";
-import { uniq } from "lodash";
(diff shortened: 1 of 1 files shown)


kasperbirch1 (approved):
LGTM 👍

PR #1466, opened 2024-10-01
DDFHER-90 - Refactor `getInstantLoanBranchHoldings` to use `materialGroup.name`

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFHER-90

#### Description

This pull request refactors the `getInstantLoanBranchHoldings` method to use `materialGroup.name` instead of `.description`. This change addresses the issue where `.description` is an optional field, potentially leading to inconsistencies or missing data during the execution of this method.

#### Test 
https://varnish.pr-1608.dpl-cms.dplplat01.dpl.reload.dk/

image

### DIFF
--- a/cypress/fixtures/material/instant-loan/holdings.json
+++ b/cypress/fixtures/material/instant-loan/holdings.json
@@ -24,7 +24,7 @@
             "available": false,
             "periodical": null,
             "materialGroup": {
-              "name": "standard",
+              "name": "31 dages lånetid til alm lånere",
               "description": "31 dages lånetid til alm lånere"
             }
           },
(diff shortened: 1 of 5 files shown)


cypress/fixtures/material/instant-loan/holdings.json
```diff
@@ -24,7 +24,7 @@
             "available": false,
             "periodical": null,
             "materialGroup": {
-              "name": "standard",
+              "name": "31 dages lånetid til alm lånere",
```
kasperg, line 27: Consider updating the test instead of the test data.

"31 dages lånetid til alm lånere" is not really a value which can be expected for material group names. Such alterations to fixtures which have been copy/pasted from live data make the fixture difficult to maintain.

If a change is necessary and if you want to do something else then update the test to react to the name "standard" instead.

kasperg (approved):
:+1: Nicely done.

PR #1481, opened 2024-10-09
Ddfbra 47 flere samtidige reserveringer

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFBRA-47

#### Description
This PR disables the "approve reservation" button if the user already clicked it. This prevents multiple requests from being fired before resolving. 
I also took the liberty to (for consistency's sake) unite the way we work with request state in regards to reservation - between the reservation modal and edit reservation modal in the user profile (reservations list page).

#### Screenshot of the result
-

-

### DIFF
--- a/src/components/reservation/ReservationModalBody.tsx
+++ b/src/components/reservation/ReservationModalBody.tsx
@@ -135,7 +135,8 @@ export const ReservationModalBody = ({
     reservableManifestations ?? [],
     !!selectedPeriodical
   );
-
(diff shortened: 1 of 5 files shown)


Dresse (approved):
So in the end, the problem was located in our end? It was simply because the request was taking longer time, and in the meantime, the user would continue clicking the "Reservation" button multiple times?

Nice find Adam :) 👍

PR #1482, opened 2024-10-11
DDFHER-100 - Implement `useAvailabilityBlacklist` for `MaterialAvailabilityTextPhysical` to manage materials count

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFHER-100

#### Description
The materials count in the phrase "The library has XX copies" will now be regulated by the availability blacklist.

The current logic for displaying the number of copies in "The library has XX copies" takes the total count of materials but does not account for availability restrictions. This leads to incorrect information being presented to users when certain copies are unavailable.

This pull request enables the `useAvailabilityBlacklist` in the `useGetHoldings` hook, which is used in the `MaterialAvailabilityTextPhysical` component. With this update, the number of copies shown will now be controlled by the availability blacklist, ensuring that only accessible copies are displayed.


src/apps/material/helper.ts
```diff
@@ -512,23 +512,32 @@ export const getBlacklistedQueryArgs = (
...
+    return {
+      ...args,
+      ...formatBranches([
+        branchesFromConfig("availability", config),
+        branchesFromConfig("pickup", config)
+      ])
+    };
+  }
+
+  return args;
```
kasperg, line 540: Please reconsider this change.

Reading through the code I cannot see how the updated implementation if different from the current one - except being more verbose.

src/apps/material/helper.ts
```diff
@@ -543,19 +552,18 @@ export const getAvailability = async ({
...
-  useAvailabilityBlacklist?: boolean;
+  blacklist: BlacklistType;
   options?: {
     query?: UseQueryOptions<Awaited<ReturnType<typeof getHoldingsV3>>>;
   };
 }) => {
-  const blacklistedBranches = useAvailabilityBlacklist ? "both" : "pickup";
   const { data, isLoading, isError } = useGetHoldingsV3(
-    getBlacklistedQueryArgs(faustIds, config, blacklistedBranches),
+    getBlacklistedQueryArgs(faustIds, config, blacklist),
```
kasperg, line 566: :+1: This change makes sense to me.

We could consider using the same default value as was used before. I think it is OK to make it explicit though.

src/components/material/MaterialAvailabilityText/physical/MaterialAvailabilityTextPhysical.tsx
```diff
@@ -21,7 +21,8 @@ const MaterialAvailabilityTextPhysical: React.FC<
   const faustIds = convertPostIdsToFaustIds(pids);
   const { data, isLoading, isError } = useGetHoldings({
     faustIds,
-    config
+    config,
+    blacklist: "availability"
```
kasperg, line 25: :+1: This previously used the default value `false` for `useAvailabilityBlacklist`. This was mapped to the `pickup` type. Now it is the `availability` type. This is as requested in the ticket.

src/components/find-on-shelf/FindOnShelfModalBody.tsx
```diff
@@ -54,7 +54,7 @@ const FindOnShelfModalBody: FC<FindOnShelfModalBodyProps> = ({
   );
   const { data, isLoading } = useGetHoldings({
     faustIds: faustIdArray,
-    useAvailabilityBlacklist: true,
+    blacklist: "availability",
```
kasperg, line 57: I am unsure whether this is as requested.

`useAvailabilityBlacklist: true` was mapped to the `both` type. Now it is the `availability` type.

src/components/material/periodical/MaterialPeriodical.tsx
```diff
@@ -22,7 +22,8 @@ const MaterialPeriodical: FC<MaterialPeriodicalProps> = ({
 
   const { data, isLoading, isError } = useGetHoldings({
     faustIds: [faustId],
-    config
+    config,
+    blacklist: "pickup"
```
kasperg, line 26: :+1: This previously used the default value `false` for `useAvailabilityBlacklist`. This was mapped to the `pickup` type. Now it is the `pickup` type. Behavior should be unchanged.

src/components/reservation/ReservationModalBody.tsx
```diff
@@ -126,7 +126,11 @@ export const ReservationModalBody = ({
   const { mutate: mutateAddReservations } = useAddReservationsV2();
   const { mutate: mutateOpenOrder } = useOpenOrderMutation();
   const userResponse = usePatronData();
-  const holdingsResponse = useGetHoldings({ faustIds, config });
+  const holdingsResponse = useGetHoldings({
+    faustIds,
+    config,
+    blacklist: "availability"
+  });
```
kasperg, line 133: I am unsure whether this is as requested.

This previously used the default value `false` for `useAvailabilityBlacklist`. This was mapped to the `pickup` type. Now it is the `availability` type.

src/core/utils/useReservableFromAnotherLibrary.tsx
```diff
@@ -13,7 +13,7 @@ const useReservableFromAnotherLibrary = (
   const config = useConfig();
   const { data: holdingsData } = useGetHoldings({
     faustIds: getAllFaustIds(manifestations),
-    useAvailabilityBlacklist: true,
+    blacklist: "both",
```
kasperg, line 16: :+1: `useAvailabilityBlacklist: true` was mapped to the `both` type. Now it is the `availability` type. Behavior should be unchanged.

kasperg (approved):

PR #1491, opened 2024-10-14
Ddfbra 100 filtrering sker ikke

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFBRA-100

#### Description
This PR fixes an issue for PR #790 - where category search didn't work. This was happening due to the FBI API spec change, where all the facets are now gathered inside the `FacetFieldEnum` which is CAPITALIZED, but the simple search query filters cannot be. So we now have a way to map between them. 

In the last commit ( f540ae59 ) I also took the liberty to fix an issue where persisted filters keep stacking up between searches. If you search for Harry `in books` and then follow up with Adam `in e-books` you would end up with a search for Adam `in books and ebooks`. So now whenever the user searches we reset persisted filters before redirecting to the search page.

#### Screenshot of the result
-

-

### DIFF
--- a/src/core/storybook/serviceUrlArgs.ts
+++ b/src/core/storybook/serviceUrlArgs.ts
@@ -91,7 +91,7 @@ export default {
   [serviceUrlKeys.dplCms]: process.env.CMS_BASEURL ?? "https://dpl-cms.docker",
   [serviceUrlKeys.cover]: "https://cover.dandigbib.org",
   [serviceUrlKeys.materialList]: "https://prod.materiallist.dandigbib.org",
-  [serviceUrlKeys.fbi]: "https://fbi-api.dbc.dk/next-present/graphql",
-  [serviceUrlKeys.fbiLocal]: "https://fbi-api.dbc.dk/next/graphql",
-  [serviceUrlKeys.fbiGlobal]: "https://fbi-api.dbc.dk/next-present/graphql"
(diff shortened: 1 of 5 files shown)


src/core/storybook/serviceUrlArgs.ts
```diff
@@ -91,7 +91,7 @@ export default {
   [serviceUrlKeys.dplCms]: process.env.CMS_BASEURL ?? "https://dpl-cms.docker",
   [serviceUrlKeys.cover]: "https://cover.dandigbib.org",
   [serviceUrlKeys.materialList]: "https://prod.materiallist.dandigbib.org",
-  [serviceUrlKeys.fbi]: "https://fbi-api.dbc.dk/next-present/graphql",
-  [serviceUrlKeys.fbiLocal]: "https://fbi-api.dbc.dk/next/graphql",
-  [serviceUrlKeys.fbiGlobal]: "https://fbi-api.dbc.dk/next-present/graphql"
+  [serviceUrlKeys.fbi]: "https://temp.fbi-api.dbc.dk/next-present/graphql",
+  [serviceUrlKeys.fbiLocal]: "https://temp.fbi-api.dbc.dk/next/graphql",
+  [serviceUrlKeys.fbiGlobal]: "https://temp.fbi-api.dbc.dk/next-present/graphql"
```
Dresse, line 96: Is this because of the updated FBI API?

Dresse (approved):
LGTM, just have a single question 👍

PR #1501, opened 2024-10-24
Upgrade Cypress

JacobArrow:
#### Link to issue

https://reload.atlassian.net/browse/DDFBRA-136

#### Description

- Bump Cypress to v13.15.0
- Disable Test Isolation
- use 'first()' on 'within()'
- Import Cypress plugins

### DIFF
--- a/cypress.config.js
+++ b/cypress.config.js
@@ -6,13 +6,13 @@ module.exports = defineConfig({
   defaultCommandTimeout: 10000,
   pageLoadTimeout: 10000,
   requestTimeout: 30000,
-  testFiles: "@(apps|components)/**/*.test.@(ts|tsx)",
-  integrationFolder: "./src",
   retries: {
     runMode: 3,
     openMode: 0
   },
   e2e: {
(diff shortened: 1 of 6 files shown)


cypress.config.js
```diff
@@ -1,18 +1,23 @@
...
   retries: {
     runMode: 3,
     openMode: 0
   },
   e2e: {
-    baseUrl: "http://localhost:57021"
+    supportFile: "cypress/support/index.ts",
+    specPattern: "./src/@(apps|components)/**/*.test.@(ts|tsx)",
+    baseUrl: "http://localhost:57021",
+    testIsolation: false,
```
kasperbirch1, line 18: Good workaround for now. But I think that our tests should be updated in the future so that they are isolated

kasperbirch1 (approved):
LGTM 👍  But I think @kasperg should also approve it

kasperg (approved):
:+1: Nice. I appreciate the effort made in splitting the work into individual commits.

PR #1503, opened 2024-10-25
Upgrade eslint-plugin-jsx-a11y

JacobArrow:
#### Link to issue

https://reload.atlassian.net/browse/DDFNEXT-207

#### Description

Upgrade eslint-plugin-jsx-a11y to version 6.10.0


kasperg (approved):
:+1: It seems like a later version makes the previous issue go away.

PR #1505, opened 2024-10-25
Upgrade unfetch

JacobArrow:
#### Link to issue

https://reload.atlassian.net/browse/DDFNEXT-232

#### Description

Upgraded unfetch to v5.0.0


kasperg (approved):
:+1: Nice and simple update. Must be because we have moved other dependencies since the initial attempt.

PR #1538, opened 2024-11-11
Upgrade Typescript to newest version

JacobArrow:
#### Link to issue

https://reload.atlassian.net/browse/DDFBRA-181

#### Description

- Upgrade Typescript to version 5.6.3
- Cleanup Cypress commands
- Fix type errors caused by Typescript upgrade

### DIFF
--- a/cypress/support/index.ts
+++ b/cypress/support/index.ts
@@ -86,22 +86,18 @@ Cypress.Commands.add(
 
 // Data cy attribute selector helpers.
 const visible = (checkVisible: boolean) => (checkVisible ? ":visible" : "");
-Cypress.Commands.add("getBySel", (selector, checkVisible = false, ...args) => {
-  return cy.get(`[data-cy="${selector}"]${visible(checkVisible)}`, ...args);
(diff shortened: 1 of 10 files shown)


cypress/support/index.ts
```diff
@@ -123,7 +115,6 @@ declare global {
        */
       createFakeLibrarySession(): void;
       createFakeAuthenticatedSession(): void;
-      createFakeLibrarySession(): void;
```
JacobArrow, line 126: It seems the dublication was introduced due to an error while rebasing, according to the original PR

src/core/utils/withSuffix.tsx
```diff
@@ -2,13 +2,14 @@ import { ActionCreatorWithPayload } from "@reduxjs/toolkit";
 import React from "react";
 import { store } from "../store";
 
-export default <T,>(
+export default function withSuffix<T extends object>(
```
kasperbirch1, line 5: Nice! I think this is more readable now.

kasperbirch1 (approved):
LGTM 👍

PR #1551, opened 2024-11-18
Update main Storybook dependency with other addons

kasperg:
#### Description

Update main Storybook dependency with other addons


JacobArrow (approved):
👍🏼

PR #1552, opened 2024-11-19
Add placeholder for game platform facet translation 

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFBRA-43

#### Description
This PR adds a placeholder for game platform facet text. We need the FBI API to publish their changes first, and we want these to not affect our app (crashing) in the meantime until we have the new specs, so this is a temporary solution to avoid errors until next week when we can update the autogenerated code from FBI API.

#### Screenshot of the result
-

-


Dresse (approved):
LGTM 👍

PR #1555, opened 2024-11-20
DDFBRA-183 - Anonymous and logged in user can try ebook

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFBRA-183
https://reload.atlassian.net/browse/DDFBRA-184
https://github.com/danskernesdigitalebibliotek/dpl-design-system/pull/791
https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/1779

#### Description

This pull request adds a new feature that allows users to try an e-book. It introduces a `Reader` app that loads teasers based on an `identifier`. This functionality is compatible with materials classified under the following types:  
- "e-bog"  
- "billedbog (online)"  
- "tegneserie (online)"  
- "årbog (online)"  

The `Reader` app is also designed to handle reserved materials by using an `orderId` instead of an `identifier`.  

Additional Changes:  
<s>- Added a full-screen modal for the reader. </s>
- Introduced a `MaterialSecondaryButton` and `MaterialSecondaryLink` to share behavior between `MaterialButtonsFindOnShelf`.

**Important Notes:**  
<s>During the implementation, I discovered a bug that requires the use of `window.location.reload();` when closing the `ReaderModal` to enable reopening it. Initially, I suspected this issue was introduced in this pull request. However, further investigation revealed that the same behavior is reproducible with the "FindOnShelf" feature on the live site ([https://bibliotek.kk.dk](https://bibliotek.kk.dk)).  

A separate ticket has been created to address this issue: [https://reload.atlassian.net/browse/DDFHER-140](https://reload.atlassian.net/browse/DDFHER-140).  </s>

The modal error is not related to why the Reader didn’t work. We experienced the same issue in the Go project. The problem seems to occur because Reader loads some scripts during onPageLoad. To address this, we restructured the architecture and made Reader an app instead of a component. This app now resides at /reader?* in the CMS.

**Update**
The code from [danskernesdigitalebibliotek/dpl-react#1562](https://github.com/danskernesdigitalebibliotek/dpl-react/pull/1562) was so similar that I merged that PR into this one.

This means that it is now possible to “Try eBook” using the Reader component and “Try Audiobook” using the Player.

Here is a matrial that has both ebook and audiobook:
https://varnish.pr-1779.dpl-cms.dplplat01.dpl.reload.dk/work/work-of:870970-basis:138374319?type=lydbog+%28online%29

#### Test

https://varnish.pr-1779.dpl-cms.dplplat01.dpl.reload.dk/work/work-of:870970-basis:62986115?type=tegneserie+%28online%29

https://github.com/user-attachments/assets/bb301ffd-4d2f-4b1c-8134-cc638dee021f


src/components/material/material-buttons/online/MaterialButtonOnlinePublizon.tsx
```diff
@@ -0,0 +1,65 @@
...
+            placeholders: { "@materialType": t("audiobookText") }
+          })}
+          size={size || "large"}
+          onClick={() => {
+            open(playerModalId(getManifestationIsbn(manifestations[0])));
+          }}
+          dataCy={`${dataCy}-player-teaser`}
+          ariaDescribedBy={t("onlineMaterialTeaserText")}
+        />
+      )}
```
Adamik10, line 60: Maybe we should add some sort of backup if `hasReaderManifestation` is false-y and `hasPlayerManifestation` is also false'y. 
Possibly just some text that says we can't show the material - like "missing material" or "missing data", or a disabled button that doesn't do anything, just has a text inside that convey this message.

src/components/material/material-buttons/online/MaterialButtonOnlinePublizon.tsx
```diff
@@ -0,0 +1,65 @@
...
+import { useText } from "../../../../core/utils/text";
+import { ButtonSize } from "../../../../core/utils/types/button";
+
+type MaterialButtonOnlinePublizonType = {
+  size?: ButtonSize;
+  manifestations: Manifestation[];
+  dataCy?: string;
+};
+
+const MaterialButtonOnlinePublizon: FC<MaterialButtonOnlinePublizonType> = ({
```
Adamik10, line 21: I would maybe reconsider the naming of this component. 
Maybe `MaterialButtonsOnlineInternal` 
- adding "s" at ..Buttons.. because there's going to be more of them inside (right? the loan + try)
- ..Internal.. serving as the opposite of `MaterialButtonOnlineExternal` that are all the online materials that lead outside of our app (note that that component is in singular because it only shows a single button)

If you really don't like that naming, then I can think of something like `MaterialButtonsOnlineEReol` or `MaterialButtonsOnlineViewable`

src/components/material/material-buttons/generic/MaterialSecondaryButton.tsx
```diff
@@ -0,0 +1,58 @@
...
+        size="large"
+        onClick={onClick}
+        dataCy={dataCy}
+        ariaDescribedBy={ariaDescribedBy}
+      />
+    );
+  }
+
+  return (
+    <button
```
Adamik10, line 44: Can we use our own `<Button />` component here?

src/components/material/player-modal/PlayerModal.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+  return (
+    <Modal
+      dataCy="player-modal"
+      // Todo: Find a better way to generate the modalId
+      modalId={playerModalId(identifier || orderId || "")}
+      screenReaderModalDescriptionText={t("playerModalDescriptionText")}
+      closeModalAriaLabelText={t("playerModalCloseButtonText")}
+    >
+      <Player identifier={identifier} />
```
Adamik10, line 18: Can we center the player inside modal pls?
image

src/components/reader-player/Player.stories.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+type Story = StoryObj<typeof Player>;
+
+export const WithIdentifier: Story = {
+  render: () => <Player identifier="9788702366600" />
+};
+
+// Works only if the matrial is reserved
+export const WithOrderId: Story = {
+  render: () => <Player orderId="" />
```
Adamik10, line 28: Could we pls somehow center it on the page so that it all doesn't get cut off in the story?
image

Adamik10 (changes requested):
Beautiful work :) I have a couple comments. 

Also I'm not sure if it's just happening to me, but if I stay on [this page (player on the demo site)](https://varnish.pr-1779.dpl-cms.dplplat01.dpl.reload.dk/work/work-of:870970-basis:138374319?type=lydbog+%28online%29&modal=player-modal-9788702393279) for long enough (3-10 mins) it throws an error (see picture) - 
image

src/apps/material/material.entry.tsx
```diff
@@ -11,6 +11,7 @@ import { GlobalEntryTextProps } from "../../core/storybook/globalTextArgs";
 interface MaterialEntryTextProps {
   alreadyReservedText: string;
   approveReservationText: string;
+  audiobookText: string;
```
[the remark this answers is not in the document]
kasperbirch1: They should all be here 🙏
https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/1779

src/components/Buttons/Button.tsx
```diff
@@ -43,6 +45,7 @@ export const Button: React.FC<ButtonProps> = ({
       disabled={disabled}
       onClick={onClick}
       id={id}
+      aria-describedby={ariaDescribedBy}
```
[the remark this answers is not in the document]
kasperbirch1: Skærmbillede 2024-12-03 kl  09 32 25

If it is 'undefined', it will not show 'aria-described by' at all. Therefore, it is important that we do not send an empty string, as the attribute would then be there without a value.

The aria-describedby attribute is typically used to provide additional context or description for a control beyond its label. For example: `findOnShelfExpandButtonExplanationText`

src/components/material/material-buttons/generic/MaterialSecondaryButton.tsx
kasperbirch1: Yes and no. 😅  This is used to create a button that appears like a link. 
image

We could add a prop for link styling, but since it wouldn't benefit other use cases, it might be better to avoid adding this to the `<Button />` component?

src/components/material/material-buttons/online/MaterialButtonOnlinePublizon.tsx
kasperbirch1: I like the idea. 😊 When we implement the rest of the logic for loans and reservations, we need to consider several things such as quotas, quantities, loans, etc. And here, I definitely think we should disable the button and display an explanatory text underneath. This logic will be implanted when we get there, and there are already tickets for it.

As it is right now, users will always be able to try the teaser for the materials that has the right type.

src/components/material/player-modal/PlayerModal.tsx
kasperbirch1: 👍

src/components/reader-player/Player.stories.tsx
kasperbirch1: Yes, that is not how it should look, but that’s not how it is on my storybook. 😅

Skærmbillede 2024-12-03 kl  11 13 09

Adamik10 (approved):
Yay :))

PR #1564, opened 2024-11-27
DDFBRA-229 - Test favorites

JacobArrow:
#### Link to issue

https://reload.atlassian.net/browse/DDFBRA-229

#### Description

While working with upgrades within the project, I discovered that we don't actually test the functionality of favorites.
This PR extends `favorites-list.test.ts` and `material.test.ts` to also test the functionality of favorites.

### DIFF
--- a/src/apps/favorites-list/favorites-list.test.ts
+++ b/src/apps/favorites-list/favorites-list.test.ts
@@ -52,7 +52,7 @@ describe("Favorites list", () => {
   });
 
   // Test that the list of favorites is displayed with materials correctly
-  it("Favorites list basics", () => {
+  it("Favorites list render as expected", () => {
     // Wait for element not in skeleton screen to prevent testing prematurely.
     cy.get(".cover").should("be.visible");
 
--- a/src/apps/favorites-list/favorites-list.test.ts
+++ b/src/apps/favorites-list/favorites-list.test.ts
@@ -6,6 +6,7 @@ describe("Favorites list", () => {
       win.sessionStorage.setItem(TOKEN_LIBRARY_KEY, "random-token");
     });
 
+    // To get the list of favorites
     cy.intercept("GET", "**list/default**", {
       statusCode: 200,
       body: {
@@ -16,12 +17,17 @@ describe("Favorites list", () => {
           "870970-basis:29630364"
         ]
       }
-    }).as("favorites");
+    }).as("Favorites");
 
-    // To fill the heart
-    cy.intercept("HEAD", "**list/default/work-of**", {
+    // To mark materials as favorites
+    cy.intercept("HEAD", "**list/default/**", {
       statusCode: 200
-    });
+    }).as("Favorite list service");
+
+    // To delete a material from favorites
+    cy.intercept("DELETE", "**list/default/**", {
+      statusCode: 200
+    }).as("Favorite list service");
 
     cy.interceptRest({
       aliasName: "Availability",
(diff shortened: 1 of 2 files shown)


src/apps/favorites-list/favorites-list.test.ts
```diff
@@ -45,31 +51,75 @@ describe("Favorites list", () => {
     cy.visit("/iframe.html?id=apps-favorite-list--primary&viewMode=story");
   });
 
+  // Test that the list of favorites is displayed with materials correctly
   it("Favorites list basics", () => {
```
Adamik10, line 55: These testings were built by ITK before they yeeted out of the project. 
The way we do it in the other tests (written by Reload) is that we make many small tests so that we can see in the Git log when a test doesn't pass - which one it was. 
I would really appreciate it if we divided the test into several many tests to that every it() tests one main thing. So when things fail, we see it in the log that for example "it("shows the header") test failed, or it("material links to correct material") test failed. It's so much easier when things aren't passing that way. It happened to me way too many times in the past that one of these monster tests didn't pass and then I had to actually run it locally to figure out what's wrong and look for it inside the it() - which part it is that is failing. 
Another downside with this approach (many test cases inside one it() ) is that the test doesn't run to the end if something fails in the beginning. If it's divided into many subtests we can precisely see which parts work and which don't. Even if it's the first test that fails - all the others will still run so we can see if those are failing too. 
So my recommendation is to take every // Comment ..and make these into separate it() cases. 🌻

src/apps/favorites-list/favorites-list.test.ts
```diff
@@ -45,31 +51,75 @@ describe("Favorites list", () => {
...
     // Wait for element not in skeleton screen to prevent testing prematurely.
     cy.get(".cover").should("be.visible");
 
-    // 2.a. Header "Favorites"
+    // Header has "Favorites"
```
Adamik10, line 59: So this would be something like `it("Shows the correct header")` 
Then we also avoid having comments in the code and these definitions will be printed out in Cypress.

src/apps/favorites-list/favorites-list.test.ts
```diff
@@ -45,31 +51,75 @@ describe("Favorites list", () => {
...
     cy.get(".content-list-page").find("h1").should("have.text", "Favorites");
-    // Number of materials on list
+
+    // Sub header shows correct number of materials
```
Adamik10, line 62: And this one would be something like `it("Shows the correct number of materials in sub-header")`

src/apps/material/material.test.ts
```diff
@@ -539,6 +539,34 @@ describe("Material", () => {
       .should("have.attr", "aria-pressed", "false");
   });
 
+  it("Can favorite a material", () => {
```
Adamik10, line 542: yeah like this one you introduced - it does a singular thing. Wonderful 🎉 So if it is failing we know that it's this one very precise thing that's causing trouble.

Adamik10 (changes requested):
I likey. Good job!
I just have a comment in regards to structuring that I think should be addressed before we merge this in 😉

src/apps/favorites-list/favorites-list.test.ts
JacobArrow: While I totally agree with your point, and that we generally should divide tests into something small and contained, I don't think that it make sense with this exact case. 

I have changed the wording of the test, as it was too vague, to actually tell what it did, and when or if it fails, we know its because the favorite list is not rendered as expected.

PR #1565, opened 2024-11-28
DDFBRA-236 - Logged in user can go from digital loans to material

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFBRA-236
https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/1814/

#### Description
This pull request replaces the link from the digital loan list to point to the material page instead of eReolen.

Unfortunately, I also found a bug in our code that causes an infinite re-render. This issue is also present on the develop branch, and I didn't find an easy solution to fix it. Therefore, this should be fixed in a separate ticket.

#### Test
If you have an online loan, you can view it under the “Digital Loans” section. The button, which was previously labeled “Go to eReolen” and linked to eReolen, has been updated to now correctly link to the work with the appropriate type selected.

https://varnish.pr-1814.dpl-cms.dplplat01.dpl.reload.dk/user/me/loans

https://github.com/user-attachments/assets/8e1940fc-f05a-476f-a561-598f46e6a4f4

### DIFF
--- a/src/apps/loan-list/modal/material-details-modal.tsx
+++ b/src/apps/loan-list/modal/material-details-modal.tsx
@@ -1,3 +1,4 @@
+//  Todo: Fix infinite loop in MaterialDetailsModal
 import React, { FC, ReactNode } from "react";
 import Modal from "../../../core/utils/modal";
 import { useText } from "../../../core/utils/text";
@@ -33,6 +34,7 @@ const MaterialDetailsModal: FC<MaterialDetailsModalProps> = ({
   children
 }) => {
   const t = useText();
+  // console.log("Rendering MaterialDetailsModal:");
 
   if (!modalId) {
     return null;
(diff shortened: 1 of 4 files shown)


src/apps/loan-list/modal/material-details-modal.tsx
```diff
@@ -33,6 +34,7 @@ const MaterialDetailsModal: FC<MaterialDetailsModalProps> = ({
   children
 }) => {
   const t = useText();
+  // console.log("Rendering MaterialDetailsModal:");
```
Adamik10, line 37: Maybe we should remove this line?

src/apps/loan-list/modal/material-details-modal.tsx
```diff
@@ -1,3 +1,4 @@
+//  Todo: Fix infinite loop in MaterialDetailsModal
```
Adamik10, line 1: Bahahaha. How fun. 
Try to describe it better what needs to be fixed so that people who don't know about the bug at all would be able to reproduce it 😉

Adamik10 (approved):
Beautiful job. We clap and cheer. 
I have 2 little comments for you to consider 🙏 but otherwise approved 😉

PR #1572, opened 2024-12-05
Update FBI graphql with gamePlatform

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFBRA-43

#### Description
This PR updates autogenerated FBI API code to be aligned to the newest API specs. Among other things, gamePlatform is added as one of the facetFieldEnum values. 

#### Screenshot of the result
-

-

### DIFF
--- a/src/components/facet-browser/helper.ts
+++ b/src/components/facet-browser/helper.ts
@@ -136,7 +136,7 @@ export const getFacetFieldTranslation = (name: FacetFieldEnum) => {
     case FacetFieldEnum.Year.toLowerCase():
       return "facetYearText";
     // TODO: Replace this case when FBI API specs have updated
-    case "gameplatform":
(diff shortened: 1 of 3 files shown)


src/components/facet-browser/helper.ts
```diff
@@ -136,7 +136,7 @@ export const getFacetFieldTranslation = (name: FacetFieldEnum) => {
     case FacetFieldEnum.Year.toLowerCase():
       return "facetYearText";
     // TODO: Replace this case when FBI API specs have updated
```
kasperbirch1, line 138: I think this should be removed now? 😊

kasperbirch1 (approved):
LGTM 👍

src/components/facet-browser/helper.ts
Adamik10: Indeed :)

PR #1597, opened 2024-12-13
DDFBRA-283 - Ensure `v1/products/` is never called without an identifier

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFBRA-283

#### Description

We have identified a recurring issue in the code where attempts are made to access the first element of an empty array (`array[0]`), which results in `undefined`.

This issue occurs across all articles where the identifier is an empty array.

Additionally, our TypeScript implementation requires the `useGetV1ProductsIdentifier` hook to receive a string. To satisfy TypeScript, we currently use `|| ""`, even though the hook is only activated when a value is present. However, this approach has inadvertently allowed errors to occur, and we have been uncertain about how to address this properly.

This change ensures that the `|| ""` fallback in `useGetV1ProductsIdentifier` is never activated, effectively preventing the error from occurring.

#### Test

https://varnish.pr-1861.dpl-cms.dplplat01.dpl.reload.dk/search?q=hest

Select "articles" in the facet and make sure no drafts are created for `v1/products/`

### DIFF
--- a/src/apps/loan-list/materials/utils/digital-material-fetch-hoc.tsx
+++ b/src/apps/loan-list/materials/utils/digital-material-fetch-hoc.tsx
@@ -35,7 +35,13 @@ const fetchDigitalMaterial =
         data: productsData,
         isSuccess: isSuccessDigital,
         isLoading
-      } = useGetV1ProductsIdentifier(item.identifier);
+      } = useGetV1ProductsIdentifier(item.identifier, {
+        query: {
+          // We never want to pass an empty string to the API
+          // So we only enable the query if we have an isbn
+          enabled: !!item.identifier
+        }
+      });
 
       useEffect(() => {
         if (productsData && isSuccessDigital && productsData.product) {
--- a/src/components/availability-label/availability-label.tsx
+++ b/src/components/availability-label/availability-label.tsx
@@ -1,4 +1,5 @@
 import React, { memo } from "react";
+import { first } from "lodash";
 import { useDeepCompareEffect } from "react-use";
 import { useText } from "../../core/utils/text";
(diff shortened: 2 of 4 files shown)


src/apps/loan-list/materials/utils/digital-material-fetch-hoc.tsx
```diff
@@ -35,7 +35,13 @@ const fetchDigitalMaterial =
...
         isSuccess: isSuccessDigital,
         isLoading
-      } = useGetV1ProductsIdentifier(item.identifier);
+      } = useGetV1ProductsIdentifier(item.identifier, {
+        query: {
+          // We never want to pass an empty string to the API
+          // So we only enable the query if we have an isbn
+          enabled: !!item.identifier
+        }
+      });
```
kasperg, line 44: This check may be unnecessary given the check on L30 but I guess it is nice to have around given the problems caused if we allow an empty string to pass.

kasperg (approved):
:+1: LGTM

I would appreciate if you get someone else to OK it as well.

src/components/availability-label/availability-label.tsx
```diff
@@ -45,7 +46,7 @@ export const AvailabilityLabel: React.FC<AvailabilityLabelProps> = ({
     accessTypes,
     access,
     faustIds,
-    isbn: isbns ? isbns[0] : null,
+    isbn: first(isbns) ?? null,
```
Adamik10, line 49: I would maybe consider using the `||` operator here to make sure we catch all possible false-y values

Adamik10 (approved):
I have one consideration :)

PR #1633, opened 2025-01-08
Prevent double click on links (logout bug)

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFBRA-331

#### Description
This PR adds loading state to Link component to prevent multiple clicks. Since the link is often used as a button in the interface, we needed a way to prevent multiple clicks on it before the action is fully carried out.

#### Screenshot of the result
-

-


ThomasGross (changes requested):
Since this change is only use for this exact case, I don't think we should implement this change to every link on the site. 

Consider creating another component or having this click lock as a conditional option passed down from props. :)

ThomasGross (changes requested):
I think this approach is better but I would like preserve the standard link behavior. Render a component with the href on the anchor tag and nothing else unless canOnlyBeClickedOnce is set to true. 

In the code right now there's still a lot of hijacking of the default link behavior with the async function and overwriting the click and keyup events. 😊

Adamik10:
It's not something that I introduced in this PR. And also it does serve a specific purpose for some use cases as we've seen going through the code. I stay firm in this solution :))

PR #1635, opened 2025-01-09
Update graphql schema

JacobArrow:
#### Description

The gateway schema has been updated by DBC, which require us to handle a new `traceId` field for facets. This PR updates our generated graphql schema and queries to use the new field.


kasperg (approved):
:+1: LGTM

PR #1643, opened 2025-01-13
Type safety for Cypress intercepts

JacobArrow:
#### Link to issue

As a part of my work for GO!
[DDFBRA-234](https://reload.atlassian.net/browse/DDFBRA-234)

#### Description

Our Cypress tests rely on intercepting Graphql post requests to simulate real actions or data through fixtures.
As a part of that we have created a custom Cypress command which receives the operation name of the request we want to intercept. This PR builds on top of this behavior and adds type safety to the operation name through a codegen plugin which generates an object of all available operations, which then can be made into a union of strings.


dbc-gateway.codegen.yml
```diff
@@ -5,7 +5,7 @@ schema:
         Authorization: "Bearer ${STORYBOOK_LIBRARY_TOKEN}"
 documents: "src/**/*.graphql"
 generates:
-  src/core/dbc-gateway/generated/graphql.tsx:
+  src/core/dbc-gateway/generated/graphql.ts:
```
spaceo, line 8: I am not sure it was a typo.
I think it was because the generated file contains hooks.
But if it doesn't matter that a hook-containing file is called `.ts` then so be it.

JacobArrow: According to the React docs, hooks should use the `.js` or `.ts` file extension
https://react.dev/learn/reusing-logic-with-custom-hooks

src/core/fetchers/types.ts
```diff
@@ -29,3 +31,7 @@ export type RequestCallback = (
 export type ErrorType<TErrorData> = TErrorData;
 
 export type BodyType<TBodyData> = TBodyData;
+
+type QueryOperations = keyof typeof operationNames.Query;
+type MutationOperations = keyof typeof operationNames.Mutation;
+export type Operations = QueryOperations | MutationOperations;
```
kasperg, line 37: Please consider your placement of these types.

This is a generic file with types for fetchers used across multiple services. Now you have added a new type which is only relevant to a specific service. You also use very specific names `QueryOperations`, `MutationOperations` and `Operations`.

I would prefer is this was kept within the `dbc-gateway` (which is the service in question) or Cypress tests (where the types are used).

If there is a larger plan with this - e.g. that this should include all operations for all GraphQL services then I would expect it to be mentioned in a code comment or the like.

PR #1714, opened 2025-02-27
Handle TS error differently

JacobArrow:
#### Link to issue

https://reload.atlassian.net/browse/DDFBRA-407

#### Description

Replace `@ts-ignore` with `@ts-expect-error`

This way TypeScript will let us know when a fix is made in the `react-use` package, and that we can safely remove the Directive Comments, which saves the need for a TODO in the code.


Adamik10 (approved):
👏

PR #1725, opened 2025-03-03
Switch to Node 20 Iron LTS instead of specific version

kasperg:
#### Description

`lts/iron` is the latest released version of Node 20 which has long term support.

This should ensure that we get the most recent version of the major release we are currently using. Rolling updates within this should be safe to do as long a we stay within the major version. We expect any errors to be caught build time.

At this point we need Node 20.10 or newer for babel-loader 10 in #1720


kasperbirch1 (approved):
LGTM 👍

PR #1732, opened 2025-03-07
DDFNEXT-752 - Fix `2025.10.0`

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFNEXT-752

#### Description

Refactor `MaterialSearchList` to use work titles directly instead of a helper function.

The work data comes from `useComplexSearchWithPaginationQuery` and uses `WorkSmall`, which does not contain `mainLanguages` and `materialTypes`, both of which are required for `getWorkTitle`.


Dresse (approved):
LGTM 👍

kasperg (approved):
:+1: Appreciate the updates here.

PR #1739, opened 2025-03-11
Never render an inactive access url

JacobArrow:
#### Link to issue

https://reload.atlassian.net/browse/DDFBRA-502

#### Description

This PR refactors how we get the correct access element before rendering the external manifestation link button. By making sure that we find the access element, which has the status as "OK", an inactive link will not get rendered.


cypress/fixtures/material-buttons/material-buttons-movie-fbi-api.json
```diff
@@ -521,7 +521,8 @@
                 "__typename": "AccessUrl",
                 "origin": "www.filmstriben.dk",
                 "url": "http://www.filmstriben.dk/fjernleje/film/details.aspx?id=9000004315",
-                "loginRequired": false
+                "loginRequired": false,
```
Adamik10, line 524: wonderful job with the updated fixtures too!

Adamik10 (approved):
Brava 👏 BRAVA 👏

PR #1742, opened 2025-03-12
DDFSAL-34 - Add material type to reservation tracking

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFSAL-34
https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/2168

#### Description
Updated reservation tracking to include material type. 

This change ensures that libraries have the necessary data for proper tracking, as workId alone was not sufficient.

I went down a bit of a rabbit hole by not using [0] on arrays, as Copilot first commented, since it may be undefined. This led to some changes that are not ticket-relevant but improve the current code. My Git commit should be descriptive of why I made the changes.

I have chosen to ignore the last Copilot suggestions, as I don’t think they are relevant.

#### Test
https://varnish.pr-2168.dpl-cms.dplplat01.dpl.reload.dk/

### DIFF
--- a/src/apps/material/helper.ts
+++ b/src/apps/material/helper.ts
@@ -26,7 +26,7 @@ import {
 import { UseConfigFunction } from "../../core/utils/config";
 import {
   flattenCreators,
-  getManifestationType,
+  getMaterialType,
   orderManifestationsByYear
 } from "../../core/utils/helpers/general";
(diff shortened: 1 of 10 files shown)


kasperg (changes requested):
:+1: I appreciate the intention here but I think the code structure here needs a second look.

kasperg (approved):
:+1: LGTM

PR #1758, opened 2025-03-24
DDFSAL-36 - Track click to open reservation modal for specific manifestation

kasperbirch1:
#### Links
https://reload.atlassian.net/browse/DDFSAL-36
https://ui.lagoon.dplplat01.dpl.reload.dk/projects/dpl-cms/dpl-cms-pr-2186
https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/2186

#### Description
This tracks when the modal is opened for a specific manifestation, not the actual reservation itself, which we already track with `statistics.reservation`.

#### Test
https://varnish.pr-2186.dpl-cms.dplplat01.dpl.reload.dk/

### DIFF
--- a/src/components/material/MaterialMainfestationItem.tsx
+++ b/src/components/material/MaterialMainfestationItem.tsx
@@ -176,6 +176,7 @@ const MaterialMainfestationItem: FC<MaterialMainfestationItemProps> = ({
           size="small"
           workId={workId}
           materialTitleId={mainfestationTitleId}
+          isSpecificManifestation
         />
       </div>
     </div>
--- a/src/components/material/material-buttons/physical/MaterialButtonPhysical.tsx
+++ b/src/components/material/material-buttons/physical/MaterialButtonPhysical.tsx
@@ -37,8 +37,8 @@ const MaterialButtonPhysical: FC<MaterialButtonPhysicalProps> = ({
   const onClick = () => {
     if (isSpecificManifestation && first(pids)) {
       track("click", {
-        id: statistics.specificManifestation.id,
-        name: statistics.specificManifestation.name,
+        id: statistics.reserveSpecificManifestation.id,
+        name: statistics.reserveSpecificManifestation.name,
         trackedData: `${first(pids)}`
       });
     }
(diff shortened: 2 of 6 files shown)


src/components/material/material-buttons/physical/MaterialButtonPhysical.tsx
```diff
@@ -3,30 +3,45 @@ import { useModalButtonHandler } from "../../../../core/utils/modal";
...
+  pids: Pid[];
 }
 
 const MaterialButtonPhysical: FC<MaterialButtonPhysicalProps> = ({
   manifestationMaterialType,
   faustIds,
   size,
-  dataCy = "material-button-physical"
+  dataCy = "material-button-physical",
+  isSpecificManifestation,
```
kasperg, line 27: I wondered if we could avoid adding this prop by checking for the length of the pids. If we only pass 1 then we have a specific manifestation but I understand that this does not entirely cover the expected behavior.

src/components/material/MaterialMainfestationItem.tsx
```diff
@@ -176,6 +176,7 @@ const MaterialMainfestationItem: FC<MaterialMainfestationItemProps> = ({
           size="small"
           workId={workId}
           materialTitleId={mainfestationTitleId}
+          isSpecificManifestation
```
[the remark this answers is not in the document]
kasperbirch1: I'm not sure that I fully understand what you mean. The reason this prop was added is that `<MaterialButtons />` is a reusable component. Adding the `isSpecificManifestation` prop allows us to indicate that is a specific manifestation within the `MaterialButtons` component.

In JSX, writing a prop without an explicit value (e.g., `isSpecificManifestation`) is shorthand for `isSpecificManifestation={true}`.

src/components/material/material-buttons/physical/MaterialButtonPhysical.tsx
kasperbirch1: Good point — let's check with Tue on that.

src/components/material/MaterialMainfestationItem.tsx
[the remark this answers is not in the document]
kasperg: > In JSX, writing a prop without an explicit value (e.g., isSpecificManifestation) is shorthand for isSpecificManifestation={true}.

That was the point is was missing. Thanks for pointing this out.

src/components/material/material-buttons/physical/MaterialButtonPhysical.tsx
kasperg: I do not think this is necessary.

If we checked on the length of the array we would not be able to distinguish between users clicking on the "main" buttons at the top where there is only 1 manifestation for the selected type and users clicking on single manifestations in the details. I expect that is what they want.

kasperg (approved):
:+1: LGTM

PR #1759, opened 2025-03-24
Upgrade codecov action to version 5

JacobArrow:
#### Link to issue

https://reload.atlassian.net/browse/DDFBRA-323

#### Description

This PR upgrades the usage of codecov in our Github action to version 5.
While upgrading I discovered that it is now required to use an access token, which has been added to the repo.
I also discovered that the action was running inside a limited cypress container, which did not have the required packages available. As the cypress container was added to address a Chrome bug, that has since been fixed, it could be safely removed.


Adamik10 (approved):
🎉

PR #1788, opened 2025-04-11
DDFSAL 3

LasseStaus:
#### Link to issue

[Jira ticket](https://reload.atlassian.net/browse/DDFSAL-3)

#### Description

This pr does the following: 

- Creates a MaterialGrid with Related work recommendations on a material-page, as described in the ticket. This is a component, and is applied to the `material.tsx` page.

- Reorganize the structure of MaterialGrid (Automatic/manual), RecommendedMaterial & related helpers/utils. Before,  these were all apps, and were using other apps. Now they should be seperated into components being used by the apps. 

- Adjust various testing cases: 

There is an connection between `src/core/utils/helpers/lazy-load.ts` & the cypress tests related to the material page. As a result of adding component and making the page longer, lazy-loading is not triggered & many tests fail. 

The quick fix applied to solve this & other content-not-in-view related tests was to increase the duration of the scroll used in cypress. 

More elegant suggestions are very welcome. 

#### Screenshot of the result

image

Test by browsing various material pages & verify that they are displayed 8 recommended materials at the bottom. 

For testing, refer to the CMS link in [this pr](https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/2283)

Link directly to pr: https://varnish.pr-2283.dpl-cms.dplplat01.dpl.reload.dk/work/work-of:870970-basis:140259411?type=bog


src/apps/material/material.tsx
```diff
@@ -160,112 +161,119 @@ const Material: React.FC<MaterialProps> = ({ wid }) => {
...
+                mainIconPath={CreateIcon}
+              />
+            }
+          >
+            <MaterialReviews pids={hasReview.map((review) => review.pid)} />
+          </DisclosureControllable>
+        )}
+      </section>
+      {work && (
+        <section>
```
kasperbirch1, line 272: Is this the HTML structure we want? I think I might have expected it to just be in inside the "main" section as the others?

Skærmbillede 2025-04-25 kl  14 33 59

src/apps/material-grid/automatic/MaterialGridAutomatic.tsx
```diff
@@ -20,7 +20,10 @@ const MaterialGridAutomatic: React.FC<MaterialGridAutomaticProps> = ({
   description,
   selectedAmountOfMaterialsForDisplay
 }) => {
+  const t = useText();
+  const buttonText = t("buttonText");
```
kasperbirch1, line 24: Is there a reason why we don't just do this buttonText={t("buttonText")}?

src/components/material-grid-related/MaterialGridRelated.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+export type MaterialGridRelatedProps = {
+  work: Work;
+};
+
+const MaterialGridRelated: React.FC<MaterialGridRelatedProps> = ({ work }) => {
+  const t = useText();
+  const { pid } = work.manifestations.bestRepresentation;
+
+  const title = t("materialGridRelatedTitleText");
```
kasperbirch1, line 20: Same here 🤓

src/components/material-grid-related/MaterialGridRelated.tsx
```diff
@@ -0,0 +1,48 @@
...
+} from "../material-grid/MaterialGrid";
+import MaterialGridSkeleton from "../material-grid/MaterialGridSkeleton";
+
+export type MaterialGridRelatedProps = {
+  work: Work;
+};
+
+const MaterialGridRelated: React.FC<MaterialGridRelatedProps> = ({ work }) => {
+  const t = useText();
+  const { pid } = work.manifestations.bestRepresentation;
```
kasperbirch1, line 18: Place use `getWorkPid` from src/core/utils/helpers/general.ts

kasperbirch1 (approved):
Great work! I’ve left a few minor comments you might want to consider, but overall it’s good to go. 👍

src/apps/material-grid/automatic/MaterialGridAutomatic.tsx
LasseStaus: I see what you mean! 
I just like putting things in a const first so the props look a bit cleaner and easier to read, with no function calls inside the prop passing.

src/apps/material/material.tsx
LasseStaus: Yep i think so! I'd like it be in another section to separate it from the "displayed" material content. I see this as a new section, containing recommendations 👍

src/apps/material-grid/link-automatic/MaterialGridLinkAutomatic.stories.tsx
```diff
@@ -79,7 +79,7 @@ export const Primary: Story = {
     title: "Recommended materials",
     description:
       "This is a long description of the materials selected, or whatever else you want to put in here",
-    link: "https://dapple-cms.docker/advanced-search?sublocation=fantasy&advancedSearchCql=%27harry%27&linked=true",
+    link: "https://dapple-cms.docker/advanced-search?advancedSearchCql=%27harry%27&linked=true",
```
kasperbirch1, line 82: Maybe we should write a comment or a ticket to update the data so we have something that uses location and sublocation.

kasperbirch1 (approved):
LGTM 👍

PR #1806, opened 2025-04-25
Update FBS API: Patron data retrieval

kasperg:
#### Link to issue

https://reload.atlassian.net/browse/DDFSAL-97

#### Description

This is the initial work at updating our usage of FBS API to stop using [deprecated API versions that will be removed soon](https://ciceroconnect.zendesk.com/hc/da/articles/11396856511517-%C3%86ndringer-til-CMS-API-december-2024) and switch to some that will remain.

This first step updates the project to stop using `/external/agencyid/patrons/patronid/v2` for retrieval of patron data and use `/external/agencyid/patrons/person/patronid/v2` instead.

Patron data retrieval is handled by our own hook `usePatronData()`. To preserve compatibility in the remaining codebase we transform the response from the new API to that of the old. The key changes in our regard deal with handling of email addresses and phone numbers.

Tests are updated accordingly.

This should be the first of a number of changes to address the situation.

I recommend going through the changes commit by commit.


src/core/fbs/fbs.ts
```diff
@@ -862,7 +863,8 @@ export type GetHoldingsLogisticsV1QueryError = ErrorType<void>;
 /**
  * @summary Get placement holdings for bibliographical records.
  */
-export const useGetHoldingsLogisticsV1 = <
+
+export function useGetHoldingsLogisticsV1<
```
kasperbirch1, line 867: Just curious — is there a reason this was changed from an arrow function?

src/core/utils/helpers/usePatronData.ts
```diff
@@ -1,13 +1,42 @@
...
+        ? {
+            // As we use the FBS adapter where we are only able to retrieve
+            // the patron matching the current authentication token it seems
+            // safe to assume that authentication is valid if we can retrieve
+            // data.
+            authenticateStatus: "VALID",
+            patron: convertPersonPatronV2ToPatronV5(data)
+          }
+        : null;
+    }
```
kasperbirch1, line 34: Clever idea! Well done. 👍

kasperbirch1 (approved):
Really great work! I found two places that I think should also be updated. I’ve made the changes in a separate pull request to avoid messing with your Git history. Please take a look at that pull request and feel free to merge it if you agree.

https://github.com/danskernesdigitalebibliotek/dpl-react/pull/1807

src/core/fbs/fbs.ts
kasperg: Not as far as I know. This is generated file so I have no personal stake in this change.

kasperg:
Closed in favour of #1818

PR #1821, opened 2025-04-30
DDFSAL 1 DDFSAL 4

LasseStaus:
- **Add `material.tsx` texts that are used in `MaterialGridRelated.tsx`**
- **Update `MaterialGridRelated` to have filters for switching displayed materials.**
- **Add select filters & inline filters for `MaterialGridRelated`**

#### Link to issue

#### Description

#### Screenshot of the result


src/components/material-grid-related/MaterialGridRelated.tsx
```diff
@@ -19,26 +39,123 @@ const MaterialGridRelated: React.FC<MaterialGridRelatedProps> = ({ work }) => {
...
+    return opts;
+  }, [
+    allQueriesLoaded,
+    recommendationMaterials,
+    seriesMaterials,
+    authorMaterials,
+    t
+  ]);
+
+  React.useEffect(() => {
```
kasperbirch1, line 120: Not a huge deal, but I noticed that we usually deconstruct useEffect from the React import elsewhere. It might be good to stick with the same pattern for consistency.

src/components/material-grid-related/MaterialGridRelated.tsx
```diff
@@ -19,26 +39,123 @@ const MaterialGridRelated: React.FC<MaterialGridRelatedProps> = ({ work }) => {
...
+  const { data: seriesData, isLoading: seriesLoading } =
+    useComplexSearchWithPaginationQuery({
+      cql: `term.series='${seriesObject?.title}'`,
+      limit: 8,
+      offset: 0,
+      filters: {}
+    });
+
+  const [filter, setFilter] =
+    React.useState<MaterialGridFilterType>("recommendation");
```
kasperbirch1, line 72: Same as I wrote about useEffect. ☺️

kasperbirch1 (approved):
Great work 👍  this is a super useful feature! I’ve tested it across multiple materials, and it seems to be working perfectly.

Regarding rendering an empty <section> when there are no results — I think that’s not the right approach.

src/components/material-grid-related/MaterialGridRelated.tsx
```diff
@@ -19,26 +39,123 @@ const MaterialGridRelated: React.FC<MaterialGridRelatedProps> = ({ work }) => {
...
+    useComplexSearchWithPaginationQuery({
+      cql: creatorCqlString,
+      limit: 8,
+      offset: 0,
+      filters: {}
+    });
+
+  const { data: seriesData, isLoading: seriesLoading } =
+    useComplexSearchWithPaginationQuery({
+      cql: `term.series='${seriesObject?.title}'`,
```
kasperg, line 65: Consider whether we should disable (`enabled: false`) this hook if there is no series for the object.

It seems to me like we would get a `term.series=''` query but that does not make much sense to me.

src/components/material-grid-related/MaterialGridRelatedSelect.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+export function MaterialGridRelatedSelect({
+  filter,
+  onChange,
+  options
+}: MaterialGridRelatedSelectProps) {
+  const t = useText();
+
+  return (
+    <div className="dropdown dropdown--grey-borders input-with-dropdown__dropdown material-grid-related__dropdown">
```
kasperg, line 23: Please consider whether we can use the existing [`Dropdown`](https://github.com/danskernesdigitalebibliotek/dpl-react/blob/3b69c39a82e283ff5722168ccb6761cf559a5ab9/src/components/Dropdown/Dropdown.tsx) component here.

src/components/material-grid-related/helper.ts
```diff
@@ -0,0 +1,56 @@
...
+}
+
+export function extractAuthorMaterials(
+  data?: ComplexSearchWithPaginationQuery
+): MaterialGridItemProps[] {
+  if (!data?.complexSearch?.works) return [];
+  return data.complexSearch.works.map((work) => ({
+    wid: work.workId as WorkId
+  }));
+}
```
kasperg, line 45: Consider if both these functions are necessary.

They look exactly the same to me - apart from the name.

src/components/material-grid-related/MaterialGridRelated.tsx
```diff
@@ -19,26 +39,123 @@ const MaterialGridRelated: React.FC<MaterialGridRelatedProps> = ({ work }) => {
...
+  const { data: recommendationData, isLoading: recommendationLoading } =
+    useWorkRecommendationsQuery({
+      pid,
+      limit: 8
+    });
+
+  const { data: creatorData, isLoading: creatorLoading } =
+    useComplexSearchWithPaginationQuery({
+      cql: creatorCqlString,
+      limit: 8,
```
kasperg, line 58: This effectively limits the number of retrieved items to a maximum of 8 - which is the same number as the material grid will display without expanding. This is convenient but I wonder if it is as intended by the business.

Another approach would be to retrieve *all* the corresponding work (ids) and then let the component expand if requested by the user. This would make it possible to present extensive authorship or long running series.

This would likely also hand warranted a new query where we only retrieve a hitcount and work ids.

Oh well. This works.

src/components/material-grid-related/helper.ts
```diff
@@ -0,0 +1,56 @@
...
+} from "../../core/dbc-gateway/generated/graphql";
+import { WorkId } from "../../core/utils/types/ids";
+import { MaterialGridItemProps } from "../material-grid/MaterialGrid";
+import {
+  MaterialGridFilterOption,
+  MaterialGridFilterType
+} from "./MaterialGridRelated.types";
+
+export function prepareCreatorCql(
+  creators: string[] | undefined | null
```
kasperg, line 13: Consider if this argument type is appropriate.

I wonder why this function would need to accept `undefined | null`.

src/components/reservation/reservation.test.ts
```diff
@@ -11,6 +11,11 @@ describe("Reservation", () => {
       operationName: "WorkRecommendations",
       fixtureFilePath: "material/material-grid-related-recommendations.json"
     });
+    cy.interceptGraphql({
+      operationName: "complexSearchWithPagination",
+      fixtureFilePath:
+        "material/material-grid-related-author-recommendations.json"
+    });
```
kasperg, line 18: :+1: Consider this will also require fixtures to be added in this CMS.

kasperg (approved):
:+1: Nice work. I have a couple of things for you to consider before this is merged.

src/components/material-grid-related/helper.ts
LasseStaus: Yeah thats not necessary. Have replaced with a `extractMaterialsFromComplexSearch` instead that serves both.

src/apps/material/material.test.ts
```diff
@@ -578,6 +578,39 @@ describe("Material", () => {
...
+
+    cy.contains(
+      ".material-grid-related__filter-button",
+      "By same author"
+    ).click();
+
+    cy.contains(
+      ".material-grid-related__filter-button",
+      "In same series"
+    ).click();
```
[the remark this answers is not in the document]
LasseStaus: Changed the data-cy, good catch. 

Regarding the intercept it's a good idea, but i am actually running all 3 queries beforehand, in other to get the numbers/count displayed next to each filter type, så there is no new query after clicking / selecting a filter. 

I assume that was the test case right?

src/components/material-grid-related/helper.ts
LasseStaus: 👍 I changed this helper

src/components/reservation/reservation.test.ts
LasseStaus: Tests on the CMS are green so i would assume it is not necessary

PR #1823, opened 2025-05-02
Prefer active DBC Webarkiv link for online materials

JacobArrow:
#### Link to issue

https://reload.atlassian.net/browse/DDFBRA-502

#### Description

Return the first active DBC Webarkiv element, but fallback to any other active element, if one from DBC is not found.


Adamik10 (approved):
👏

PR #1863, opened 2025-05-22
DDFSAL 80 use covers from fbi

LasseStaus:
#### Link to issue

[DDFSAL-80](https://reload.atlassian.net/browse/DDFSAL-80)

#### Link to PR test 
https://varnish.pr-2548.dpl-cms.dplplat01.dpl.reload.dk/

#### Description

This PR refactors the retrieval of covers away from coverservice and into FBI. 

It was decided to continue using the approach of having a separate query for fetching covers inside the cover component, as opposed to refactoring every work/manifestation query and retrieve covers as part of these queries. 

Based on this, the changes the in the pr include: 

1. Rewriting the cover query to fetch from FBI. 
Covers can either be retrieved by the use of PIDS, or isbn. If isbn is used, the pid from the returned bestrepresentation will be used. 

2. Update helper functions & types for managing the change and move away from the types from coverservice

3. Update all tests that failed as a result of these changes. 

All tests that failed have been fixed so they intercept the new graphql queries. It is very possible there are tests or test cases in other files that display a cover in their test, that is now not present. I did not manually go through every visual to account for covers in tests, that were not testing the presence of covers in their test cases. 

### Comments

Something i didnt do, was to consider if some sizes of images should be changed directly. Should an old xlarge size now be large? Should a small be a xsmall?
Please consider this and lets talk about it. For now i just let the sizes be the sizes (in terms of props) but added a displaysize where necessary to keep the same sizes. 

[DDFSAL-80]: https://reload.atlassian.net/browse/DDFSAL-80?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/apps/reservation-list/list/reservation-list.test.ts
```diff
@@ -304,11 +306,14 @@ describe("Reservation list", () => {
...
-      })
-      .as("cover");
+    cy.interceptGraphql({
+      operationName: "GetCoversByPids",
+      fixtureFilePath: "cover.json"
+    });
+    cy.interceptGraphql({
+      operationName: "GetBestRepresentationPidByIsbn",
+      fixtureFilePath: "cover-get-best-representation-by-isbn.json"
+    });
```
kasperg, line 316: :+1: We need to add this extra intercept here as we have a reservation which might not refer to a pid.

src/apps/reservation-list/list/reservation-list.test.ts
```diff
@@ -885,10 +885,8 @@ describe("Reservation list", () => {
       "/iframe.html?path=/story/apps-reservation-list--reservation-list-entry"
     );
 
-    cy.interceptRest({
-      aliasName: "work-bestrepresentation",
-      httpMethod: "POST",
-      url: "**/next*/**",
+    cy.interceptGraphql({
+      operationName: "getManifestationViaBestRepresentationByFaust",
       fixtureFilePath: "reservation-list/work-bestrepresentation.json"
```
kasperg, line 890: Nice to have this fixed up. Before we intercept rest for a GraphQL call using a Rest interception. That it not right.

src/apps/reservation-list/list/reservation-list.test.ts
```diff
@@ -910,6 +908,10 @@ describe("Reservation list", () => {
       .find(".modal-details__title")
       // Details should also contain the best representation title.
       .should("contain", "Best representation of dummy title");
+    cy.getBySel("modal-details__header")
+      .find("img")
+      .should("have.attr", "src")
+      .and("match", FbiCoverUrlPattern);
```
kasperg, line 914: :+1: Nice.

src/apps/reservation-list/list/reservation-list.test.ts
```diff
@@ -927,6 +929,7 @@ describe("Reservation list", () => {
             pickupDeadline: null,
             expiryDate: "2022-09-21",
             dateOfReservation: "2022-06-14T09:00:50.059",
+            //  identifier: "9781844135417", //
```
kasperg, line 932: Please do not add commented out code in VCS unless you provide a comment explaining why it is relevant.

src/apps/reservation-list/modal/reservation-details/reservation-details.test.ts
```diff
@@ -394,19 +401,13 @@ describe("Reservation details modal", () => {
     cy.visit(
       "/iframe.html?path=/story/apps-reservation-list--reservation-list-physical-details-modal"
     );
-    // These two cy.wait are split up because of the proceedance of the requests
-    cy.wait(["@physical_reservations", "@digital_reservations"]);
-    cy.wait(["@work"]);
```
kasperg, line 399: :+1: for getting rid of some of these `wait()` calls.

src/apps/reservation-list/modal/reservation-details/reservation-details.test.ts
```diff
@@ -613,6 +614,10 @@ describe("Reservation details modal", () => {
       fixtureFilePath: "reservation-list/work.json"
     });
 
+    cy.interceptGraphql({
+      operationName: "GetCoversByPids",
+      fixtureFilePath: "cover.json"
+    });
```
kasperg, line 620: Please consider if/why we need to add this extra intercept.

It seems as if we already intercepted the same operation in the `beforeEach()`. If we need to change something then I would suggest that we change an `cy.interceptRest()` to a `cy.interceptGraphql()` with the correct operation name.

This also applies elsewhere in this file.

src/components/autosuggest-material/autosuggest-material.tsx
```diff
@@ -75,7 +75,7 @@ const AutosuggestMaterial: React.FC<AutosuggestMaterialProps> = ({
             <div className="autosuggest__material-card">
               <Cover
                 animate
-                size="xsmall"
+                size="xSmall"
```
kasperg, line 78: 😁  Casing matters.

src/components/cover/cover.types.ts
```diff
@@ -0,0 +1,48 @@
+import { Cover } from "../../core/dbc-gateway/generated/graphql";
+
+/**
+ * The keys in the GraphQL `Cover` object that refer to the actual image sizes.
+ * Present in the data.
+ */
+export type FbiCoverImageSizeKey = Extract<
+  keyof Cover,
+  "xSmall" | "small" | "medium" | "large"
+>;
```
kasperg, line 10: I appreciate this level of commenting.

I think it is important for the reader to get an understanding of why our custom types exist.

src/components/cover/helper.ts
```diff
@@ -46,41 +66,81 @@ export const getCoverUrl = ({
...
+ */
+export function resolveCoverPids({
+  idType,
+  ids,
+  resolvedIsbnPid
+}: {
+  idType: CoverIdType;
+  ids?: (Pid | string)[];
+  resolvedIsbnPid?: string;
+}): string[] {
```
kasperg, line 116: Since this function resolves pids then how about returning `Pid[]` instead of the generic `string[]`?

src/components/cover/helper.ts
```diff
@@ -46,41 +66,81 @@ export const getCoverUrl = ({
...
 
+/**
+ * Filters out null entries from a list of manifestations.
+ * Use when GraphQL returns `(Manifestation | null)[]`.
+ */
+export function filterNonNullManifestations<T>(
+  manifestations: (T | null | undefined)[] | null | undefined
+): T[] {
+  return manifestations?.filter((m): m is T => m != null) ?? [];
+}
```
kasperg, line 99: Can we replace this with something like [`_.compact()`](https://lodash.com/docs/4.17.15#compact) with the same level of type safety?

src/apps/material/instant-loan.test.ts
```diff
@@ -8,15 +8,19 @@ describe("Instant Loan", () => {
...
       fixtureFilePath: "material/user.json"
     });
 
-    cy.interceptRest({
-      aliasName: "Cover",
-      url: "**/api/v2/covers?**",
+    cy.interceptGraphql({
+      operationName: "GetCoversByPids",
       fixtureFilePath: "cover.json"
     });
```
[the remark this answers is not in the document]
LasseStaus: Yep removed duplicate..  /sigh, ved ik hvordan jeg har puttet flere styks ind

src/apps/material/periodical.test.ts
```diff
@@ -1,20 +1,14 @@
...
-      },
-      {
-        fixture: "images/cover.jpg"
-      }
-    );
+
+    cy.interceptGraphql({
+      operationName: "WorkRecommendations",
+      fixtureFilePath: "material/material-grid-related-recommendations.json"
+    });
```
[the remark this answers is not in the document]
LasseStaus: i think it's necessary because if the material-grid-related-recommendations don't load (they are present at the bottom of the page), the test stops in a forbidden since that query is not mocked / intercepted.

src/apps/material/periodical.test.ts
```diff
@@ -43,6 +37,11 @@ describe("Material - Periodical", () => {
       fixtureFilePath: "material/periodical-fbi-api.json"
     });
 
+    cy.interceptGraphql({
+      operationName: "GetCoversByPids",
+      fixtureFilePath: "cover.json"
+    });
+
```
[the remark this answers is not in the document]
LasseStaus: Another duplicate. /sigh..

cypress/fixtures/cover-get-best-representation-by-isbn.json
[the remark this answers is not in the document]
LasseStaus: Done 👍

src/components/cover/helper.ts
kasperg: Lodash is actually pretty good at type hinting with `@types/lodash`. [As I read the code it infers the types correctly](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/lodash/fp.d.ts#L302).

I am OK with letting this stay. I general I prefer that we use the helpers we are provided instead of trying to implement our own or letting CoPilot do it.

src/components/cover/helper.ts
kasperg: Let's leave it as is.

src/components/cover/helper.ts
LasseStaus: I see! Noted for next time then 👍

PR #1922, opened 2025-06-17
DDFSAL-105 - Prevent modal re-opening on Enter/Space

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFSAL-105

#### Test
https://varnish.pr-2528.dpl-cms.dplplat01.dpl.reload.dk/

#### Dev links
https://ui.lagoon.dplplat01.dpl.reload.dk/projects/dpl-cms/dpl-cms-pr-2528

https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/2528

#### Description
<s>This pull request resolves an issue where the material details modal would reopen if users closed it using the Space or Enter key.

Initially, I attempted to address this by setting returnFocusOnDeactivate: false on the FocusTrap inside the main modal:

```ts
// Set returnFocusOnDeactivate to false to prevent the modal from reopening
// when the close button is used with keyboard navigation.
returnFocusOnDeactivate: false,
```

However, this approach introduced an unintended side effect: focus was returned to the <body> element rather than the previously focused element, which is undesirable for accessibility and keyboard navigation flow.

Upon further investigation, I discovered the root cause was a lingering keyup event. When the modal is closed, FocusTrap restores focus to the original trigger element. If the key that closed the modal (Enter or Space) is still “in flight” during the keyup phase, it re-triggers the same handler—causing the modal to reopen.

The final solution was to handle the keyboard event on keydown instead of keyup. This ensures the handler executes before focus is restored or any synthetic clicks are dispatched, thereby preventing the issue.</s>

This pull request introduces a consistent approach to handling user interactions with modals across various components by replacing `onClick` handlers with `onMouseUp` and adding `onKeyUp` handlers that utilize a new helper function, `isActivationKeys`. This ensures better accessibility and standardization for keyboard navigation. Additionally, the `isActivationKeys` function is implemented in `src/core/utils/helpers/general.ts` to centralize the logic for detecting activation keys like "Enter" and "Space."

### DIFF
--- a/src/apps/loan-list/materials/stackable-material/stackable-material.tsx
+++ b/src/apps/loan-list/materials/stackable-material/stackable-material.tsx
@@ -9,7 +9,7 @@ import fetchDigitalMaterial from "../utils/digital-material-fetch-hoc";
 import MaterialStatus from "./material-status";
 import { LoanId } from "../../../../core/utils/types/ids";
 import ListMaterialSkeleton from "../../../reservation-list/reservation-material/list-material-skeleton";
-import { isActivationKeys } from "../../../../core/utils/helpers/general";
+import { isEnterOrSpacePressed } from "../../../../core/utils/helpers/general";
 
 export interface StackableMaterialProps {
   loan: LoanType;
@@ -53,7 +53,7 @@ const StackableMaterial: FC<StackableMaterialProps & MaterialProps> = ({
         // `!focused` prevents opening material details modal after clicking
         // enter on pager. Pager gives focus to the next stackable material too
         // quickly while still registering the enter key press.
-        if (isActivationKeys(e.key) && !focused) {
+        if (isEnterOrSpacePressed(e.key) && !focused) {
           handleOpenDueDateModal();
         }
       }}
--- a/src/core/utils/helpers/general.ts
+++ b/src/core/utils/helpers/general.ts
@@ -555,7 +555,7 @@ export const formatDanishPhoneNumber = (phoneNumber: string) => {
 export const parseBoolean = (value?: string): boolean =>
   value?.toLowerCase() === "true";
 
-export const isActivationKeys = (key: string): boolean => {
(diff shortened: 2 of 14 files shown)


src/apps/loan-list/materials/stackable-material/stackable-material.tsx
```diff
@@ -48,7 +48,7 @@ const StackableMaterial: FC<StackableMaterialProps & MaterialProps> = ({
       )}
       role="button"
       onClick={handleOpenDueDateModal}
-      onKeyUp={(e) => {
+      onKeyDown={(e) => {
```
Adamik10, line 51: please make sure to test this then holding the button down for a bit to see how it reacts

kasperbirch1: Good point, there’s something I need to look into before merging.

src/core/utils/helpers/general.ts
```diff
@@ -555,6 +555,10 @@ export const formatDanishPhoneNumber = (phoneNumber: string) => {
 export const parseBoolean = (value?: string): boolean =>
   value?.toLowerCase() === "true";
 
+export const isActivationKeys = (key: string): boolean => {
```
spaceo, line 558: Would it be possible to name it more precisely?
Either the context could be thrown into the name, eg.: "isModalActivationKeys"
Or a bit clumsy, yet understandable, alternative could be: "keysEnterOrSpaceWasPressed"
Or you could make a function with the previous name suggested: "keysEnterOrSpaceWasPressed" and then make a function for each purpose that points at "keysEnterOrSpaceWasPressed" - example:
```typescript
// Global helper
export const keysEnterOrSpaceWasPressed = (key: string): boolean => {
  return key === "Enter" || key === " ";
};

// Local function
export const materialModalIsOpen = (key: string): boolean => keysEnterOrSpaceWasPressed(key)
```

PR #1923, opened 2025-06-17
DDFSAL-178 - Fix error when chane pincode from mobile (iPhone safari)

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFSAL-178

#### Test
https://varnish.pr-2529.dpl-cms.dplplat01.dpl.reload.dk/

#### Dev links
https://ui.lagoon.dplplat01.dpl.reload.dk/projects/dpl-cms/dpl-cms-pr-2529

https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/2529

#### Description

<s>On iOS/Safari, a successful pincode change request returns a 204 No Content response with no body, and it often omits or misreports the Content-Type header. Attempting to call .json() or catching a SyntaxError from parsing an empty or non-JSON response leads our client to incorrectly treat the update as a failure.

Changes
	1.	Immediately short-circuits on 204 responses by returning null, avoiding any attempt to read the response body.
	2.	Reads the response body into a text variable and only invokes JSON.parse if the Content-Type is explicitly application/json.
	3.	Removes the inner SyntaxError catch—any parse errors now propagate to the outer catch block and are consistently converted into a FetchFailedCriticalError.

These changes ensure that pincode updates on iPhone no longer trigger false failure states, while also simplifying and unifying the fetcher’s error-handling logic.</s>

Fix JSON parsing errors for empty responses

Problem
Empty responses were causing JSON parsing errors in the FBS fetcher.

Solution
Now handles empty responses by checking response.text() before attempting JSON parsing. If the response text is empty, returns null instead of trying to parse it as JSON.

### Note:
https://stackoverflow.com/questions/55681242/i-keep-getting-the-error-the-string-did-not-match-the-expected-pattern-for-my

### DIFF
--- a/src/core/fbs/mutator/fetcher.ts
+++ b/src/core/fbs/mutator/fetcher.ts
@@ -50,13 +50,21 @@ export const fetcher = async <ResponseType>({
       );
     }
 
-    try {
-      return (await response.json()) as ResponseType;
-    } catch (e) {
-      if (!(e instanceof SyntaxError)) {
-        throw e;
-      }
+    // If the server returns 204 No Content, there is intentionally no body to parse
+    if (response.status === 204) {
+      return null;
     }
+
+    const contentType = response.headers.get("content-type") || "";
+    const text = await response.text();
+
+    // If the response is JSON, parse it (or return null when the body is empty)
+    if (contentType.includes("application/json")) {
+      return text ? (JSON.parse(text) as ResponseType) : null;
+    }
+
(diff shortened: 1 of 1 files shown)


kasperg (changes requested):
I need more info about this change. Let's talk about it.

Reading your comments and the code does not make sense to me.

kasperg (approved):
:+1: Much nicer and simpler

PR #1926, opened 2025-06-18
fix: remove generateItemId function

ThomasGross:
#### Link to issue

https://reload.atlassian.net/browse/DDFBRA-24

#### Description

fix: remove generateItemId function to ensure keys are not changing during re-render


Adamik10 (approved):
VERY NICE!  👏

PR #1994, opened 2025-08-04
DDFSAL-269 - Refactor search header to use local suggestions query

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFSAL-269

#### Test
https://varnish.pr-2651.dpl-cms.dplplat01.dpl.reload.dk/

#### Dev links
https://ui.lagoon.dplplat01.dpl.reload.dk/projects/dpl-cms/dpl-cms-pr-2651

https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/2651

#### Description
The library wants to only show materials that are available. Therefore, I’ve added localSuggestionsFromQueryString in src/components/autosuggest/autosuggest-local.graphql, generated new hooks and types, and replaced the old hook in the relevant places. Related tests have also been updated.

Skærmbillede 2025-08-04 kl  14 20 28


kasperbirch1:
> In my opinion we could keep the whole `suggestionsFromQueryString` scope. It is an implementation detail that the query uses `localSuggest`. Since we do not need `suggestionsFromQueryString` and `localSuggestionsFromQueryString` we could just stick with the usual naming and not refactor the whole thing. Does it make sense?

Yes, please take a look now

PR #2036, opened 2025-09-01
DDFSAL-312 - copy link

kasperbirch1:
#### Link to issue

https://reload.atlassian.net/browse/DDFSAL-312

#### Test
https://varnish.pr-2730.dpl-cms.dplplat01.dpl.reload.dk/

#### Dev links
https://ui.lagoon.dplplat01.dpl.reload.dk/projects/dpl-cms/dpl-cms-pr-2730

https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/2730


src/apps/material/helper.ts
```diff
@@ -684,6 +685,19 @@ export const getWorkTitle = (work: Work): string => {
...
 };
 
+// Constants
+export const MANIFESTATION_HASH_PREFIX = "manifestation-" as const;
+
+export const createManifestationId = (pid: string): string => {
+  return `${MANIFESTATION_HASH_PREFIX}${pid}`;
+};
+
+export const shouldOpenManifestationDetails = (pid: string): boolean => {
```
xendk, line 695: Function name focused on one use case, but it's more generic than that. `pidInUrlHash` would be a better description. But also see next comment.

src/core/utils/helpers/url.ts
```diff
@@ -200,3 +200,13 @@ export const isUrlValid = (text: string) => {
 export const currentLocationWithParametersUrl = (
   params: Record<string, string>
 ) => appendQueryParametersToUrl(new URL(getCurrentLocation()), params);
+
+export const getUrlHashWithPrefix = (prefix: string): string => {
```
xendk, line 204: This is used twice. Once in `shouldOpenManifestationDetails` where it's not needed and once for `shouldOpenEditionsDisclosure` which is just interested in whether there is one. Maybe it should be a `hasMaterialUrlHash` (in another file, of course)?

src/apps/material/helper.ts
```diff
@@ -684,6 +685,19 @@ export const getWorkTitle = (work: Work): string => {
...
+  const manifestationId = createManifestationId(pid);
+  const hashWithPrefix = getUrlHashWithPrefix(MANIFESTATION_HASH_PREFIX);
```
xendk, line 697: `getUrlHashWithPrefix` is a bit useless here, whether the prefix matches is irrelevant.

The entire function could be:

`createManifestationUrlHash(pid) === getUrlHash()`.

Assuming a general `getUrlHash` wrapper around `window.location.hash`. But then we might completly eliminate this function and just use `'#' + createManifestationUrlHash(pid) === window.location.hash` where we need to figure out if the current hash matches the pid.

Or turn the thing around, rather than have this take a pid, convert it to a hash and then compare that to the hash gotten from `getUrlHashWithPrefix`, go the other way and have a `getManifestationFromUrlHash()` that returns the pid from the hash or null if none found. Then later you can just do a

```typescript
                  const shouldOpenDetails = getManifestationFromUrlHash() === manifestation.pid;
```
and
```typescript
  const shouldOpenEditionsDisclosure = !!getManifestationFromUrlHash();
```

And avoid having to pass `MANIFESTATION_HASH_PREFIX` around.

src/components/material/MaterialMainfestationItem.tsx
```diff
@@ -163,11 +168,18 @@ const MaterialMainfestationItem: FC<MaterialMainfestationItemProps> = ({
...
-          />
+          <>
+            <MaterialDetailsList
+              id={detailsId}
+              className="mt-24"
+              data={detailsListData}
+            />
+            <CopyLink
+              label={t("copyLinkToEditionText")}
+              url={`${window.location.origin}${window.location.pathname}${window.location.search}#${manifestationId}`}
```
xendk, line 179: Considering that you created a function to return the hash, it seems odd that you just grab `window.location` here. A `getCurrentUrlWithHash()` function perhaps?

src/core/utils/UseScrollToLocation.tsx
```diff
@@ -2,19 +2,29 @@ import { useEffect, useState } from "react";
 
 // The dependency prop is given to the hook so that it can scroll based on
 // external changes (e.g. scroll when some data on the page loads).
-export const useScrollToLocation = (dependency: unknown) => {
+export const useScrollToLocation = (dependency: unknown, delay: number = 0) => {
```
xendk, line 5: Probably my lack of React skills here, but I would have thought you'd be able to compose effects so you could do something like `useDelay(300, useScrollToLocation(locator))`.

xendk (changes requested):
I find this difficult to read and seems needlessly complicated at places.

And the commit messages isn't helping much as they're just lists of changes, not the why or how.

src/core/utils/UseScrollToLocation.tsx
kasperbirch1: No longer relevant.  We now use useScrollAfterFetchWithRetry

PR #2063, opened 2025-09-16
Loading text when OpeningHoursEditor is waiting for CMS. DDFSAL-79

rasben:
Screenshot 2025-09-16 at 14 37 51

### DIFF
--- a/src/apps/opening-hours-editor/OpeningHoursEditor.tsx
+++ b/src/apps/opening-hours-editor/OpeningHoursEditor.tsx
@@ -13,6 +13,7 @@ import DialogFormAdd from "./DialogFormAdd";
 import { OpeningHoursCategoriesType } from "./types";
 import { useConfig } from "../../core/utils/config";
 import { useText } from "../../core/utils/text";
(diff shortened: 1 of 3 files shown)


src/apps/opening-hours-editor/OpeningHoursEditor.tsx
```diff
@@ -57,6 +59,35 @@ const OpeningHoursEditor: React.FC<OpeningHoursEditorType> = ({
...
+            inset: 0,
+            backgroundColor: "rgba(255, 255, 255, 0.75)",
+            display: "flex",
+            gap: "10px",
+            flexDirection: "column",
+            justifyContent: "center",
+            alignItems: "center",
+            boxSizing: "border-box",
+            zIndex: 10
+          }}
```
ThomasGross: We should avoid using inline styles. Make a class in the design system instead.

ThomasGross (approved):
Some styling issues but otherwise good! 👍

Approved when styling is fixed

ThomasGross (approved):

PR #2069, opened 2025-09-24
fix: fallbackFocus is required in case of async components inside the…

ThomasGross:
… children of focusTrap

#### Link to issue

https://reload.atlassian.net/browse/DDFSAL-391

#### Description

To prevent errors in audio book player we added a fallback to document body. This change prevents focusTrap to throw an error which previously made the material page non-interactive and showed an error banner. 

This is further described in: https://github.com/focus-trap/focus-trap?tab=readme-ov-file#fallbackfocus

### DIFF
--- a/src/core/utils/modal.tsx
+++ b/src/core/utils/modal.tsx
@@ -14,7 +14,6 @@ import {
   currentLocationWithParametersUrl,
   redirectToLoginAndBack
 } from "./helpers/url";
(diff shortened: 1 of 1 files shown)


kasperbirch1 (approved):
I'm a little unsure about the consequences of this. I approve but think that Nino should test modals thoroughly.

PR #2077, opened 2025-09-29
chore: update dependency ignore list for happy-dom, caniuse-lite, and…

ThomasGross:
… @types/react

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -84,3 +84,9 @@ updates:
         versions: [">=4"]
       - dependency-name: "dotenv"
         update-types: ["version-update:semver-major"]
+      - dependency-name: "happy-dom"
+        update-types: ["version-update:semver-major"]
+      - dependency-name: "caniuse-lite"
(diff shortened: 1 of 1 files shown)


ThomasGross:
Closing this PR 

> This should not be merged.
> 
> Even though tests are failing in #2076, #2074 and #2073 it is not because of incompatibility.
> 
> When running tests in parallel with Cypress Cloud orchestrating the distribution we can end up in situations where a test run does not actually run any tests. This in turn will not generate any code coverage which again will make generating a code coverage report fail. This in turn has made CI runs fail seemingly randomly.
> 
> I have tried to address this in #1784 but without immediate success.
> 
> When tests have failed I have checked out the cause in GitHub Action workflows (which you should anyway). If the code coverage problem occurred then I have rerun the test.
> 
> This has also been discussed in https://reload.zulipchat.com/#narrow/channel/240325-DDF/topic/Flaky.20code.20coverage/with/523239215.

It turn out to be flaky tests. All PR's are now merged. Closing this PR.

